PR #8, opened 2015-06-02
Account.isValidAddress fix

bartekn:
`Account.isValidAddress` was returning `false` for every input.

I created some tests and one of them fails when old network address is passed as an argument (`g4VPBPrHZkfE8CsjuG2S4yBQNd455UWmk`). Do we want `Account.isValidAddress` to return `false` is such case?


lib/account.js
```diff
@@ -8,7 +8,7 @@ Object.defineProperty(exports, "__esModule", {
     value: true
 });
 
-var decodeBase58Check = require("./index").decodeBase58Check;
+var decodeBase58Check = require("./base58.js").decodeBase58Check;
```
nullstyle: Please remove the .js suffix.

nullstyle:
Could you please explain why simply changing the import from `index` to `base58` fixes this issue?  Is babel not properly polyfilling module bindings?

bartekn:
It was throwing `undefined is not a function` error. My guess is it was happening because Account export is before base58 export in index.js file:
```js
export {Account} from "./account";

export * from "./base58";
```

nullstyle:
Yeah, that makes me think it's just a present deficiency of babel's module system polyfill (unless I'm understaning es6 bindings incorrectly, which totally could be the case).

:+1: after you fix the test suite

nullstyle:
Also... js-stellar-base should not validate old addresses.  It should only deal with validity as defined by the new network.

PR #9, opened 2015-06-05
add repository field in package.json

irisli:
This should fix the warnings when installing something:
```
npm WARN package.json js-stellar-lib@0.0.7 No repository field.
```


bartekn:
:+1:

PR #13, opened 2015-07-10
Fix signing:

nullstyle:
This commit makes js-stellar-base compatible with the latest
stellar-core with regards to signing.

In addition, it adds `verify` to Keypair, signature helper methods to
transaction, and it rewrites some of the TransactionBuilder for less
duplication.


FredericHeem:
Can someone validate this critical PR ?

bartekn:
:+1: LGTM but needs to be rebased.

PR #15, opened 2015-07-22
Switch keys from Base58Check to StrCheck (Base32 + CRC16-XModem)

bartekn:
As in https://github.com/stellar/ruby-stellar-base/pull/4


src/strkey.js
```diff
@@ -70,14 +54,13 @@ export function encodeBase58Check(versionByteName, data) {
...
-  return encodeBase58(unencoded);
+  return base32.encode(unencoded).toString();
 }
 
 function calculateChecksum(payload) {
-  let inner = hash(payload);
-  let outer = hash(inner);
-  return outer.slice(0,4);
-} 
+  let checksum = ("0000"+crc.crc16xmodem(payload).toString(16)).slice(-4); // Padding to 2 bytes
```
nullstyle: You could provide some more explanatory text around this block? it's not clear what your algorithm is and why.  For example:

- What  are you padding to 2 bytes?  what is it's length prior?  Why is it being padded to two bytes?
- Are you converting a big-endian value to little-endian, or is it the other way around?  The code doesn't specify, it's just flipping endianness.

test/unit/operation_test.js
```diff
@@ -19,9 +19,9 @@ describe('Operation', function() {
 
     describe(".payment()", function () {
         it("creates a paymentOp", function () {
-            var destination = "gspbxqXqEUZkiCCEFFCN9Vu4FLucdjLLdLcsV6E82Qc1T7ehsTC";
+            var destination = "GB7ANT5Q6PZ6KJGAXGQ63ZX3TM7BBLANJTUJ4SQOAHRVWQL6HNVJPYLZ";
```
nullstyle: You've changed the destination in this test to a different value than the `createAccount()` test that used the same address before.
switch_keys_from_base58check_to_strcheck__base32___crc16-xmodem__by_bartekn_ _pull_request__15_ _stellar_js-stellar-base

Is that intentional?

nullstyle:
Too many of the new addresses in the test suite have been changed, not simply translated consistently, such that I cannot easily determine whether the test suite is still properly testing things from the diff alone...  Could you resolve this?

For example, every time I see the previous address of "gspbxqXqEUZkiCCEFFCN9Vu4FLucdjLLdLcsV6E82Qc1T7ehsTC", I expect to see the same new value replacing it.

src/strkey.js
bartekn: I realised there is a self-explanatory way to implement this function.

test/unit/operation_test.js
bartekn: No, I just assumed that since tests are independent it doesn't matter. Am I wrong?

bartekn:
> Too many of the new addresses in the test suite have been changed, not simply translated consistently, such that I cannot easily determine whether the test suite is still properly testing things from the diff alone... Could you resolve this?

Doing this right now.

nullstyle:
Awesome!  Thanks so much

bartekn:
Fixed.

test/unit/strkey_test.js
```diff
@@ -0,0 +1,61 @@
...
+let seedEncoded      = StellarBase.encodeCheck("seed", unencodedBuffer);
+
+describe('StellarBase#decodeCheck', function() {
+
+  it("decodes correctly", function() {
+    expectBuffersToBeEqual(StellarBase.decodeCheck("accountId", accountIdEncoded), unencodedBuffer);
+    expectBuffersToBeEqual(StellarBase.decodeCheck("seed", seedEncoded), unencodedBuffer);
+  });
+
+  it("throws an error when the version byte is not defined", function() {
```
nullstyle: This test could be better if it used a valid base32 encoded value with a bad version byte for the second parameter sent to decode check.  Since we're checking the error string, it's not very important though; otherwise it could lead to false positives since the test does not supply a value that matches only the defined context.

nullstyle:
Looks great!  :+1:

test/unit/strkey_test.js
bartekn: Fixed in f54d324.

PR #23, opened 2015-08-21
Getting rid of encapsulation violations as described in #22

### LINKED ISSUE #22: Fix XDR encapsulation violations
There are a number of encapsulation violations, that is, a number of objects within this project that access the private state of different object.  For reason's I don't want to fully document here--google "why is encapsulation important" for reasoning--we should fix this.

How do you detect this violation?  We use a leading underscore to denote a property that is private to an object.  For example, every object that is an instance of an XDR struct keeps its internal attribute state in a property named `_attributes`.  The only acceptable accessor to these properties is the object itself, thus, the only valid access of that private statement is of the form `this._attributes`.  Anytime you see `notThis._somePrivateAttribute` we have an encapsulation violation.

While I have not enumerated every instance of this violation in the codebase, I have identified some patterns:

1.  Accessing the value of an XDR union.  There are several instances where the value of a union is accessed through the private value `_value`.  Instead, you should call the accessor method `value()`.  Thus `body._value` becomes `body.value()`

2.  Accessing the switch (i.e. disciminant) of a union.  Like 1, you should not be referring to  `someUnion._switch` directly, instead using `someUnion.switch()`

3.  Accessing xdr struct attributes through the `_attributes` private attributes.  Each attribute on a struct has a helper accessor method that should be used.  Instead of `signer._attributes.weight` it should be `signer.weight()`.

There are probably more patterns... as you see them, feel free to ask a question about how to resolve the problem.

dydt:
Fixes #22 

I changed `operationToObject` to take an `operation` object, not `operation._attributes` and updated the code and tests accordingly.

I also fixed some encapsulation violations in `asset_tests`.

### DIFF
--- a/lib/operation.js
+++ b/lib/operation.js
@@ -417,87 +417,88 @@ var Operation = exports.Operation = (function () {
                     return encodeCheck("accountId", accountId.ed25519());
                 }
 
-                var obj = {};
-                var attrs = operation.body._value && operation.body._value._attributes;
-                switch (operation.body["switch"]().name) {
+                var result = {};
+                var attrs = operation.body().value();
+                switch (operation.body()["switch"]().name) {
                     case "createAccount":
-                        obj.type = "createAccount";
-                        obj.destination = accountIdtoAddress(attrs.destination);
-                        obj.startingBalance = attrs.startingBalance.toString();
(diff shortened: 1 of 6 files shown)


nullstyle:
lgtm

PR #24, opened 2015-08-21
Keypair.fromBase58Seed method

bartekn:
This PR adds `Keypair.fromBase58Seed` method to create a new format Keypair from the old base58 format. Required when converting old format keys persisted in stellar-wallet.


nullstyle:
Could you put a code comment on `fromBase58Seed` explainging that this address format is deprecated and this method is provided for helping the transition to base32?

Other than that, :+1:

PR #36, opened 2015-09-04
Add network support

nullstyle:
For compatibility with https://github.com/stellar/stellar-core/pull/728


src/transaction.js
```diff
@@ -68,15 +69,12 @@ export class Transaction {
     */
     signatureBase() {
         return Buffer.concat([
-            this.signatureBasePrefix(),
+						Network.current().networkId(),
```
irisli: what happened to the indentation here?

src/network.js
```diff
@@ -0,0 +1,54 @@
+import {hash} from "./hashing";
+
+export let Networks = {
+	PUBLIC: "Public Global Stellar Network ; September 2015",
+	TESTNET: "Test SDF Network ; September 2015",
+};
```
bartekn: `const` will prevent from accidentally overwriting values.

nullstyle: Ah, good catch.  Thanks!

PR #39, opened 2015-09-10
Network helper functions

bartekn:
Helper function for `Network`:
- `usePublicnet()` (lower case `n`)
- `useTestnet()` (lower case `n`)


bartekn:
Close in favour of #40.

PR #41, opened 2015-09-11
Change default network to testnet

bartekn:
This PR changes default network to testnet. Reasoning:
* Testnet is, in most cases, the first network developers are interacting with,
* With default Public Network developers can send real funds by accident. They should explicitly state they want to use Public Network (by calling `usePublicNet()`).


nullstyle:
:+1:

PR #51, opened 2015-09-30
Master key dependent on network pass

### LINKED ISSUE #42: KeyPair.master
With the recent change in stellar-core, the master key has changed.
The code https://github.com/stellar/js-stellar-base/blob/master/src/keypair.js#L36 is now incorrect.

dydt:
Small change that fixes #42.


nullstyle:
:+1:

PR #64, opened 2015-12-09
Throw Error when the best rational approximation cannot be found

bartekn:
Throw `Error` in `best_r` when the best rational approximation cannot be found. Operations that contain such approximations would be [rejected](https://github.com/stellar/stellar-core/blob/eed89649c2060b8e9dacffe2cec4e8b258b32416/src/transactions/ManageOfferOpFrame.cpp#L390) by stellar-core anyway so it's better to reject them early.


nullstyle:
+1, although IMO we should probably think about allowing customizable rounding modes in the future that would ensure that an error can't occur within the function.

Issue #63, opened 2015-12-08
not in sync with latest horizon: address => account_id


FredericHeem:
The new horizon has changed `address` in `account_id`, however, `js-stellar-base` is still using `address`: https://github.com/stellar/js-stellar-base/blob/master/src/transaction_builder.js#L112

bartekn:
We're in the process of changing this (the process started with this issue: https://github.com/stellar/horizon/issues/214). Will update it very soon and depreciate methods.

PR #65, opened 2015-12-09
Replace Address with Account ID

bartekn:
To implement: https://github.com/stellar/horizon/issues/214
Close https://github.com/stellar/js-stellar-base/issues/63


nullstyle:
+1.  Thanks!

PR #73, opened 2016-01-28
price fraction in Operation.managerOffer and Operation.createPassiveOffer

FredericHeem:
The PR allows the price to be an object with numerator and denominator in Operation.managerOffer and Operation.createPassiveOffer. Otherwise, precision is lost, for instance, `1000` becomes `999.99999`.

### DIFF
--- a/src/operation.js
+++ b/src/operation.js
@@ -597,20 +597,22 @@ export class Operation {
      * @private
      */
     static _toXDRPrice(price) {
-        if(price.n && price.d){
-          if ((price.n < 0) || (price.d < 0)) {
-            throw new Error('price numerator and denominator must be positive');
-          }
-          return new xdr.Price(price);
+        let xdrObject;
+        if (price.n && price.d) {
+            xdrObject = new xdr.Price(price);
+        } else {
+            price = new BigNumber(price);
+            let approx = best_r(price);
+            xdrObject = new xdr.Price({
+                n: parseInt(approx[0]),
+                d: parseInt(approx[1])
+            });
         }
-        price = new BigNumber(price);
-        if (price.lte(0)) {
+
+        if (xdrObject.n() < 0 || xdrObject.d() < 0) {
             throw new Error('price must be positive');
         }
-        let approx = best_r(price);
-        return new xdr.Price({
-            n: parseInt(approx[0]),
-            d: parseInt(approx[1])
-        });
+
+        return xdrObject;
     }
 }
--- a/src/operation.js
+++ b/src/operation.js
@@ -307,7 +307,9 @@ export class Operation {
     * @param {Asset} opts.selling - What you're selling.
     * @param {Asset} opts.buying - What you're buying.
     * @param {string} opts.amount - The total amount you're selling. If 0, deletes the offer.
-    * @param {number|string|BigNumber} opts.price - The exchange rate ratio (selling / buying).
+    * @param {number|string|BigNumber|Object} opts.price - The exchange rate ratio (selling / buying)
+    * @param {number} opts.price.n - The price numerator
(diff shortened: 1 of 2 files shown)


bartekn:
Are you sure that precision is lost? I've just tested it and:
```js
console.log(StellarBase.Operation._toXDRPrice(1000));
```
gives me:
```
ChildStruct { _attributes: { n: 1000, d: 1 } }
```
which is a correct value. Can you share your code?

For prices that can't be represented as a float number/string you can use `BigNumber` object. For example:
```js
var price = new BigNumber(1).dividedBy(3); // 0.(3)
```

FredericHeem:
Sorry but I haven't explain properly the loss of precision, suppose the following trade:
* buy 1000 EUR
* sell 1100 USD
* price is 1000/1100 EUR/USD
 At that point precision is lost on the price and 1000 EUR becomes 999.9990 EUR

bartekn:
Then you can pass it like this:
```
var price = new BigNumber(1000).dividedBy(1100);
```
But I agree that the form `{n: 1000, d: 1100}` looks better. Just one thing: can you change:
```js
         if(price.n && price.d){
           if ((price.n < 0) || (price.d < 0)) {
             throw new Error('price numerator and denominator must be positive');
           }
           return new xdr.Price(price);
         }
```
to:
```js
if (price.n && price.d) {
  price = new BigNumber(n).dividedBy(d);
}
```
This will allow developers to pass larger numbers too (as strings) and all checks will be performed only once. Notice that we're checking if the number is positive just a few lines after the code you've added:
```js
if (price.lte(0)) {
  throw new Error('price must be positive');
}
```

FredericHeem:
According to the xdr spec, the price is as follows:
```
struct Price
{
    int32 n; // numerator
    int32 d; // denominator
};
```

The PR just tried to set directly these numbers without any attempt of conversion which could lose precision.

bartekn:
Yes, the final values must be `int32` but these can be approximated from the larger values, for example (corner case):
```
n: 4294967294 d: 2 => n: 2147483647 d: 1
```
But the main advantage of creating `BigNumber` object will be a single path for checking if the value is correct.

FredericHeem:
The issue with BigNumbers in that case is that they cannot be serialized and deserialized without loss.
The bottom line is to let developer setting the price exactly how it will be on the wire, without any (lossly) conversion.

nullstyle:
To me it seems like the most sensible thing to do is front load the conversion from String/Number into the rational form, the perform your checks on the ratio.  Then instead of "short circuit a return if the input is a ratio", we "if the input isn't a ratio, convert it to one, then peform validation"

bartekn:
So you're proposing something like this:
```js
static _toXDRPrice(price) {
    let xdrObject;
    if (price.n && price.d) {
        xdrObject = new xdr.Price(price);
    } else {
        price = new BigNumber(price);
        let approx = best_r(price);
        xdrObject = new xdr.Price({
            n: parseInt(approx[0]),
            d: parseInt(approx[1])
        });
    }
    
    if (xdrObject.n() < 0 || xdrObject.d() < 0) {
        throw new Error('price must be positive');
    }
    
    return xdrObject;
}
```
Am I correct?

nullstyle:
yeah, exactly.

FredericHeem:
`BigNumber.toFraction()` could be used instead of `best_r ` ?

nullstyle:
toFraction cannot deal with the precision limits we specifically have.

bartekn:
Thanks @FredericHeem!

FredericHeem:
Thanks for the merge, do you mind bumping the npm version ?

bartekn:
Done.

PR #80, opened 2016-03-07
Make 0 a valid number for transaction fee

armed10:
The way the base sdk checks for a submitted fee rules 0 out. Since having 0 as transaction fee is actually fine in the core config, I see no reason why it can't be set in this SDK.

I've also taken the liberty of adding a test for this, but I'm not sure if this is the right way to implement this test or not. 

If this pull request fails to meet any of the contributing requirements, please let me know and I will fix them.

### DIFF
--- a/src/transaction_builder.js
+++ b/src/transaction_builder.js
@@ -9,6 +9,7 @@ import {Memo} from "./memo";
 import BigNumber from 'bignumber.js';
 import clone from "lodash/clone";
 import map from "lodash/map";
+import isUndefined from "lodash/isUndefined";
 
 let BASE_FEE     = 100; // Stroops
 let MIN_LEDGER   = 0;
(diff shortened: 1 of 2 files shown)


src/transaction_builder.js
```diff
@@ -66,7 +66,8 @@ export class TransactionBuilder {
         this.operations    = [];
         this.signers       = [];
 
-        this.baseFee    = opts.fee || BASE_FEE;
+        this.baseFee    = (isNaN(opts.fee) ? BASE_FEE : opts.fee);
```
bartekn: The value of `opts.fee` when the variable is unset is `undefined` so I'd change `isNan` to `isUndefined`.

armed10:
@bartekn isUndefined is not an existing function, is it part of some library that the sdk currently uses or should I define it somewhere?

bartekn:
Sorry, I didn't realize that `isNan` is not imported from lodash in this file. You need to import `isUndefined` like this:
```
import isUndefined from "lodash/isUndefined";
```

armed10:
@bartekn anything else I can do to improve this PR?

PR #86, opened 2016-08-03
Optional ed25519

### LINKED ISSUE #68: Moving ed25519 to optionalDependencies
Some users have problems with compiling [`ed25519`](https://www.npmjs.com/package/ed25519) bindings (and using our package). This package allows us to sign transactions extremely fast but for most use cases [`tweetnacl`](https://www.npmjs.com/package/tweetnacl) speed is enough and it does not require compilation.

Let's discuss if we want to move it to [`optionalDependencies`](https://docs.npmjs.com/files/package.json#optionaldependencies).

bartekn:
Move `ed25519` to `optionalDependencies` and fallback to `tweetnacl` if `ed25519` installation failed. Confirmed working as expected in the clean (git, node, npm, no compilation tools) Windows virtual machine.

Developers can check if native Ed25519 package is used by examining `StellarBase.fastSigning` variable.

This is a breaking change and will be published in `0.6.0`.

Close #68.

### DIFF
--- a/src/signing.js
+++ b/src/signing.js
@@ -17,7 +17,6 @@ export function verify(data, signature, publicKey) {
   return actualMethods.verify(data, signature, publicKey);
 }
 
-var useNacl;
 // if in node
 if (typeof window === 'undefined') {
   // NOTE: we use commonjs style require here because es6 imports
@@ -25,12 +24,12 @@ if (typeof window === 'undefined') {
   let ed25519;
   try {
     ed25519 = require("ed25519");
-    useNacl = false;
+    fastSigning = true;
   } catch (err) {
-    useNacl = true;
+    fastSigning = false;
   }
 
-  if (!useNacl) {
+  if (fastSigning) {
     actualMethods.sign = function(data, secretKey) {
       data = new Buffer(data);
       return ed25519.Sign(data, secretKey);
@@ -44,18 +43,15 @@ if (typeof window === 'undefined') {
         return false;
       }
     };
-    fastSigning = true;
   }
 } else {
-  useNacl = true;
+  fastSigning = false;
 }
 
-if (useNacl) {
+if (!fastSigning) {
   // fallback to tweetnacl.js if we're in the browser or
   // if there was a failure installing ed25519
   let nacl = require("tweetnacl");
-
-  fastSigning = false;
 
   actualMethods.sign = function(data, secretKey) {
     data      = new Buffer(data);
--- a/src/signing.js
+++ b/src/signing.js
@@ -7,7 +7,7 @@
 
 var actualMethods = {};
 
-export var fastSigning;
+export const fastSigning = checkFastSigning();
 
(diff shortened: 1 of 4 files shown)


src/signing.js
```diff
@@ -7,6 +7,8 @@
 
 var actualMethods = {};
 
+export var fastSigning;
```
nullstyle: This should probably be a `const`

src/signing.js
```diff
@@ -15,31 +17,46 @@ export function verify(data, signature, publicKey) {
   return actualMethods.verify(data, signature, publicKey);
 }
 
-
+var useNacl;
```
nullstyle: Why have both `useNacl` and `fastSigning`?  They appear to be always the same value.

src/signing.js
bartekn: OK, after a few tests it turns out that `const` does not ensure immutability after exporting. Check the following code:
```js
var StellarBase = require("stellar-base");
StellarBase.Networks.PUBLIC = "test";
console.log(StellarBase.Networks.PUBLIC);
```
I think babel ensures `const` immutability only within a single file.

I can try to use `Object.defineProperty` but the other `const`s need to be changed too so I think this should be done in a separate PR.

nullstyle: Yeah, although it would be nice for it to enforce constness, at the very least it acts as documentation to other developers.

bartekn: I think there is no way to achieve this because:
* `export` statments can be only occur at the top of the file. 
* Setting `fastSigning ` value depends on the flow of the program.

So after thinking about it, there are two ways to do it:
* `Object.defineProperty` (ugly).
* Create a method (`isFastSigningAvailable()`?) that will return a copy of `fastSigning` value.

nullstyle: > export statments can be only occur at the top of the file.

That's not true.  AFAIK, _import_ statements can only occur at the top of the file.

> Setting fastSigning value depends on the flow of the program.

declared functions (as opposed to function expressions) can be called just fine during the initial evaluation of a script.  The definitions are hoisted above the execution of the script.  The following, for example, works:

```js

export const foo = load();

function load() {
    return "hello"
}

export const blah = 3;

console.log(foo);
console.log(blah);
```

bartekn: I tried it before posting my first comment:
```
SyntaxError: /Users/bartek/js-stellar-base/src/signing.js: 'import' and 'export' may only appear at the top level (30:4)
```
The reason is I had an `export` statement inside the `if`statment. However, the solution seems to be setting `const` value to the value returned by a function (the first line in your example). I will change it.

nullstyle: Good to know about the export change...  I guess at some point it changed in the standard from how babel behaved.  The behavior is not very clear to me, even looking at the discussion on the final syntax @ http://www.2ality.com/2014/09/es6-modules-final.html

nullstyle:
+1

Issue #87, opened 2016-08-09
Document enums and flags (and make them simpler to use?)


Mr0grog:
This basically the same issue as https://github.com/stellar/java-stellar-sdk/issues/31, but for the JS SDK.

It's possible to use enums from the generated XDR classes to make code clearer, as in setting flags on an account:

```js
StellarSdk.Operation.setOptions({
  setFlags: StellarSdk.xdr.AccountFlags.authRevocableFlag().value |
    StellarSdk.xdr.AccountFlags.authRequiredFlag().value
});

// or slightly less awkward if you make yourself a shortcut:
var flags = StellarSdk.xdr.AccountFlags;
StellarSdk.Operation.setOptions({
  setFlags: flags.authRevocableFlag().value | flags.authRequiredFlag().value
});
```

It'd be helpful if these were documented.

It might also be nice if working with these was easier, though that's maybe better handled as a separate issue? (maybe the `setFlags` option could take an object, or the flags were exposed at the root level, or the flags weren't methods and you didn't have to get the `value` property from them, etc.).

PR #88, opened 2016-09-07
Export account flags constants

bartekn:
Adds and exports the account flags so `Operation.setOptions` can be used like this now:
```js
StellarBase.Operation.setOptions({
  setFlags: StellarBase.AuthRevocableFlag | StellarBase.AuthImmutableFlag
});
```

Close https://github.com/stellar/js-stellar-base/issues/87


nullstyle:
+1

PR #90, opened 2016-10-03
No default network

bartekn:
Some users forget to switch to public network (by executing `Network.usePublicNetwork()`) before deploying apps because right now test network is a default choice, selected automatically. This PR changes `Network` module to not select a default network.

Merging into `0.7.0` branch because it's a breaking change.

### DIFF
--- a/src/keypair.js
+++ b/src/keypair.js
@@ -80,6 +80,9 @@ export class Keypair {
    * @returns {Keypair}
    */
   static master() {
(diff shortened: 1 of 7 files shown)


jedmccaleb:
I think we also might have it fetch the network from horizon if horizon is set but the network hasn't been set. (I know this would happen in the SDK and not here)

jedmccaleb:
r+

Issue #94, opened 2017-01-17
add support for the new signature types


jedmccaleb:
hash(x) and hash(tx)
stellar/stellar-core#846
stellar/stellar-core#965

PR #95, opened 2017-01-19
Hash x

bartekn:
This code allows adding `hashX`, `hashTx` signers to accounts and sign transactions using `hashX` preimage. It also contains some code cleanup preparing js-stellar-base for `0.7.0` release.

This will be merged into `0.7.0` branch as it contains breaking changes.

Close https://github.com/stellar/js-stellar-base/issues/94.


src/operation.js
```diff
@@ -300,21 +302,41 @@ export class Operation {
...
+                key = new xdr.SignerKey.signerKeyTypeHashX(opts.signer.hash);
+                setValues++;
+            }
 
-            attributes.signer = new xdr.Signer({
-                pubKey: Keypair.fromAccountId(opts.signer.pubKey).xdrAccountId(),
-                weight: opts.signer.weight
-            });
+            if (setValues != 1) {
+                throw new Error("Signer object must contain exactly one of signer.pubKey, signer.hash, signer.tx must be set.");
```
vogel: Please improve grammar here:
Signer object must contain exactly one of signer.pubKey, signer.hash, signer.tx
or
Exactly one of signer.pubKey, signer.hash, signer.tx must be set.

It looks like you merged these two sentences

src/transaction.js
```diff
@@ -61,6 +62,18 @@ export class Transaction {
     }
 
     /**
+     * Add `hashX` signer preimage as signature.
+     * @param {Buffer|String} preimage Preimage of hash used as signer
+     * @returns {void}
+     */
+    signHashX(preimage) {
```
vogel: should not here be another function, signHashTx, as well?

bartekn: I think that if stellar-core processes the transaction with a hash that matches account `hashTx` signer it automatically increases weight of the transaction so there is no need to add another signature to transaction.

But I'm not 100% sure. @jedmccaleb can you confirm?

src/operation.js
bartekn: Thanks, fixed.

src/transaction.js
vogel: Yes, you are right. Sorry, its late here...

src/operation.js
```diff
@@ -557,7 +579,14 @@ export class Operation {
 
                 if (attrs.signer()) {
                     let signer = {};
-                    signer.pubKey = accountIdtoAddress(attrs.signer().pubKey());
+                    if (attrs.signer().key()._arm == "ed25519") {
```
nullstyle: Please don't break encapsulation.  `arm()`

bartekn: Thx, I couldn't find it. Would be great to have a small API reference in js-xdr. I will try to update union [example file](https://github.com/stellar/js-xdr/blob/master/examples/union.js).

src/strkey.js
```diff
@@ -6,10 +6,123 @@ import isNull from "lodash/isNull";
...
+  publicKey:  6 << 3, // G
+  seed:      18 << 3, // S
+  hashTx:    19 << 3, // T
+  hashX:     23 << 3  // X
 };
 
+/**
+ * StrKey is a helper class that allows encoding and decoding strkey.
+ */
+export class StrKey {
```
bartekn: Yeah, I was thinking for some time before changing this but I decided to change it to match [stellar-core naming](https://github.com/stellar/stellar-core/blob/next-db-and-proto-2/src/crypto/StrKey.h#L18-L21). I think that names you listed are easy to understand for end users but can mislead developers working on the code (ex. seed/secret key, "what the heck is account id?"). But I'm OK with changing this. @jedmccaleb?

jedmccaleb:
let's change the name of hash(tx)  ->  "Pre-authorized Transaction" this will avoid the hash(x) hash(tx) confusion.

Also I think calling ed25519 public keys "public keys" is better than accountID.  public keys can be signers or node ids also they are not all accountIDs.  but lets discuss at standup

bartekn:
> let's change the name of hash(tx) -> "Pre-authorized Transaction" this will avoid the hash(x) hash(tx) confusion.

Hash(tx) is used in two places currently. Do you want to change it to this:
* [`Operations.setOptions`](https://github.com/stellar/js-stellar-base/pull/95/files#diff-10d9a1b0c842bd277267de68d17be5aeR267) has an argument `hashtx` -> change to: `preauthorizedTransactionHash`
* [`StrKey.encodeHashTx`](https://github.com/stellar/js-stellar-base/pull/95/files#diff-9d57f7753705cd849cd80cba58cf264fR73) (and decodeHashTx) -> `StrKey.encodePreauthorizedTransactionHash`

or something different?
> Also I think calling ed25519 public keys "public keys" is better than accountID. public keys can be signers or node ids also they are not all accountIDs. but lets discuss at standup

Sorry for confusing you during the standup. I agree with "public keys" naming but I was thinking if we should also change other occurences of `accountId` (and my payment operation example was invalid because it's using `destination` not `accountId` name).

Here's a proposed naming convention that imho solves current issues but also won't _break things_ when a new public key signature system is introduced:
* `Public Key` is a strkey-encoded ed25519 (or any other public key signature system we will use in a future) public key. For ed25519 it starts with the letter `G`.
  * `Keypair.fromAccountId` will be renamed to: `Keypair.fromPublicKey` with argument being either ed25519 public key or any other future public key signature system public key. 
  * `Keypair.fromSecret` should stay as it is. This method should determine key type based on the input (ed25519/others seed/private key). We don't use "seed" or "key" words because we should allow passing either strkey-encoded seed/key or raw private key.
  * `Keypair.accountId` will be renamed to `Keypair.publicKey` - returns strkey-encoded public key.
  * Because there may be several public key types in a future, existing `Strkey` ed25519 public key methods will be renamed to:
    * `encodeEd25519PublicKey`
    * `decodeEd25519PublicKey`
    * `encodeEd25519SecretSeed`
    * `decodeEd25519SecretSeed`
    * similar convention for future public key signature systems.
* `Account ID` is a `Public Key` of some account in the ledger.
  * `Account.constructor` will still use `accountId` argument name because `Account` class represents an account in the ledger.

jedmccaleb:
I vote for:
Operations.setOptions has an argument hashtx -> change to: preAuthTx
StrKey.encodeHashTx (and decodeHashTx) -> StrKey.encodePreAuthTx

bartekn:
> Operations.setOptions has an argument hashtx -> change to: preAuthTx

👍

>  StrKey.encodeHashTx (and decodeHashTx) -> StrKey.encodePreAuthTx

I think it should stay as it is to be in sync with stellar-core naming. StrKey is lower level helper class and most developers won't use it directly.

jedmccaleb:
the name is changing in stellar core also:  https://github.com/stellar/stellar-core/issues/1148

nullstyle:
I'll work on flowing these changes through the go library next week.

Issue #101, opened 2017-02-26
Improve Keypair instanciation


pakokrew:
Keypair instanciation fails if we pass to the object only the secret seed, although it could deduce public from seed.

For now we must write this kind of code to instanciate : 
```js
  if (keypair.secretSeed) {
    return Keypair.fromSecret(keypair.secretSeed);
  }
  return Keypair.fromPublicKey(keypair.publicKey);
```

Also, it could be nice to have the same name for the getters than for the object properties (now we have `secretSeed` for instanciation and `secret` for the getter)

bartekn:
You can use: 
```js
let keyPair = Keypair.fromSecret(secretSeed);
console.log(keyPair.secret()); // S...
console.log(keyPair.publicKey()); // G...
```
Constructor should be used only with raw keys.

pakokrew:
I think you misunderstood my issue.
On the [reference](https://stellar.github.io/js-stellar-sdk/Keypair.html), when we use `new Keypair(keys)`, the keys object is :

```
publicKey: string
secretSeed: string (optional)
```
The public key is mandatory but secret seed is optional. What I meant is that we could either pass public or seed in props. When we pass seed we don't need to pass public key.

pakokrew:
And the `keypair.secret()` method has a different name from `keys.secretSeed` property.

PR #104, opened 2017-03-08
Improve Keypair class

bartekn:
In https://github.com/stellar/js-stellar-base/issues/101 @pakokrew noticed that `Keypair.constructor` usage is not clear. I realized that this class can be further improved (see https://github.com/stellar/js-stellar-base/pull/96).
* `Keypair.constructor` now requires `type` field to define public-key signature system used in this instance (so `Keypair` can support other systems in a future). It also checks if public key and secret key match if both are passed (to prevent nasty bugs).
* `Keypair.fromRawSeed` has been renamed to `Keypair.fromRawEd25519Seed` to make it clear that the seed must be Ed25519 seed.

### DIFF
--- a/src/keypair.js
+++ b/src/keypair.js
@@ -28,13 +28,16 @@ export class Keypair {
       throw new Error("Invalid keys type");
     }
 
+    this.type = keys.type;
+
     if (keys.secretKey) {
       keys.secretKey = new Buffer(keys.secretKey);
 
       let secretKeyUint8 = new Uint8Array(keys.secretKey);
       let naclKeys = nacl.sign.keyPair.fromSeed(secretKeyUint8);
 
-      this._secretKey = keys.secretKey;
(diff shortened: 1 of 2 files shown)


src/keypair.js
```diff
@@ -9,33 +9,51 @@ export class Keypair {
...
-      this._secretSeed = new Buffer(keys.secretSeed);
-      this._secretKey = new Buffer(keys.secretKey);
+      this._secretKey = keys.secretKey;
+      this._publicKey = new Buffer(naclKeys.publicKey);
+
+      if (keys.publicKey && !this._publicKey.equals(new Buffer(keys.publicKey))) {
+        throw new Error("secretKey does not match publicKey");
+      }
+    } else {
+      this._publicKey = new Buffer(keys.publicKey);
```
pakokrew: Maybe throw if no public key and secret are provided ?

PR #108, opened 2017-05-15
Docs: Account: typo (minor)

strogonoff:
Was confused by the docs where “Account tracts the sequential number”. I suspect it should be “tracks”.


bartekn:
Thanks!

PR #110, opened 2017-07-13
Add `browser` prop to package.json

andywer:
Make webpack & friends ignore the `ed25519` package, so you don't need the `IgnorePlugin` in the application's webpack config. The `stellar-base` package knows best what dependencies it has and how they work, so it should do the ignoring 😉

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,9 @@
     "version": "gulp build",
     "postversion": "git push && git push --tags"
   },
+  "browser": {
+    "ed25519": false
+  },
   "repository": {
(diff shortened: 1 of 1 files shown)


pakokrew:
That's a good idea :+1:

andywer:
There just seems to be something wrong with the tests.

bartekn:
Thanks!

andywer:
Hey guys! Would you mind publishing the patched version?

bartekn:
`0.7.5` released.

PR #117, opened 2018-01-12
[WIP] adding initial proposal behind common object

kaplanmaxe:
This PR is a proposal to add a `Common` object. This is currently a WIP, as I would like to get feedback before I continue working on this.

I can imagine that the majority of the people using this SDK are going to be using it for common operations (sending transactions, offers, trusts, inflation, etc). While I find this SDK easy to use, I think it would be even easier to have wrappers for common operations that developers could just pass in config objects and have the operation be completed. While these wrappers won't work for everything, I believe they will work for a good amount of use cases, and would provide a very easy way for developers to jump into Stellar.

If you find this useful, please let me know, and I can continue working on this.


tomerweller:
Thanks @kaplanmaxe. We'd like to keep this library simple and compact.
With that said, if there's demand for these common operation functions, I'd consider publishing them as a separate npm module.

PR #132, opened 2018-06-19
Fix Buffer for node v10 deprecation

robertDurst:
As of node v10, `new Buffer` is deprecated.

They recommend we replace it with one of the following:
* Buffer.from(str, encoding="utf8")
* Buffer.alloc(number, fill)
* Buffer.allocUnsafe()

Here is what node.js says:

> Drop support for Node.js ≤ 4.4.x and 5.0.0 — 5.9.x
> This is the recommended solution nowadays that would imply only minimal overhead.
> 
> The Node.js 5.x release line has been unsupported since July 2016, and the Node.js 4.x release line reaches its End of Life in April 2018 (→ Schedule). This means that these versions of Node.js will not receive any updates, even in case of security issues, so using these release lines should be avoided, if at all possible.
> 
> What you would do in this case is to convert all new Buffer() or Buffer() calls to use Buffer.alloc() or Buffer.from(), in the following way:
> 
> For new Buffer(number), replace it with Buffer.alloc(number).
> For new Buffer(string) (or new Buffer(string, encoding)), replace it with Buffer.from(string) (or Buffer.from(string, encoding)).
> For all other combinations of arguments (these are much rarer), also replace new Buffer(...arguments) with Buffer.from(...arguments).

References:
- https://nodesource.com/blog/understanding-the-buffer-deprecation-in-node-js-10/
- https://nodejs.org/en/docs/guides/buffer-constructor-deprecation/
- https://nodejs.org/api/buffer.html#buffer_buffer_from_buffer_alloc_and_buffer_allocunsafe

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ matrix:
   include:
   - node_js: 6
   - node_js: 8
+  - node_js: 10
   - node_js: 6
     env: SAUCE=true
 before_install:
--- a/src/base58.js
+++ b/src/base58.js
@@ -36,7 +36,7 @@ export function decodeBase58Check(versionByteName, encoded) {
     throw new Error(`Decoded address length is invalid. Expected 37, got ${decoded.length}`);
   }
 
-  return new Buffer(data);
+  return Buffer.from(data);
 }
 
 function calculateChecksum(payload) {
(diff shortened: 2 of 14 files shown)


bartekn:
Let's add Node 10 to `.travis.yml` matrix to check if tests pass.

robertDurst:
@bartekn Node 10 added and tests still pass

PR #139, opened 2018-07-16
Extracted operations from operation.js

robertDurst:
According to Codeclimate, one of the most offending issues in this repo is:
> File operation.js has 480 lines of code (exceeds 250 allowed). Consider refactoring. 

So, this is my proposed refactor of `operation.js`.

Essentially, I extracted the operation methods from `operation.js` into a new directory called `operations`. Then, for each operation function, I imported the function and attached it back to the `Operation` class as a static method.

**I took a couple stylistic leaps of faith here, so if anything looks wonky, I am happy to discuss!**


bartekn:
Looks good, however it breaks `jsdoc` (docs not generated for extracted methods). Please add:
```js
 * @function
 * @alias Operation.accountMerge
```
to jsdoc of each method in `ops` and it should start generating it again.

robertDurst:
@bartekn ok, should be fixed.

bartekn:
Sorry, I haven't noticed it last time: can we also move `inflation` method so all operations are in one place?

robertDurst:
Good catch -- totally forgot inflation was an operation

robertDurst:
Made codeclimate just a little bit happier (before `inflation` moved it was **F** --> **D**):

screen shot 2018-07-20 at 8 55 16 am

PR #145, opened 2018-10-20
Correcting variable name

mulbert:


tomquisel:
@bartekn if it's cool with you, we can merge changes like this with no CLA but where the contribution obviously doesn't include third party code. @jedmccaleb seems to be on board with that.

PR #147, opened 2018-10-29
Convert text memo value to String from Buffer

charlie-wasp:
Wouldn't it be better to always return a string value for text memos, no matter, is it buffer or array underneath?

I realise that it breaks backward compatibility. However, maybe there is a way to handle that?

### DIFF
--- a/src/memo.js
+++ b/src/memo.js
@@ -1,4 +1,6 @@
 import {default as xdr} from "./generated/stellar-xdr_generated";
+import isArray from "lodash/isArray";
+import isBuffer from "lodash/isBuffer";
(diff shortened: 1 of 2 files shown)


bartekn:
This has been actually changed very recently (in `0.9.0`) to use `Buffer`. Memo can contain any bytes so also bytes not compatible with `utf-8`. Merging this PR would make things like https://github.com/stellar/js-stellar-sdk/issues/196 possible again.

More info:
* https://github.com/stellar/js-stellar-base/pull/142
* https://github.com/stellar/js-xdr/pull/22

PR #149, opened 2018-12-06
Miss match source code logic and error message in validate length check(memo-hash)

fukaoi:
Hi core developer.

I Setted less than 64byte string  when I want to use hash-memo. 
` addMemo(StellarSdk.Memo.hash('xxxxxxxxxxxxxxxxxxxxx ')) `

Validation check 64byte in souce dode, but defined error message 'Expect a 32 byte hash..... '

* source code 
```js
 if (isString(value)) {
      if (!/^[0-9A-Fa-f]{64}$/g.test(value)) {
        throw error;
      }
    }
```

* error message
```js
new Error("Expects a 32 byte hash value or hex encoded string. Got " + value);
```

### DIFF
--- a/src/memo.js
+++ b/src/memo.js
@@ -131,7 +131,7 @@ export class Memo {
   }
 
   static _validateHashValue(value) {
-    let error = new Error("Expects a 32 byte hash value or hex encoded string. Got " + value);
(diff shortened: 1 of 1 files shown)


bartekn:
32 bytes are encoded to 65 character string because each byte is encoded using 2 letters (ex. `255` = `0xFF`).

PR #154, opened 2019-01-24
Standardize coding style

### LINKED ISSUE #153: Add Prettier + Eslint
See https://github.com/stellar/js-stellar-sdk/issues/220 and implement the same configs as https://github.com/stellar/js-stellar-sdk/pull/221.

morleyzhi:
Fixes #153. See also https://github.com/stellar/js-stellar-sdk/pull/221.

* Add prettier and eslint (replacing jshint)
* Add pre-commit hook for running prettier
* Add Airbnb's eslint rules
* Fix code with linter issues


bartekn:
Can you fix [errors](https://travis-ci.org/stellar/js-stellar-base/jobs/484133743#L503) detected in CI before the review?

package.json
```diff
@@ -21,7 +33,7 @@
     "stellar"
   ],
   "engines": {
-    "node": ">=4"
+    "node": ">=6.14.0"
```
bartekn: Any reason for `6.14`? I think `>=6` should be good.

src/generated/stellar-xdr_generated.js
```diff
@@ -6,3656 +6,3549 @@
...
-//           AccountID destination;
-//       case INFLATION:
-//           void;
-//       case MANAGE_DATA:
-//           ManageDataOp manageDataOp;
-//       case BUMP_SEQUENCE:
-//           BumpSequenceOp bumpSequenceOp;
-//       }
-//
-// ============================================================
```
bartekn: This is autogenerated file and should be ignored. Please revert it to the previous version.

src/operation.js
```diff
@@ -201,21 +235,23 @@ export class Operation {
...
+      // Decimal places (max 7)
+      amount.decimalPlaces() > 7 ||
       // NaN or Infinity
-      case (amount.isNaN() || !amount.isFinite()):
-        return false;
-      default:
-       return true;
+      amount.isNaN() ||
+      !amount.isFinite()
+    ) {
```
bartekn: This was part of #131 to improve [Cognitive Complexity](https://docs.codeclimate.com/docs/cognitive-complexity). Is linter feeling bad about this code?

src/operation.js
```diff
@@ -247,49 +283,38 @@ export class Operation {
...
+        (isValidFunction && isValidFunction(value, name)):
         return value;
       default:
         throw new Error(`${name} value is invalid`);
     }
   }
 
-  /**
-   * @private
-   */
```
bartekn: Please leave `@private` tags. Otherwise these function will become visible in jsdocs.

test/support/io-helpers.js
```diff
@@ -1,10 +1,14 @@
-import {times} from "lodash";
+const times = require('lodash/times');
```
bartekn: Why?

src/generated/stellar-xdr_generated.js
morleyzhi: Should be easy to have the linters ignore `src/generated`!

package.json
morleyzhi: ESLint, possibly more tools, only support 6.14 and above. Since built files will be ES5ed, they'll still work on 6.0-6.13, but I guess we could miss some bugs that are fixed in 6.14. But I think that's pretty unlikely. 

So the Travis files are set to install 6.14, and that's the version in NPM, so it makes sense to set the engine to the same value, since Babel could in the future use that to determine its transpile target.

gulpfile.js
```diff
@@ -33,91 +37,104 @@ gulp.task('test', function(done) {
...
 gulp.task('build:browser', ['lint:src'], function() {
-  return gulp.src('src/browser.js')
-    .pipe(plugins.webpack({
-      output: { library: 'StellarBase' },
-      module: {
-        loaders: [
-          { test: /\.js$/, exclude: /node_modules/, loader: 'babel-loader'}
+  return gulp
+    .src('src/browser.js')
+    .pipe(plumber())
```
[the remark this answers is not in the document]
morleyzhi: I'm okay with leaving it out. I think we should add a separate command that you run during dev, since right now any error will crash the watcher. But we can tackle that separately.

src/operation.js
morleyzhi: Interesting. I see what that article is trying to say, but I think the switch statement is too clever: switch statements aren't usually used like this, so it could easily be misread. I generally favor clear (in this case, recognizable) code over clever code. (Also note that "switch statements" is on the same list of breaks in flow as "sequences of logical operators".)

test/support/io-helpers.js
morleyzhi: Not at my laptop so can't confirm this, but I made it a rule that only `src/` gets processed through Babel, so files outside of it have to use the syntax supported by Node 6.14. I can't figure out exactly if `import` is supported, but my guess is it wasn't and I had to switch. But I could be wrong. I can revert it and see if imports work properly when I get in in a couple hours.

morleyzhi:
@bartekn All the changes you suggested should be in now, thanks!

test/support/io-helpers.js
morleyzhi: (Turns out imports are supported so I fixed this.)

bartekn:
LGTM but please remove `lib` before merging.

PR #156, opened 2019-02-08
toXDR helper

morleyzhi:
Add `Transaction.prototype.toXDR` so you can more easily get the base64-encoded XDR, for entry into the Stellar Lab XDR Viewer for example.

Also, run more files through prettier.


src/vendor/bs58.js
```diff
@@ -9,15 +9,17 @@
...
+for (var i = 0; i < ALPHABET.length; ++i) {
   ALPHABET_MAP[ALPHABET.charAt(i)] = i;
 }
 var BASE = 58;
 
 function decode(string) {
   if (string.length === 0) return [];
 
-  var i, j, bytes = [0];
+  var i,
```
tomquisel: This formatting change is funky. Are you sure it's better?

morleyzhi: Good call, that should just be ignored

PR #161, opened 2019-02-13
API for directly adding signatures to transactions

### LINKED ISSUE #109: Add signatures to transaction builder
Add a method to `TransactionBuilder` that will add existing signatures to the transaction. This will be helpful for multi-signature transactions, so one of the parties can provide a signature to be injected in a different session. Thanks @brahman81 for suggesting it.

morleyzhi:
- Add `Transaction.prototype.addSignature(publicKey: string, signature: string)` and `Transaction.prototype.getKeypairSIgnature(keypair: Keypair)` to allow, for example, parties to pre-sign their part of multi-signature transactions (fixes #109)
- Fix syntax errors in the TransactionBuilder example code (fixes https://github.com/stellar/js-stellar-sdk/issues/115)


src/transaction.js
```diff
@@ -71,6 +72,35 @@ export class Transaction {
     });
   }
 
+  /**
+   * Add a signature to the transaction. Useful when a party wants to pre-sign
+   * a transaction but doesn't want to give access to their secret keys.
+   * @param {string} publicKey The public key of the signer
+   * @param {string} signature The base64 value of the signature XDR
```
bartekn: I think it doesn't work with base64 encoded signatures. Tests seem to add byte arrays instead of b64 encoded strings.

src/transaction.js
```diff
@@ -71,6 +72,35 @@ export class Transaction {
...
+  addSignature(publicKey = '', signature = '') {
+    if (!signature) {
+      throw new Error('Invalid signature');
+    }
+
+    let hint;
+
+    try {
+      hint = Keypair.fromPublicKey(publicKey).signatureHint();
+      console.log('calculated hint: ', hint);
```
bartekn: Remove `console.log`.

src/transaction.js
morleyzhi: Good catch, should be fixed now!

morleyzhi:
@bartekn Now that I think about this... I'm not sure the method I employed is going to work because the pre-signer would need to guess the user's next sequence number. 

The docs mention a different strategy for pre-signing a transaction:
https://www.stellar.org/developers/guides/concepts/multi-sig.html#pre-authorized-transaction

I'm guessing that's already implemented in the JS SDK, if Stellar Lab already implements it? If that's the case, can we close this PR without merging since the use case we're trying to solve already has a solution?

bartekn:
[Preauthorized transaction signer](https://www.stellar.org/developers/guides/concepts/multi-sig.html#pre-authorized-transaction) is a different thing. It's basically a transaction that _unlocks_ the account. For example, you can design bonds using preauthorized transactions:
- Tx1 sends the first payment, is valid only after Jan 1st and sets the Tx2 as the signer.
- Tx2 sends the second payment, is valid only after Feb 1st and sets the Tx3 as the signer.
- and so on...

What we want to do in this PR is something like: You and me have a shared account, I sign a transaction and send it to you, you sign it and send a signature back, I add a signature using the method in this PR.

I agree that this is kind of redundant, because instead of sending me a signature you could send me a full transaction that I can later import and using `sign` method. But the new method can be useful in same cases. For example, sending a transaction to multiple signers requires the whole process to be sequential (person A signs, sends to B; person B signs, sends to C...). Adding ready signatures can be done in a simultaneous way (A sends transaction to B, C, D, E... they send back signatures and A appends them on a transaction).

morleyzhi:
@bartekn So does it seem like this PR will accomplish the use case you described? The documentation (https://github.com/stellar/js-stellar-base/pull/161/commits/898096dba59713b51517a4999dc08bf2002b895c#diff-42d8d2088a96641b563b25ad908b0c0f) should describe how end-users can use this.

test/unit/transaction_test.js
```diff
@@ -152,6 +152,74 @@ describe('Transaction', function() {
...
+          maxTime: 1739392569
+        }
+      }
+    )
+      .addOperation(
+        StellarBase.Operation.payment({ destination, asset, amount })
+      )
+      .build();
+
+    const signature = signer.sign(presignHash).toString('base64');
```
tomquisel: should this use `getKeypairSignature()` to better demonstrate the use case?

blueskytoday:
Hi I see you work with adding signatures to an existing XDR. 

I have a special interest to let a third party sign the transaction without passing the transaction as XDR. I heard it should be possible just to hand over the transaction-hash, let this get signed and add the signature to the XDR.

What is your view on this is this possible? If yes, an example would be awesome. Thanks.

abuiles:
Hi, the following should do it

```
txHash = tx.hash();
signature = signer.sign(presignHash).toString('base64')
```

Once you get back the signature, you can append it to the transaction

```
tx.addSignature(signer.publicKey(), signature);
```

Also, https://stellar.stackexchange.com/ might be a better place for this kind of questions.

Issue #166, opened 2019-02-19
Deprecate default base fee


tomerweller:
Currently, there's a base fee hardcoded into the `TransactionBuilder` class which is used as a default:
https://github.com/stellar/js-stellar-base/blob/master/src/transaction_builder.js#L10

This is bad because: 
- The base fee can go up. 
- It ignores current network fee status.

Action items: 
- [x] The next breaking-change revision of this library should issue a warning in development when the `fee` property is not provided
- [ ] Later, turn the warning into an error.
- [ ] Change the default base fee from 100 stroops to something like 500 stroops? (open for suggestions)

PR #167, opened 2019-02-20
Warn about non-explicit transaction fees

morleyzhi:
If a user calls `TransactionBuilder` without a fee, issue a console log. Also update code examples to provide a fee, and export BASE_FEE. Part of #166.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,14 +5,14 @@ A breaking change will get clearly notified in this log.
 
 ## In `master`
 
+- _Warning_ Calling TransactionBuilder without a `fee` param is now deprecated
+  and will issue a warning. In a later release, it throw an error.
 - Add a `toXDR` function for transactions that lets you get the transaction as a
   base64-encoded string (so you may enter it into the Stellar Laboratory XDR
   viewer, for one)
 - Fix TransactionBuilder example syntax errors
 - Use more thorough "create account" documentation
 - Add `Date` support for `TransactionBuilder` `timebounds`
-- _Warning_ Calling TransactionBuilder without a `fee` param is now deprecated
-  and will issue a warning. In a later release, it will instead throw an error.
 
 ## 0.11.0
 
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ A breaking change will get clearly notified in this log.
 
 ## In `master`
 
+- _Warning_ Calling TransactionBuilder without a `fee` param is now deprecated
+  and will issue a warning. In a later release, it throw an error.
```
ire-and-curses: In a later release, it **will** throw an error.

src/transaction_builder.js
```diff
@@ -7,7 +7,7 @@ import { Keypair } from './keypair';
 import { Transaction } from './transaction';
 import { Memo } from './memo';
 
-const BASE_FEE = 100; // Stroops
+export const BASE_FEE = 100; // Stroops
```
ire-and-curses: Any merit in relaxing the constant nature of this until we implement [#233](https://github.com/stellar/js-stellar-sdk/issues/233)? I saw a comment earlier about users having to hack the hard-coded value in the SDK.

src/transaction_builder.js
```diff
@@ -73,6 +73,13 @@ export class TransactionBuilder {
     }
     this.source = sourceAccount;
     this.operations = [];
+
+    if (isUndefined(opts.fee)) {
+      console.log(
+        '[TransactionBuilder] The `fee` parameter of `TransactionBuilder` is required. Future versions of this library will error if not provided.'
```
bartekn: Maybe add a package name here? Will be helpful to find what generated this message in larger projects.

src/transaction_builder.js
morleyzhi: Users shouldn't change this value; if they don't want to use it, they should use their own constant or variable.

src/transaction_builder.js
morleyzhi: The challenge I was trying to work around is that most people would import `stellar-sdk`, and a few would import `stellar-base`, and we don't know at this juncture which they're importing. So I think it's better to be vague and let the dev make the connection, vs. naming the wrong one or naming both.

CHANGELOG.md
morleyzhi: Fixed!

PR #169, opened 2019-02-27
Remove the `crypto` library

### LINKED ISSUE #128: it breaks webpack compilation with angular 6
ISSUE UPDATE:

We (Stellar.org) found a fix, and it'll go live in the next release of `stellar-base` and `stellar-sdk`, either next Monday or the following one.

---

i'm not able to complete the migration from angular 5 to 6 of my web application because of stellar-base trying to import nodejs core modules.

image

steps to reproduce the issue:

```bash
cd /tmp
npx @angular/cli new project
cd project
yarn || npm i
yarn add stellar-sdk || npm i stellar-sdk
```

and update `app.component.ts` like this:

```typescript
import { Component } from '@angular/core';
import { Transaction } from 'stellar-sdk';

@Component({
  selector: 'app-root',
  templateUrl: './app.component.html',
  styleUrls: ['./app.component.css']
})
export class AppComponent {
  title = 'app';

  constructor() {
    console.log(Transaction);
  }
}
```

and then:

`npx ng serve`

ps: with angular v5 it works.

morleyzhi:
`crypto` is a node built-in, so for browsers we have to add a polyfill for it (which breaks in Angular 6+).  It's not needed, so factor it out. Fixes #128.


test/unit/operation_test.js
```diff
@@ -1,5 +1,5 @@
 import BigNumber from 'bignumber.js';
-import crypto from 'crypto';
+import randomBytes from 'randombytes';
```
tomquisel: I don't think this import is needed here?

bartekn:
Isn't `randombytes` used in tests only? Tests are not included in the final bundle so I'm wondering if we need a new package then.

morleyzhi:
@bartekn `randombytes` wouldn't be included in the final bundle either, so there's no downside and the upside is we don't have to write our own function that generates random bytes to spec.

PR #170, opened 2019-03-01
Check received amounts

morleyzhi:
Throw an error when manageOffer would result in receiving less than 0.0000001 of a token (fixes https://github.com/stellar/js-stellar-sdk/issues/209)

### DIFF
--- a/src/operation.js
+++ b/src/operation.js
@@ -254,6 +254,47 @@ export class Operation {
     return true;
   }
 
+  static isValidReceivedAmount(unitPriceRaw, amount) {
+    if (!isString(amount)) {
+      return false;
+    }
+
+    // if the amount is zero, it's fine (zero amount = kill the offer)
+    if (new BigNumber(amount).isZero()) {
+      return true;
+    }
+
+    const unitPrice =
+      unitPriceRaw.n && unitPriceRaw.d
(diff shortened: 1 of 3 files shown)


bartekn:
I think such things should be checked at the protocol level. Users should really check tx result after network processed the transaction.

morleyzhi:
@bartekn If a user creates a transaction and sees a success (or lack of a failure) that actually failed, they're going to think that the SDK is not working properly. I think the point of the JS SDK is to make it so people don't have to deal with these edge cases. 

Plus we're already checking that the amounts are valid versus letting the protocol reject invalid amounts. This is basically the same thing.

So I can't see what's wrong about merging this fix, unless it's logically wrong?

bartekn:
> Plus we're already checking that the amounts are valid versus letting the protocol reject invalid amounts.

Correct, but we don't check if a user has a sufficient balance. We should still check obvious things that make a transaction malformed (ex. amount > 0, etc.) but the rest should be checked at the protocol level.

> unless it's logically wrong?

You don't know the exact price the trade will be executed at. The price is actually the max price you're willing to offer so it's possible that the amount bought will be higher than 10<sup>-7</sup> (because you'll buy it cheaper).

PR #172, opened 2019-03-14
Add `toString()` to Asset

charlie-wasp:
Hi there! 👋 

What about overriding `Asset.toString()` method? I believe it would be useful to return something more meaningful from that. I ended up with `code-issuer` format and just `native` for lumens.

What do you think?

P.S. Sorry for the formating mess in a diff, it's prettier in the pre-commit hook 😕

### DIFF
--- a/src/asset.js
+++ b/src/asset.js
@@ -149,4 +149,12 @@ export class Asset {
   equals(asset) {
     return this.code === asset.getCode() && this.issuer === asset.getIssuer();
   }
+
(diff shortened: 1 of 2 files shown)


charlie-wasp:
@abuiles thank you, I'm on it!

charlie-wasp:
@abuiles done ✅

morleyzhi:
:thumbsup:

PR #174, opened 2019-03-25
Also verify signature when adding to transaction

TorstenStueber:
This PR will add code to verify a signature when added to a transaction.

This is valuable for scenarios where a party receives a signature from another party and wants to add this signature to a transaction. The first party needs to make sure that the signature from the other party is valid.

This will become apparent once the transaction is submitted to horizon – it will fail if the signature is invalid. However, for smart contract schemes (say, payment channels) this will usually happen much later.

### DIFF
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -101,6 +101,7 @@ export class Transaction {
   /**
    * Add a signature to the transaction. Useful when a party wants to pre-sign
    * a transaction but doesn't want to give access to their secret keys.
+   * This will also verify whether the signature is valid.
    *
    * Here's how you would use this feature to solicit multiple signatures.
    * - Use `TransactionBuilder` to build a new transaction.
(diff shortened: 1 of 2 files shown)


src/transaction.js
```diff
@@ -130,18 +131,25 @@ export class Transaction {
...
 
     try {
-      hint = Keypair.fromPublicKey(publicKey).signatureHint();
+      keypair = Keypair.fromPublicKey(publicKey);
+      hint = keypair.signatureHint();
     } catch (e) {
       throw new Error('Invalid publicKey');
     }
 
+    if (!keypair.verify(this.hash(), signatureBuffer)) {
```
vcarl: This isn't a safe assumption to make. This only validates the "master key", but doesn't account for other signers that may be on the account. See the [docs on multisignature](https://www.stellar.org/developers/guides/concepts/multi-sig.html#additional-signing-keys) for more info.

TorstenStueber: I am not sure I understand your comment. This verifies that the provided `signature` is a valid signature of the keypair provided by the `publicKey` for this transaction.

So this does not verify the master key but it verifies the signer that provided the signature – as intended. Am I missing something?

TorstenStueber:
@tomerweller This is a valid concern. But even though adding a signature to a transaction is not technically depending on the network, it is implicitly doing so. A combination of a signature and a transaction only makes sense for a specific network, otherwise the signature would be invalid. I don't see a use case where one wants to add a signature to a transaction while not having a specific network in mind (namely the one where the signature is valid).

TorstenStueber:
@bartekn @morleyzhi any comments from you on this PR? Who would be responsible to close or accept this PR?

PR #177, opened 2019-04-17
Add required 'fee' argument to constructor example

ire-and-curses:


src/transaction_builder.js
```diff
@@ -42,7 +42,7 @@ export const TimeoutInfinite = 0;
  * a payment to `destinationB`. The built transaction is then signed by `sourceKeypair`.</p>
  *
  * ```
- * var transaction = new TransactionBuilder(source)
```
morleyzhi: The second argument is supposed to be an object with a property called fee:

`new TransactionBuilder(source, { fee })`

ire-and-curses: sorry I just realised that also - fixed

PR #187, opened 2019-05-07
Deprecate base fee with error

### LINKED ISSUE #166: Deprecate default base fee

morleyzhi:
If someone tries to build a transaction without providing the base fee, throw an error. Closes #166.

### DIFF
--- a/src/transaction_builder.js
+++ b/src/transaction_builder.js
@@ -73,7 +73,7 @@ export class TransactionBuilder {
     }
 
     if (isUndefined(opts.fee)) {
-      throw new Error('must specify fee for the transaction');
+      throw new Error('must specify fee for the transaction (in stroops)');
     }
 
     this.source = sourceAccount;
--- a/src/transaction_builder.js
+++ b/src/transaction_builder.js
@@ -71,15 +71,13 @@ export class TransactionBuilder {
     if (!sourceAccount) {
       throw new Error('must specify source account for the transaction');
     }
-    this.source = sourceAccount;
-    this.operations = [];
 
     if (isUndefined(opts.fee)) {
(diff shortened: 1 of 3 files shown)


src/transaction_builder.js
```diff
@@ -71,16 +71,14 @@ export class TransactionBuilder {
...
     }
-    this.source = sourceAccount;
-    this.operations = [];
 
     if (isUndefined(opts.fee)) {
-      // eslint-disable-next-line no-console
-      console.log(
-        '[TransactionBuilder] The `fee` parameter of `TransactionBuilder` is required. Future versions of this library will error if not provided.'
-      );
+      throw new Error('must specify fee for the transaction');
```
fnando: Should this error include an example of the expected value? E.g. `throw new Error('must specify fee for the transaction in stroops (e.g. StellarSdk.BASE_FEE or 100)`.

morleyzhi: That's a good idea. I'm hesitant to call out the library name because it's technically StellarBase.BASE_FEE, but I'm guessing the vast majority of people would see it in StellarSdk.BASE_FEE, and I don't want to invite confusion from either side.

PR #206, opened 2019-07-18
feat: deprecate Network, use explicit passphrase (1)

### LINKED ISSUE #112: Network is a global singleton
... and that's a quite big no-no, especially in JS land 😕

Consider this use case (we have a similar situation as we speak):
There is a package X using `stellar-sdk` and a package Y using `stellar-sdk` as well, but a slightly different version.

If Y is a dependency of X *or* if X and Y are both being used in another package Z then we will end up with a nasty situation. The `Network` singleton that X imports from `stellar-sdk`/`stellar-base` won't be the same instance that Y has, since they each depend on a different version (two different `stellar-base` in the `node_modules/` tree).

Thus you can run `Network.useTestNetwork()` in X and still get a `No network selected.` error in Y. Same happens when `npm link`-ing packages during development, even if all packages require the same version.

### Proposal

Is there actually a need to have this stateful `Network` singleton? I would favor a solution based on passing the pubnet/testnet switch to the `Server` constructor, since horizon servers won't suddenly change the network they are running on.

```js
import { Network, Server } from 'stellar-base'

const horizon = new Server('https://horizon-testnet.stellar.org/', { network: Network.testnet() })
```

Sorry for the wall of text 😉

Akuukis:
**See discussion at https://github.com/stellar/js-stellar-base/issues/112#issuecomment-512733154**

The following methods take optional network passphrase, and `console.warn` if it is not passed.
- static method `Keypair.master` 
- method `Transaction.addSignature`
- method `Transaction.getKeypairSignature`
- method `Transaction.hash`
- method `Transaction.sign`
- method `Transaction.signatureBase`

Also, tag whole class `Network` as deprecated.

Closes #112

### DIFF
--- a/src/keypair.js
+++ b/src/keypair.js
@@ -90,15 +90,21 @@ export class Keypair {
 
   /**
    * Returns `Keypair` object representing network master key.
+   * @param {string} [networkPassphrase] passphrase of the target stellar network (e.g. "Public Global Stellar Network ; September 2015").
    * @returns {Keypair}
    */
-  static master() {
-    if (Network.current() === null) {
+  static master(networkPassphrase) {
+    // Deprecation warning. TODO: remove optionality with next major release.
+    if (!networkPassphrase) {
+      console.warn('Global `Network.current()` is deprecated. Please pass explicit argument instead, e.g. `Keypair.master(Networks.PUBLIC)`.')
+      networkPassphrase = Network.current()
+    }
+    if (networkPassphrase === null) {
       throw new Error(
         'No network selected. Use `Network.use`, `Network.usePublicNetwork` or `Network.useTestNetwork` helper methods to select network.'
       );
     }
-    return this.fromRawEd25519Seed(Network.current().networkId());
+    return this.fromRawEd25519Seed(networkPassphrase);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


src/keypair.js
```diff
@@ -90,15 +90,21 @@ export class Keypair {
...
+   * @param {string} [networkPassphrase] passphrase of the target stellar network (e.g. "Public Global Stellar Network ; September 2015").
    * @returns {Keypair}
    */
-  static master() {
-    if (Network.current() === null) {
+  static master(networkPassphrase) {
+    // Deprecation warning. TODO: remove optionality with next major release.
+    if (!networkPassphrase) {
+      console.warn('Global `Network.current()` is deprecated. Please pass explicit argument instead, e.g. `Keypair.master(Networks.PUBLIC)`.')
+      networkPassphrase = Network.current()
```
abuiles: @Akuukis shouldn't this be `Network.current()..networkId()` ?

Akuukis: Yes. I think this PR is not relevant anymore as opinions converge to 2nd solution of the problem. Over at other PR (#207) this problem is already fixed.

abuiles:
Superseded by #207

PR #207, opened 2019-07-19
feat: deprecate Network, use explicit passphrase (2)

### LINKED ISSUE #112: Network is a global singleton

Akuukis:
**See discussion at https://github.com/stellar/js-stellar-base/issues/112#issuecomment-512733154**

The following methods take optional network passphrase, and `console.warn` if it is not passed.
- static method `Keypair.master` 
- constructor `Transaction`
- constructor `TransactionBuilder` and method `TransactionBuilder.setNetworkPassphrase`

Also, tag whole class `Network` as deprecated.

Closes #112

```ts
const tx = new TransactionBuilder(account, opts)
    .addOperation(op)
    .build()  // <-- AFTER: `Network.current()` is required at this step, e.g. via `opts`.
Network.useTestNetwork()
tx.sign(kp)  // <-- BEFORE: `Network.current()` was required at this step.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,23 +12,6 @@ Do not rely on global singleton Network. The following methods take optional net
 
 Also, whole class `Network` has been deprecated.
 
-BREAKING CHANGE: `Network.use(...)` has to be set earlier than before. Example:
-
-```ts
-// BEFORE
-const txb = new TransactionBuilder(account, opts)
-    .addOperation(op)
-const tx = txb.build()
-Network.useTestNetwork()
-tx.sign(kp)  // `Network.current()` was required before this step.
-
-// AFTER
-const txb = new TransactionBuilder(account, opts)
-    .addOperation(op)
-Network.useTestNetwork()
-const tx = txb.build()  // `Network.current()` is required before this step.
-tx.sign(kp)
-```
 
 ## [v1.0.3](https://github.com/stellar/js-stellar-base/compare/v1.0.2...v1.0.3)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,33 @@
 As this project is pre 1.0, breaking changes may happen for minor version bumps.
 A breaking change will get clearly notified in this log.
 
+## Unreleased
+
+Do not rely on global singleton Network. The following methods take optional network passphrase, and `console.warn` if it is not passed.
+- static method `Keypair.master`
+- constructor `Transaction`
+- constructor `TransactionBuilder` and method `TransactionBuilder.setNetworkPassphrase`
+
+Also, whole class `Network` has been deprecated.
+
+BREAKING CHANGE: `Network.use(...)` has to be set earlier than before. Example:
+
+```ts
+// BEFORE
+const txb = new TransactionBuilder(account, opts)
+    .addOperation(op)
+const tx = txb.build()
+Network.useTestNetwork()
+tx.sign(kp)  // `Network.current()` was required before this step.
+
+// AFTER
+const txb = new TransactionBuilder(account, opts)
+    .addOperation(op)
+Network.useTestNetwork()
+const tx = txb.build()  // `Network.current()` is required before this step.
+tx.sign(kp)
+```
+
 ## [v1.0.3](https://github.com/stellar/js-stellar-base/compare/v1.0.2...v1.0.3)
 
 ### Add
--- a/docs/reference/building-transactions.md
+++ b/docs/reference/building-transactions.md
@@ -27,13 +27,13 @@ The returned transaction will contain the sequence number of the source account.
 
 
 ```js
-StellarSdk.Network.useTestNetwork();
 // StellarBase.Network.usePublicNetwork(); if this transaction is for the public network
 // Create an Account object from an address and sequence number.
 var account=new StellarBase.Account("GD6WU64OEP5C4LRBH6NK3MHYIA2ADN6K6II6EXPNVUR3ERBXT4AN4ACD","2319149195853854");
 
 var transaction = new StellarBase.TransactionBuilder(account, {
-      fee: StellarBase.BASE_FEE
+      fee: StellarBase.BASE_FEE,
+    networkPassphrase: Networks.TESTNET
     })
         // add a payment operation to the transaction
         .addOperation(StellarBase.Operation.payment({
@@ -80,9 +80,10 @@ There are 5 types of memos:
 
 ```js
 var memo = Memo.text('Happy birthday!');
-var transaction = new StellarBase.TransactionBuilder(account, { 
+var transaction = new StellarBase.TransactionBuilder(account, {
     memo: memo,
-    fee: StellarBase.BASE_FEE
+    fee: StellarBase.BASE_FEE,
+    networkPassphrase: Networks.TESTNET
 })
         .addOperation(StellarBase.Operation.payment({
                 destination: "GASOCNHNNLYFNMDJYQ3XFMI7BYHIOCFW3GJEOWRPEGK2TDPGTG2E5EDW",
@@ -142,7 +143,6 @@ var keypair = Keypair.random();
 
 
 ```js
-StellarBase.Network.useTestNetwork();
 var key1 = Keypair.fromSecret('SBK2VIYYSVG76E7VC3QHYARNFLY2EAQXDHRC7BMXBBGIFG74ARPRMNQM');
 var key2 = Keypair.fromSecret('SAMZUAAPLRUH62HH3XE7NVD6ZSMTWPWGM6DS4X47HLVRHEBKP4U2H5E7');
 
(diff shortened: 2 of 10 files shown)


src/transaction_builder.js
```diff
@@ -150,6 +153,16 @@ export class TransactionBuilder {
     return this;
   }
 
+  /**
+   * Adds a memo to the transaction.
```
abuiles: @Akuukis I think the description here is incorrect? Can you update it to reflect what it does?

abuiles: @Akuukis I think the description here is incorrect? Can you update it to reflect what it does?

Akuukis:
@abuiles Done! Anything else?

src/transaction.js
```diff
@@ -17,13 +18,29 @@ import { Keypair } from './keypair';
...
+        'Global `Network.current()` is deprecated. Please pass explicit argument instead, e.g. `new Transaction(envelope, Networks.PUBLIC)`.'
+      );
+      if (Network.current() === null) {
+        throw new Error(
+          'No network selected. Use `Network.use`, `Network.usePublicNetwork` or `Network.useTestNetwork` helper methods to select network.'
+        );
+      }
+      networkPassphrase = Network.current().networkPassphrase();
+    }
+    this.networkPassphrase = networkPassphrase;
```
abuiles: @Akuukis after thinking a little bit more about this, I'd prefer to avoid the breaking change. For now, I'd suggest adding a getter and use that to get `this._networkPassphrase` or read from Network. 

The code above will probably become something like

```
if (typeof networkPassphrase !== 'string') {
  console.warn(
    'Global `Network.current()` is deprecated. Please pass explicit argument instead, e.g. `new Transaction(envelope, Networks.PUBLIC)`.'
  );
}
this._networkPassphrase = networkPassphrase;
```

And the getter would be something like 

```
get networkPassphrase() {
  if (this._networkPassphrase) {
    return this._networkPassphrase;
  }

  console.warn(
    'Global `Network.current()` is deprecated. Please pass explicit argument instead, e.g. `new Transaction(envelope, Networks.PUBLIC)`.'
  );

  if (Network.current() === null) {
    throw new Error(
      'No network selected. Use `Network.use`, `Network.usePublicNetwork` or `Network.useTestNetwork` helper methods to select network.'
    );
  }

  return Network.current().networkPassphrase();
}
```

docs/reference/building-transactions.md
```diff
@@ -27,13 +27,13 @@ The returned transaction will contain the sequence number of the source account.
...
 ```js
-StellarSdk.Network.useTestNetwork();
 // StellarBase.Network.usePublicNetwork(); if this transaction is for the public network
 // Create an Account object from an address and sequence number.
 var account=new StellarBase.Account("GD6WU64OEP5C4LRBH6NK3MHYIA2ADN6K6II6EXPNVUR3ERBXT4AN4ACD","2319149195853854");
 
 var transaction = new StellarBase.TransactionBuilder(account, {
-      fee: StellarBase.BASE_FEE
+      fee: StellarBase.BASE_FEE,
+    networkPassphrase: Networks.TESTNET
```
abuiles: align with fee

CHANGELOG.md
```diff
@@ -3,6 +3,33 @@
 As this project is pre 1.0, breaking changes may happen for minor version bumps.
 A breaking change will get clearly notified in this log.
 
+### Unreleased
```
abuiles: this needs to be rebased

Akuukis:
@abuiles Review implemented, anything else?

abuiles:
@Akuukis Thanks! I'll release this later today :)

src/network.js
```diff
@@ -24,6 +24,7 @@ let current = null;
  * Creates a new `Network` object.
  * @constructor
  * @param {string} networkPassphrase Network passphrase
+ * @deprecated
```
abuiles: @Akuukis something I forgot, can you please add a warning in the `use` method in this class too? Notice how I put a link to this PR too so people can have more context (https://git.io/fj9fG)

```
  static use(network) {
    console.warn(
      'Global `Network.current()` is deprecated. Please pass explicit argument instead, e.g. `new Transaction(envelope, Networks.PUBLIC) - see https://git.io/fj9fG for more info`.'
    );

    current = network;
  }
```

MisterTicot:
Hi guys, I just landed here after running a

```StellarSdk.Network.useTestNetwork()```

> Global class `Network` is deprecated. Please pass explicit argument instead, e.g. `new Transaction(envelope, Networks.PUBLIC)` (see https://git.io/fj9fG for more info).

The only reason I called that function was to make it happy, though: 

> Error: No network selected. Use `Network.use`, `Network.usePublicNetwork` or `Network.useTestNetwork` helper methods to select network.

Akuukis:
@MisterTicot Hi! Hmm, I'm ready to take a look at it. Can you please share the link to your code or elaborate more on the context of error? Or, even better, extract a small snippet where it fails?

mkaliberda:
How I can set network passphrase if I use Private network?

abuiles:
@mkaliberda can you elaborate please?

MisterTicot:
@Akuukis [This](https://github.com/stellar/js-stellar-base/pull/248/commits/4e50e2674ef288e3e743ed8c493287c287de9799)

marcinx:
Recently updated to newer versions and now see warnings in our logs: `Sep 23 18:48:18 localhost stardust: Global class `Network` is deprecated. Please pass explicit argument instead, e.g. `new Transaction(envelope, Networks.PUBLIC)` (see https://git.io/fj9fG for more info).`

This appears to be triggered by `StellarSdk.Network.usePublicNetwork()`, and `StellarSdk.Network.useTestNetwork()`. 

I don't understand the motivation behind this change by the way. The error message suggests that we now need to pass the network for every single transaction issued in our application? I am a little bit confused about the warning and the suggested solution (passing the network to tx builder every time).

morleyzhi:
@marcinx The original issue, https://github.com/stellar/js-stellar-base/issues/112, lays out the reasoning behind the change. In my personal opinion, having ported some code to the new way of setting networks, the change isn't that bad and accommodates more use cases like the one in the linked issue, at the expense of a few keystrokes.

msfeldstein:
Ideally you'd use an environment variable so you can easily change networks, including between testnet and mainnet deployments

marcinx:
Okay. Thanks for the follow ups.

So basically we can scrap `StellarSdk.Network.usePublicNetwork()` and forget about the singleton.

What network is selected by default if we don't pass a passphrase to TransactonBuilder? It looks like the passphrase is being set to null if no opts.networkPassphrase is given: `this.networkPassphrase = opts.networkPassphrase || null;` What happens when the passphrase is null?

We currently don't work with Testnet. Do we need to specifically add Networks.PUBLIC (or env var, or some other way, etc.) to each transaction builder call or will it automatically select PUBLIC if no such opt is provided?

morleyzhi:
Yes, I would specifically pass Networks.PUBLIC in places that expect a network passphrase.

When you don't pass in a passphrase, it will fall back to the current Network.use*() setting:
https://github.com/stellar/js-stellar-base/blob/64785eb2e3dfece3b28e3c3767edaa6b8e09bb06/src/transaction.js#L70

Ostensibly when we remove Network.use*(), we'll throw an error wherever a network passphrase is expected but not supplied.

msfeldstein:
we definitely don't want to default things to Networks.PUBLIC, if anything we'd default to TESTNET, but its better to have stuff like this just fail so it can be fixed properly instead of continuing to work in an undefined way

marcinx:
Okay, thanks for the super fast replies! Figured it all out now. I'm definitely glad I subscribed to js-stellar-base so I can keep track of these changes more continuously.

bantustic:
For those using the sdk to connect with `stellarSdk.Network.useTestNetwork()` did not work for me and says deprecated on top of the page. I have only managed to get it to work with `StellarSdk.Networks.TESTNET`. best of luck.

PR #214, opened 2019-07-25
WIP: feat(typings): brand Secret, PublicKey, AccounId

### LINKED ISSUE #185: Brand string types
## Problem
Many data structures are represented as strings, and types don't help much because string matches string. E.g.
- at some places it is confusing whenever you need an address of existing account, random publicKey or just a string.
- when any typeguarded XDR is serialized and then parsed again, it again loses it's narrowed type
- Typescript wont catch if I'm messing up any of the following: PublicKey, Address, AccountId, Secret, Transaction hash, serialized transaction, signature, and perhaps more.

## Describe the solution you'd like
I propose to brand all those strings. That way TypeScript compiler can pick up all the problems. Something like this:
```ts
import { Brand } from 'utility-types';

type PublicKey = Brand<string, "PublicKey">
type Secret = Brand<string, "Secret">
// ...

class Keypair {
    static fromSecret(secret: Secret): PublicKey
    //...
}

const publicKey = Keypair.fromSecret(Secret)
Keypair.fromSecret(publicKey)  // <-- TypeScript will complain here.
```

This even has potential to track the narrowed types through serialization if needed.

## Describe alternatives you've considered
Less strict alternative is to use simple aliases and excessive comments.
```ts
/**
 * Do not pass where Secret is expected, please!!!
 */
type PublicKey: string
```

> Related: #176, https://github.com/DefinitelyTyped/DefinitelyTyped/pull/32025

Akuukis:
> **Work in progress: DO NOT MERGE**

See also PR at `stellar-sdk`: https://github.com/stellar/js-stellar-sdk/pull/384

Closes #185

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -3,7 +3,6 @@
 
 import { Brand } from "utility-types";
 
-
 export {};
 
 /**
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,11 +1,47 @@
 // TypeScript Version: 2.9
-
 /// <reference types="node" />
+
+import { Brand } from "utility-types";
+
+
 export {};
 
+/**
+ * 56-character long string, starts with S and passes `StrKey.isValidEd25519SecretSeed()`.
+ */
+export type Secret = Brand<string, 'Secret'>;
+
+/**
+ * 56-character long string, starts with G, passes `StrKey.isValidEd25519PublicKey()`
+ * and account does not exist on the given Network..
+ *
+ * See also `AccountId`, `PublicKey`.
+ */
+export type NonExistingAccountId = Brand<string, 'NonExistingAccountId'>;
+
+/**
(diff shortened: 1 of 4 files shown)


types/index.d.ts
```diff
@@ -1,11 +1,46 @@
...
 export {};
 
+/**
+ * 56-character long string, starts with S and passes `StrKey.isValidEd25519SecretSeed()`.
+ */
+export type Secret = Brand<string, 'Secret'>;
+
+/**
+ * 56-character long string, starts with G, passes `StrKey.isValidEd25519PublicKey()`
+ * and account does not exist on the given Network..
```
abuiles: Typo: Extra `.` --- `Network..`

abuiles:
Closing this for now since we should not mess with types which are going to be auto-generated. 

If we want to introduce this, we should do it on top of the work done in #342

Issue #188, opened 2019-05-08
Port to Typescript


morleyzhi:
This is a more direct way to address #183, and to add type safety to the library.

abuiles:
@Akuukis @charlie-wasp I've kicked off  gears for this, I'm pinging you since you both show interests on helping with this :).

I created a branch called `typescript` where we can start merging updates. The first PR was very basic, it added TS to package.json and rename all the file https://github.com/stellar/js-stellar-base/pull/216

Now we need to make `npx tsc` to actually work and setup the build/test pipeline.  If you want to start updating files, I'll suggest you leave a comment here telling which files you are working on,  so we don't overlap on each others work :). Also let's try to do small PRs so we can review and merge more easily.

Let's do this! 🚀

Akuukis:
@abuiles Here's a batch of migrations:
- #240 Hashing, including fix for wider typings.
- #241 Continued Fraction
- #242 bs58
- #243 Keypair, including new error when neither secretKey or publicKey is provided
- #244 Transaction, doesn't compile yet due missing xdr types and many `any`
- #245 Base58, including a hacky typecast from `number[]` to Buffer-alike that needs investigation
- does `src/browser.ts` need anything at all? LGTM as it is.

charlie-wasp:
@abuiles I believe, I'm pretty late for the party 😅 But it seems, that work is not over yet. How can I help here? Should I take some unchecked points from the top comment and work on them?

abuiles:
@charlie-wasp hey! Better late than never --  we still got some work to do. I'm happy if you can help on this,  we have a branch with the current changes - we can continue work off that branch.

Ryang-21:
This has been addressed in https://github.com/stellar/js-stellar-sdk/releases/tag/v16.0.0-rc.1

PR #221, opened 2019-07-30
Fix XDR namespace

abuiles:
refs #188 

I got confused initially thinking this was the typings for js-xdr, but it is actually for the autogenerated xdr code.

### DIFF
--- a/src/memo.ts
+++ b/src/memo.ts
@@ -1,7 +1,7 @@
 import isUndefined from 'lodash/isUndefined';
 import isString from 'lodash/isString';
 import clone from 'lodash/clone';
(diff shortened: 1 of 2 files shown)


src/memo.ts
```diff
@@ -229,7 +229,7 @@ export class Memo<T extends MemoType = MemoType> {
    * Returns XDR memo object.
    * @returns {xdr.Memo}
    */
-  toXDRObject(): XDRMemo | null {
+  toXDRObject(): xdr.Memo | null {
```
abuiles: @Akuukis do you know how to deal with this situation in TS? Basically we can't really add typings in `./generated/stellar-xdr_generated'` since it is dynamic and autogenerated. cc @morleyzhi

PR #240, opened 2019-08-09
fix(typescript): hashing

Akuukis:


src/hashing.ts
```diff
@@ -1,7 +1,13 @@
 import { sha256 } from 'sha.js';
 
-export function hash(data) {
+type BufferAlikeES2015 = DataView | Uint8Array | Uint8ClampedArray | Uint16Array | Uint32Array | Int8Array | Int16Array | Int32Array | Float32Array | Float64Array
```
abuiles: @Akuukis this is cool, don' this type "extend" from Buffer?

PR #243, opened 2019-08-09
fix(typescript): keypair

Akuukis:
refs #188

### DIFF
--- a/src/keypair.ts
+++ b/src/keypair.ts
@@ -23,7 +23,15 @@ import xdr from './generated/stellar-xdr_generated';
  * @param {Buffer} [keys.secretKey] Raw secret key (32-byte secret seed in ed25519`)
  */
 export class Keypair {
-  constructor(keys) {
+  /**
+   * Only 'ed25519' is implemented so far.
+   */
+  public readonly type: string
+  protected readonly _publicKey: Buffer
+  protected readonly _secretKey?: Buffer
+  protected readonly _secretSeed?: Buffer
+
+  constructor(keys: Keypair.IKeys) {
     if (keys.type !== 'ed25519') {
       throw new Error('Invalid keys type');
     }
@@ -47,12 +55,14 @@ export class Keypair {
       ) {
         throw new Error('secretKey does not match publicKey');
       }
-    } else {
+    } else if(keys.publicKey) {
       this._publicKey = Buffer.from(keys.publicKey);
 
       if (this._publicKey.length !== 32) {
         throw new Error('publicKey length is invalid');
       }
+    } else {
+      throw new Error('either secretKey or publicKey must be provided')
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/keypair.ts
```diff
@@ -190,12 +201,14 @@ export class Keypair {
...
    * @returns {Buffer}
    */
-  sign(data) {
+  public sign(data: Buffer): Buffer {
     if (!this.canSign()) {
       throw new Error('cannot sign: no secret key available');
     }
 
-    return sign(data, this._secretKey);
+    // TODO: Condition above doesn't tell that `this._secretKey` is defined in a type-safe manner.
```
abuiles: use TS-TODO

src/keypair.ts
```diff
@@ -204,14 +217,24 @@ export class Keypair {
...
-  signDecorated(data) {
+  public signDecorated(data: Buffer) {
     const signature = this.sign(data);
     const hint = this.signatureHint();
 
     return new xdr.DecoratedSignature({ hint, signature });
   }
 }
+
+export namespace Keypair {
```
abuiles: 👍

src/keypair.ts
```diff
@@ -84,62 +94,63 @@ export class Keypair {
...
   }
 
   /**
    * Create a random `Keypair` object.
    * @returns {Keypair}
    */
-  static random() {
-    const secret = nacl.randomBytes(32);
+  public static random(): Keypair {
+    const secret = nacl.randomBytes(32) as Buffer;  // TODO: Investigate `Uint8Array` vs `Buffer`.
```
abuiles: TS-TODO

src/keypair.ts
```diff
@@ -84,62 +94,63 @@ export class Keypair {
...
     return this.fromRawEd25519Seed(secret);
   }
 
-  xdrAccountId() {
+  public xdrAccountId() {
```
abuiles: @Akuukis don't we need the return type here?

src/keypair.ts
```diff
@@ -84,62 +94,63 @@ export class Keypair {
...
     return new xdr.AccountId.publicKeyTypeEd25519(this._publicKey);
   }
 
-  xdrPublicKey() {
+  public xdrPublicKey() {
```
abuiles: same here, don't we need the return type?

src/keypair.ts
```diff
@@ -204,14 +217,24 @@ export class Keypair {
    * @param {Buffer} signature Signature
    * @returns {boolean}
    */
-  verify(data, signature) {
+  public verify(data: Buffer, signature: Buffer): boolean {
     return verify(data, signature, this._publicKey);
   }
 
-  signDecorated(data) {
+  public signDecorated(data: Buffer) {
```
abuiles: @Akuukis for the return value, check https://github.com/stellar/js-stellar-base/blob/typescript/src/%40types/xdr.ts#L28 --- see https://github.com/stellar/js-stellar-base/blob/typescript/src/operations/index.ts#L5

You can do  `import { xdr as xdrDef } from '../@types/xdr';`

PR #244, opened 2019-08-09
fix(typescript): transaction

Akuukis:
refs #188


src/transaction.ts
```diff
@@ -191,15 +203,16 @@ export class Transaction {
...
+    const currentNetwork = Network.current()
+    if (currentNetwork === null) {
       throw new Error(
         'No network selected. Use `Network.use`, `Network.usePublicNetwork` or `Network.useTestNetwork` helper methods to select network.'
       );
     }
 
     return Buffer.concat([
-      Network.current().networkId(),
+      currentNetwork.networkId(),
```
abuiles: damm, we need to rebase this against master to support your update on network

PR #245, opened 2019-08-09
fix(typescript): base58

Akuukis:
refs #188

### DIFF
--- a/src/base58.ts
+++ b/src/base58.ts
@@ -9,7 +9,7 @@ const versionBytes = {
   seed: 0x21 // decimal 33
 };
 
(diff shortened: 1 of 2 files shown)


src/base58.ts
```diff
@@ -30,7 +30,8 @@ export function decodeBase58Check(versionByteName, encoded) {
     );
   }
 
-  const expectedChecksum = calculateChecksum(payload);
+  // TODO: Investigate. This looks like a upstream typing bug, because empirically it works anyways.
```
abuiles: Can you use TS-TODO so it's easier for us to identify the todos related with this migration?

PR #247, opened 2019-08-13
fix(typescript): xdr - refactor namespaces

Akuukis:
Ref #188 

This is experiment how to deal with XDR definitions. Please see `src/asset.ts` - it's finished as part of this change. How does this direction looks to you?

TL;DR:
1. we need two seperate files: `@types/xdr.d.ts` for dependency, and `src/xdr_definitions.ts` for definitions.
2. definitions are types-only although public-facing, just like `horizon_api.ts` or `server_api.ts` at `stellar-sdk`.
3. we need to work a lot on it, see how operations files got all red now..


src/generated/stellar-xdr_generated.ts
```diff
@@ -3881,4 +3881,6 @@ xdr.struct("HmacSha256Mac", [
 ]);
 
 });
-export default types;
+
+export default xdr;
+export { XdrNS as Xdr } from '../xdr_definitions';
```
abuiles: @Akuukis btw, I'm iffy about touching this file mostly because it is auto-generated so we don't have control over it

abuiles:
@Akuukis can you rebase?

Akuukis:
I will rebase. But for types to work we need to make *some* changes to how the generated file works. I will look into how much magic we can accomplish with `d.ts` to go around that

Akuukis:
@abuiles the `stellar-xdr_generated.ts` is reverted back to `.js`, and all types are placed in `.d.ts` next to it. It types nicely like this, although someone need to add many xdr definitions later. PR is ready for next review!

abuiles:
@Akuukis awesome, thanks!

PR #294, opened 2020-01-28
Remove node engine restriction

rajababu3:
This is working with each node engine so remove the restriction.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,9 +35,6 @@
   "keywords": [
     "stellar"
   ],
-  "engines": {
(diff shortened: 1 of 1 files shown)


abuiles:
Thanks!

ruslan-rt:
@abuiles Since you dropped support for Node 6 and now require Node 10 (#255), you should keep the restriction, I suggest

@rajababu3 This is maybe working with each node engine now, luckily and unintentionally, but it can change in future, because the last supported version is Node 10

PR #300, opened 2020-02-18
Use console.trace to get line num in Networks.use

pselden:
Fixes https://github.com/stellar/js-stellar-sdk/issues/415


pselden:
Note -- this is a little noisy because it includes the full stack trace.

PR #303, opened 2020-02-19
Fix crash on partially defined time bounds

andywer:
Fix `Cannot read property 'toString' of undefined` error. Did occur when either `timebounds.minTime` or `timebounds.maxTime` were set, but not both.

PS: Need to test this fix some more.


andywer:
Sure, I will look into that 👍

Sorry for the delayed response.

andywer:
Changed the fix and added a regression test.

PR #305, opened 2020-02-26
WIP: feat(typescript): add XDR typings

Akuukis:
WORK IN PROGRESS

PR #307, opened 2020-02-26
Add optional format to TransactionBuilder xdr constructor

### LINKED ISSUE #306: TS Errors trying to build envelope from xdr string
**Describe the bug**
I have a base64 encoded xdr string representing a TransactionEnvelope.  I'd like to instantiate a TransactionEnvelope object using 
```const envelope = xdr.TransactionEnvelope.fromXDR(tx, "base64")```
where `tx` is a base64 encoded string.  The typescript error detection marks this as an error 
```Expected 1 arguments, but got 2.```
when trying to pass the `"base64"` string.  Without passing that string, it can't properly decode the xdr.

The error not only causes red errors in the editor, but also causes some build tools, such as stencil to fail.

**What version are you on?**
2.1.6

**To Reproduce**
Run the above line of code in a typescript aware editor like vscode.  

**Expected behavior**
There should be no error, code should compile correctly

msfeldstein:
Fixes #306 

TransactionBuilder.fromXDR type definition was missing an optional string argument to specify the format of the xdr, as seen here: https://github.com/stellar/js-xdr/blob/master/src/io-mixin.js#L18

I am unclear if all the other classes fromXDR should also have this argument, anyone know?

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -579,7 +579,7 @@ export namespace xdr {
     static fromXDR(xdr: Buffer): Memo;
   }
   class TransactionEnvelope extends XDRStruct {
-    static fromXDR(xdr: Buffer): TransactionEnvelope;
+    static fromXDR(xdr: Buffer, format?: string): TransactionEnvelope;
   }
   class DecoratedSignature extends XDRStruct {
     static fromXDR(xdr: Buffer): DecoratedSignature;


abuiles:
cc @Akuukis

msfeldstein:
honestly i can't really tell what inherits from what...  I'm gonna leave
this to @akuukis or someone who understands whats going on with xdr, it
seems like its a much bigger problem than just adding this field

Akuukis:
Yes, it should have also other stuff from IOMixin. Here's WIP on this - https://github.com/stellar/js-stellar-base/pull/305

It has a lot of typings already, most likely most of them correct, but I'm stuck to finish it without help from someone who understands XDR better.

abuiles:
@Akuukis that depends on the TS branch which includes a bigger change that just this typings :( maybe is there a way we can bring only the specific typings to fix this particular issue?

Akuukis:
There is a way but it's work, and that's gonna be wasted once TS branch will be merged. So you have to decide between "wait on TS branch" vs "let's quickfix this", where IMO the latter sadly implies that already kind-of abandoned TS branch is abandoned for real.

I'd prefer to finish TS branch, and I'm still willing to contribute, but it's blocked by either your help on XDR part or by allowing me to merge whatever I understand about XDR part myself and we fix problems later.

msfeldstein:
What are the XDR questions? I think its more important to get proper typing for all our XDR classes than to fix cases individually like this

Akuukis:
@msfeldstein Please see https://github.com/stellar/js-stellar-base/pull/305 - it has most of XDR typings in place, but as I remember it had two major problems:
1. I don't fully understand XDR library, and it is written with mixins while TS has only classes. I did my best to reverse-engineer it but I have a bad feeling about this anyways, so it requires a review.
2. That PR has several outstanding compiler errors which needs to investigated. I don't know what to do with them, and there are 3 options:
    1. I did XDR typings wrong but can't understand how it's supposed to be instead (need help).
    2. I did XDR typings right and that's actually a potential bug (need confirmation).
    3. I did XDR typings kind-of wrong/right, as it's not supposed to work but turns out to be working fine (need decision whenever that's actually 1 or 2)

I suggest that we continue this thread in comments of that PR!

abuiles:
This will be fixed here https://github.com/stellar/js-stellar-base/pull/321/files#diff-88bec6beae84369c0376b56b3bb88fe1R575

PR #315, opened 2020-03-30
Fix multiple timebounds related bugs.

### LINKED ISSUE #312: TransactionBuilder.setTimeout(0) makes .build() fail
**Describe the bug**
TransactionBuilder.setTimeout(0) doesn't work, and leads to an error at L221, where this.timebounds.maxTime can't be converted to a string because it perhaps doesn't exist?

**What version are you on?**
2.1.6

**To Reproduce**
create a builder
add your operations to it
call setTimeout(0)
call .build()

**Expected behavior**

I'd expect build() to succeed.

**Additional context**
[...]/node_modules/stellar-base/lib/transaction_builder.js:257
        this.timebounds.maxTime = _jsXdr.UnsignedHyper.fromString(this.timebounds.maxTime.toString());
                                                                                          ^

TypeError: Cannot read property 'toString' of undefined
    at TransactionBuilder.build ([...]/node_modules/stellar-base/lib/transaction_builder.js:257:91)

abuiles:
There were multiple issues around timebounds logic like setting `TimeoutInfinite` or trying to pass partially defined timebounds (which is not valid), there were some validations for this but the code was buggy.

This PR:

- Fixes timebounds validation, making sure that both `minTime` and `maxTime` are defined.
- Fixes `setTimeout(TimeoutInfinite)` - this was broken.

This fixes #312 and is related with #303

### DIFF
--- a/test/unit/transaction_builder_test.js
+++ b/test/unit/transaction_builder_test.js
@@ -260,7 +260,63 @@ describe('TransactionBuilder', function() {
       done();
     });
   });
-
+  describe('timebounds', function(){
+    it('requires maxTime', function(){
+      let source = new StellarBase.Account(
+        'GCEZWKCA5VLDNRLN3RPRJMRZOX3Z6G5CHCGSNFHEYVXM3XOJMDS674JZ',
+        '0'
+      );
+      expect(() => {
+        new StellarBase.TransactionBuilder(source, {
+          timebounds: {
+            minTime: '0'
+          },
+          fee: 100
+        }).build();
+      }).to.throw('TimeBounds has to be set or you must call setTimeout(TimeoutInfinite).');
+    });
+    it('requires minTime', function(){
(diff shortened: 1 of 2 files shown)


test/unit/transaction_builder_test.js
```diff
@@ -260,7 +260,63 @@ describe('TransactionBuilder', function() {
...
+      );
+      expect(() => {
+        new StellarBase.TransactionBuilder(source, {
+          timebounds: {
+            minTime: '1',
+            maxTime: '10'
+          },
+          fee: 100
+        }).build();
+      }).to.not.throw('TimeBounds has to be set or you must call setTimeout(TimeoutInfinite).');
```
tamirms: is it possible to assert that it doesn't throw any exception instead of testing for a specific exception?

abuiles: good catch! no exception should be thrown.

PR #317, opened 2020-04-07
Protocol 13 support

### LINKED ISSUE #320: Add protocol 13 support to with backwards compatibility with protocol 12

abuiles:
## What

Add protocol 13 support with backwards compatibility support for protocol 12.

## Why

In order to provide a seamless transition between protocol 12 and 13, we need to release a new version of the SDK that can read  and generate transactions which will work in both protocols.

This pull request:

- Updates the XDR definitions with protocol 13.
- Extends `Transaction` to work with `TransactionV1Envelope` and `TransactionV0Envelope`.
- Generates `TransactionEnvelope` which can be read by protocol 12 or protocol 13.
- Reads `TransactionEnvelope` generated by protocol 12 or protocol 13.
- Adds backward compatibility support for [CAP0018](https://github.com/stellar/stellar-protocol/blob/f01c9354aaab1e8ca97a25cf888829749cadf36a/core/cap-0018.md)
- Adds partial support for [CAP0015](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0015.md).
   - The `Transaction` class which offers an abstraction around `xdr.Transaction` doesn't support `FeeBumpTransaction` yet, however, consumers of this library can create and read fee bump transactions using `xdr.FeeBumpTransaction`.

Fix #320 

## Breaking changes

- If you are only relying on the high level methods of this library,  upgrading to the new version of the library should not bring any breaking changes. 
- If you are relying on `xdr.TransactionEnvelope` or `transaction.toEnvelope()` then there are breaking changes and you'll have to update your code to match the new API around `xdr.TransactionEnvelope`.

### DIFF
--- a/src/operations/allow_trust.js
+++ b/src/operations/allow_trust.js
@@ -11,7 +11,7 @@ import { StrKey } from '../strkey';
  * @param {object} opts Options object
  * @param {string} opts.trustor - The trusting account (the one being authorized)
  * @param {string} opts.assetCode - The asset code being authorized.
- * @param {boolean} opts.authorize - True to authorize the line, false to deauthorize.
+ * @param {boolean|number} opts.authorize - True to authorize the line, false to deauthorize.
  * @param {string} [opts.source] - The source account (defaults to transaction source).
  * @returns {xdr.AllowTrustOp} Allow Trust operation
  */
@@ -30,7 +30,13 @@ export function allowTrust(opts) {
   } else {
     throw new Error('Asset code must be 12 characters at max.');
   }
-  attributes.authorize = opts.authorize;
+
+  if (typeof opts.authorize === 'boolean') {
+    attributes.authorize = +opts.authorize;
+  } else {
+    attributes.authorize = opts.authorize;
+  }
+
   const allowTrustOp = new xdr.AllowTrustOp(attributes);
 
   const opAttributes = {};
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -42,7 +42,6 @@ export class Transaction {
     const txe = envelope.value();
     let sourceAccount;
     this._envelopeType = envelope.switch();
-
     switch (this._envelopeType) {
       case xdr.EnvelopeType.envelopeTypeTxV0():
         sourceAccount = txe.tx().sourceAccountEd25519();
@@ -286,7 +285,7 @@ export class Transaction {
         break;
       case xdr.EnvelopeType.envelopeTypeTx():
         envelope = new xdr.TransactionEnvelope.envelopeTypeTx(
-          new xdr.TransactionEnvelope({ tx, signatures })
+          new xdr.TransactionV1Envelope({ tx, signatures })
         );
         break;
       default:
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -39,14 +39,28 @@ export class Transaction {
     }
     this._networkPassphrase = networkPassphrase;
 
+    const txe = envelope.value();
+    let sourceAccount;
+    this._envelopeType = envelope.switch();
+
+    switch (this._envelopeType) {
+      case xdr.EnvelopeType.envelopeTypeTxV0():
+        sourceAccount = txe.tx().sourceAccountEd25519();
+        break;
+      case xdr.EnvelopeType.envelopeTypeTx():
+        sourceAccount = txe
+          .tx()
+          .sourceAccount()
+          .ed25519();
+        break;
+      default:
+        // TODO: Add support for FeeBumpTransaction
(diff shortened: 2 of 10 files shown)


src/transaction.js
```diff
@@ -39,14 +39,27 @@ export class Transaction {
...
+        break;
+      case xdr.EnvelopeType.envelopeTypeTx():
+        sourceAccount = txe
+          .tx()
+          .sourceAccount()
+          .ed25519();
+        break;
+      default:
+        // TODO: Add support for FeeBumpTransaction
+        throw new Error('Invalid transaction.');
```
vcarl, line 57: This could be more descriptive, is FeeBump the only one that should fail? Might be nice to have this say something like, 
> Invalid transaction, fee bump transactions are not supported yet. Additional details at `<github issue url>`

src/transaction.js
```diff
@@ -243,7 +275,22 @@ export class Transaction {
...
+          new xdr.TransactionV0Envelope({ tx, signatures })
+        );
+        break;
+      case xdr.EnvelopeType.envelopeTypeTx():
+        envelope = new xdr.TransactionEnvelope.envelopeTypeTx(
+          new xdr.TransactionV1Envelope({ tx, signatures })
+        );
+        break;
+      default:
+        throw new Error('Invalid transaction');
```
vcarl, line 292: Ditto re: descriptive error

test/unit/transaction_test.js
```diff
@@ -37,16 +43,7 @@ describe('Transaction', function() {
     done();
   });
 
-  beforeEach(function() {
-    StellarBase.Network.useTestNetwork();
-  });
-
-  afterEach(function() {
-    StellarBase.Network.use(null);
-  });
```
vcarl, line 46: 🎉

src/transaction.js
abuiles: @vcarl this is temporal, we add fee bump support on this other PR  #321

src/operations/allow_trust.js
```diff
@@ -30,7 +30,13 @@ export function allowTrust(opts) {
   } else {
     throw new Error('Asset code must be 12 characters at max.');
   }
-  attributes.authorize = opts.authorize;
+
+  if (typeof opts.authorize === 'boolean') {
+    attributes.authorize = +opts.authorize;
```
tamirms, line 35: I think it would be better to reference the xdr constant

src/transaction_builder.js
```diff
@@ -206,15 +210,22 @@ export class TransactionBuilder {
...
+    );
+
     attrs.timeBounds = new xdr.TimeBounds(this.timebounds);
 
     const xtx = new xdr.Transaction(attrs);
     xtx.operations(this.operations);
 
-    const xenv = new xdr.TransactionEnvelope({ tx: xtx });
+    const xenv = new xdr.TransactionEnvelope.envelopeTypeTx(
+      new xdr.TransactionV1Envelope({ tx: xtx })
```
tamirms, line 226: shouldn't we be creating v0 transactions?

ire-and-curses: Shouldn't this be TransactionV0 here to return protocol 12 envelopes? And then a TODO to change to V1 in the subsequent SDK release?

abuiles: @tamirms yeah, turns out I did that but forgot to push that commit. It is now fixed.

ire-and-curses: TODO to move this to protocol 13 in next release: #322

src/operations/allow_trust.js
abuiles: this particular operation is casting the value from boolean to int. People can still refer to the xdr constant in their code.

tamirms: @abuiles there is an implicit assumption that the authorized flag is 1. I think it would be better rewrite the code as:

```javascript
  if (typeof opts.authorize === 'boolean') {
    if (opts.authorize) {
        attributes.authorize = xdr.TrustLineFlags.authorizedFlag;
    } else {
        attributes.authorize = 0;
    }
  } else {
    attributes.authorize = opts.authorize;
  }
```

types/index.d.ts
```diff
@@ -333,7 +333,8 @@ export namespace Operation {
   interface AllowTrust extends BaseOperation<OperationType.AllowTrust> {
     trustor: string;
     assetCode: string;
-    authorize: boolean | undefined;
+    // this is a boolean or a number so that it can support protocol 12 or 13
+    authorize: boolean | number | undefined;
```
abuiles, line 337: @ire-and-curses @tamirms I'll actually do the same for this, add the breaking change, making it a number.

PR #321, opened 2020-04-10
Add support for CAP0015 (FeeBump).

### LINKED ISSUE #319: Implement CAP 15 (Fee bump) support in JS BASE

abuiles:
## What

Add support for CAP0015.

## Why

Protocol 13 introduces a new kind of transaction which allows other accounts to pay the fee for a given transaction. 

This pull request adds a new helper functions which makes it easy for consumers of this library to create a fee bump transaction.

This PR: 

- Add `.buildFeeBumpTransaction` to  `TransactionBuilder` which makes it easy to create `FeeBumpTransaction`. 
- Adds a new helper function to the `Transaction` class which allows consumer to know if the transaction is a `FeeBump` transaction or not.
- Adds a feature flag which allow  consumers of this library to create V1 (protocol 13) transactions using the `TransactionBuilder`.
- Extends TS definitions.

Fixes #319 

## How to create FeeBump transactions? 

Assuming you have a valid version of stellar-core and Horizon running with protocol 13, you can use the following example to create a fee bump transaction (**replace keypairs**):

```js
let sourceKP = Keypair.fromSecret("SAQZZ7K4YUOEX3MMLX5YM4VCK4E7NSKRODKNPNYCMH44JGGXNLUDNZH7");
// account to which the inner transactions belongs to
let account = await server.loadAccount(sourceKP.publicKey())
let PASS_PHRASE = "Standalone Network ; February 2017";
let destination = 'GDQERENWDDSQZS7R7WKHZI3BSOYMV3FSWR7TFUYFTKQ447PIX6NREOJM';
let amount = '20';
let innerTX = new TransactionBuilder(account, {
  fee: 100,
  networkPassphrase: PASS_PHRASE,
  timebounds: {
    minTime: 0,
    maxTime: 0,
  },
  v1: true,
})
    .addOperation(
      Operation.payment({ 
        destination,
        asset: Asset.native(),
        amount
      })
    )
    .addMemo(Memo.text('Happy birthday!'))
    .build()
innerTX.sign(sourceKP);

// Keypair for account paying for the FeeBumpTX
let feeSourceKP = Keypair.fromSecret("SB7ZMPZB3YMMK5CUWENXVLZWBK4KYX4YU5JBXQNZSK2DP2Q7V3LVTO5V");
let bumpFee = "25000000";
let transaction = TransactionBuilder.buildFeeBumpTransaction(
  feeSourceKP,
  bumpFee,
  innerTX,
  PASS_PHRASE
);
transaction.sign(feeSourceKP)

console.log("waiting for hash (fee bump tx hash)", transaction.hash().toString('hex'))
let result = await server.submitTransaction(transaction);
```

## Breaking changes

- The attribute `fee` in `Transaction` instances changed from `number` to `string`.
- The const `BASE_FEE` changed from `number` to `string`.
- The option `fee` passed to  `new TransactionBuilder({fee: ..})` changed from `number` to `string`


src/transaction.js
```diff
@@ -39,43 +39,61 @@ export class Transaction {
...
+        .tx();
+      this.source = StrKey.encodeEd25519PublicKey(tx.sourceAccount().ed25519());
+      this.feeSource = StrKey.encodeEd25519PublicKey(sourceAccount);
+    } else {
+      this.source = StrKey.encodeEd25519PublicKey(sourceAccount);
+      this.fee = tx.fee();
+    }
+
+    this._memo = tx.memo();
+    this.sequence = tx.seqNum().toString();
```
2opremio, line 84: It's hard to see what's the difference between a feebump and non-feebump. I think it would be more readable (and only slightly less efficient) to initialize the fields inconditionally as a non-feebump and amend them if it's a feebump.

src/transaction.js
```diff
@@ -258,11 +276,21 @@ export class Transaction {
...
+        tx
+      );
+    }
+
     const txSignature = new xdr.TransactionSignaturePayload({
       networkId: xdr.Hash.fromXDR(hash(this.networkPassphrase)),
-      taggedTransaction: new xdr.TransactionSignaturePayloadTaggedTransaction.envelopeTypeTx(
-        tx
-      )
+      taggedTransaction
```
2opremio, line 293: This is probably me not knowing JS well, but shouldn't this have a key?

abuiles: oh it doesn't - in JS, if the key has the same name as the variable you are assigning, it will get merged into the object.

2opremio: ok, thanks!!

src/transaction.js
abuiles: @2opremio good point!

src/transaction.js
```diff
@@ -39,43 +39,64 @@ export class Transaction {
...
+      case xdr.EnvelopeType.envelopeTypeTxFeeBump():
+        sourceAccount = txEnvelope
+          .tx()
+          .feeSource()
+          .ed25519();
+        break;
       default:
-        // TODO: Add support for FeeBumpTransaction
-        throw new Error('Invalid transaction.');
+        throw new Error('Unknown TransactionEnvelope');
```
ire-and-curses, line 62: Obviously this should never happen. But if it did: can the error say something about the envelope that was received, e.g. the integer type

src/transaction.js
```diff
@@ -39,43 +39,64 @@ export class Transaction {
...
         maxTime: timeBounds.maxTime().toString()
       };
     }
-
-    const operations = this.tx.operations() || [];
+    const operations = tx.operations() || [];
     this.operations = map(operations, (op) => Operation.fromXDRObject(op));
 
-    const signatures = txe.signatures() || [];
+    const signatures = txEnvelope.signatures() || [];
```
abuiles, line 99: expose inner signers

src/transaction.js
```diff
@@ -288,13 +319,27 @@ export class Transaction {
...
+        throw new Error('Unknown TransactionEnvelope');
     }
 
     return envelope;
   }
 
+  /**
+   * isFeeBump returns true if the transaction represents a FeeBumpTransaction
+   * @returns {boolean}
+   * @ignore tell jsdoc to not show this method for now
```
ire-and-curses, line 337: Add TODO to remove ignore in next SDK release

src/transaction.js
```diff
@@ -304,4 +349,34 @@ export class Transaction {
...
   }
+
+  /**
+   * Builds a FeeBumpTransaction
+   * @param {StrKey} feeSource - The account paying for the transaction.
+   * @param {string} fee - The maximum fee the account is willing to pay for this transaction.
+   * @param {TransactionEnvelope} innerTx - The Transaction to be bumped by this transaction.
+   * @param {string} networkPassphrase - networkPassphrase of the target stellar network (e.g. "Public Global Stellar Network ; September 2015").
+   * @returns {Transaction}
+   * @ignore tell jsdoc to not show this method for now
```
ire-and-curses, line 360: Add TODO to remove this in next release

test/unit/transaction_test.js
```diff
@@ -456,6 +456,76 @@ describe('Transaction', function() {
...
+      let transaction = StellarBase.Transaction.buildFeeBumpTransaction(
+        feeSource,
+        bumpFee,
+        innerTX.toEnvelope(),
+        networkPassphrase
+      );
+      transaction.sign(feeSource);
+      const expectedXDR =
+        'AAAABQAAAADgSJG2GOUMy/H9lHyjYZOwyuyytH8y0wWaoc596L+bEgAAAAABfXhAAAAAAgAAAABzdv3ojkzWHMD7KUoXhrPx0GH18vHKV0ZfqpMiEblG1gAAAGQAAAAAAAAACAAAAAEAAAAAAAAAAAAAAAAAAAAAAAAAAQAAAA9IYXBweSBiaXJ0aGRheSEAAAAAAQAAAAAAAAABAAAAAOBIkbYY5QzL8f2UfKNhk7DK7LK0fzLTBZqhzn3ov5sSAAAAAAAAAASoF8gAAAAAAAAAAAERuUbWAAAAQK933Dnt1pxXlsf1B5CYn81PLxeYsx+MiV9EGbMdUfEcdDWUySyIkdzJefjpR5ejdXVp/KXosGmNUQ+DrIBlzg0AAAAAAAAAAei/mxIAAABAtRbc7GZeE8cJyZ+R7XGTZKy8s1lTp0rViCHhUa7pDzkaWUI0WUQ5mjsc8+/YhJUXAlPVTbeEbiM6knnPBUFwDw==';
+
```
tamirms, line 503: I think it's worth adding asserts on the fee bump transaction signatures and the inner transaction signatures

test/unit/transaction_test.js
```diff
@@ -456,6 +456,76 @@ describe('Transaction', function() {
...
+      expect(transaction.isFeeBump()).to.equal(true);
+      expect(
+        transaction
+          .toEnvelope()
+          .toXDR()
+          .toString('base64')
+      ).to.be.equal(expectedXDR);
```
tamirms, line 510: can you decode `expectedXDR` and assert that the xdr properties match innerTx, bumpFee, and feeSource?

src/transaction.js
```diff
@@ -304,4 +349,34 @@ export class Transaction {
...
+   */
+  static buildFeeBumpTransaction(feeSource, fee, innerTx, networkPassphrase) {
```
tamirms, line 362: the fee parameter here is not consistent with `TransactionBuilder`. In `TransactionBuilder` you take a `baseFee` parameter and the compute the fee by multiplying `baseFee` with the number of operations. I think you should change `buildFeeBumpTransaction()` so that it takes a `baseFee` parameter and use that to calculate the `fee` set in the tx envelope.

I think you should also validate the `baseFee` parameter:

https://github.com/stellar/stellar-protocol/blob/master/core/cap-0015.md#validity
> The fee rate for E is at least the fee rate of F, otherwise return txINSUFFICIENT_FEE

Here are details on how the fee is calculated for fee bump transactions:

https://github.com/stellar/stellar-protocol/blob/master/core/cap-0015.md#fee-rate
> A fee-bump transaction has an effective number of operations equal to one plus the number of operations in the inner transaction. Correspondingly, the minimum fee for the fee-bump transaction is one base fee more than the minimum fee for the inner transaction. Similarly, the fee rate (see CAP-0005) is normalized by one plus the number of operations in the inner transaction rather than the number of operations in the inner transaction alone.

types/index.d.ts
```diff
@@ -511,16 +512,19 @@ export class Transaction<
...
   signatureBase(): Buffer;
   signHashX(preimage: Buffer | string): void;
   toEnvelope(): xdr.TransactionEnvelope;
   toXDR(): string;
+  static buildFeeBumpTransaction(feeSource: Keypair, fee: string, innerTx: xdr.TransactionEnvelope, networkPassphrase: string): Transaction;
 
   operations: TOps;
   sequence: string;
-  fee: number;
+  fee: number|string;
```
tamirms, line 525: seems weird to have the type be a number or a string. is the reason for keeping the number option for backwards compatibility?

src/transaction.js
abuiles: Done in #323

src/transaction.js
abuiles: #323

src/transaction.js
abuiles: @tamirms good point! I did this intentionally to let people calculate this value themselves. I'll add the validations :)

types/index.d.ts
abuiles: right, this is for backwards compatibility since we'll release this while protocol 12 is still running.

tamirms: it's possible to represent the protocol 12 fee as a string. however, users of the js sdk must be prepared for the new type. since this is a major version bump you could make the argument that it's ok to change the type from `number` to `string`. but since javascript is not statically typed I think users of the sdk will not realize the type change unless they look at the changelog

abuiles: right, this is focus only on TS users -- so having this will allow them to start preparing their code for both scenarios. I think eventually we can remove the typing once we are only returning `string`. 

This is similar to what we did to allow the transition from int to string on int64 values.

abuiles: @tamirmfs after thinking more about this, I think we don't need backwards compatibility for this field - we can change it to `string` and let it be a breaking change.

ire-and-curses: I support that - getting this changed by downstream users sooner rather than later is a win

src/transaction.js
abuiles: Fixed

test/unit/transaction_test.js
abuiles: Fixed

test/unit/transaction_test.js
abuiles: Fixed

src/transaction.js
abuiles: updated

src/transaction.js
```diff
@@ -304,4 +350,50 @@ export class Transaction {
...
+  /**
+   * Builds a FeeBumpTransaction
+   * @param {StrKey} feeSource - The account paying for the transaction.
+   * @param {string} baseFee - The max fee willing to pay per operation in inner transaction (**in stroops**). Required.
+   * @param {TransactionV1Envelope} innerTxEnvelope - The TransactionV1Envelope to be bumped by this transaction.
+   * @param {string} networkPassphrase - networkPassphrase of the target stellar network (e.g. "Public Global Stellar Network ; September 2015").
+   * @returns {Transaction}
+   * @ignore tell jsdoc to not show this method for now
+   */
+  static buildFeeBumpTransaction(
```
tamirms, line 363: what's the rationale for including this function in the `Transaction` class instead of `TransactionBuilder`? also would it make sense for `buildFeeBumpTransaction()` to take an `innerTx` argument of type `Transaction` instead of `innerTxEnvelope` which has type `TransactionV1Envelope`?

src/transaction.js
```diff
@@ -304,4 +350,50 @@ export class Transaction {
...
+    feeSource,
+    baseFee,
+    innerTxEnvelope,
+    networkPassphrase
+  ) {
+    const BASE_FEE = '100';
+    const innerOps = innerTxEnvelope.tx().operations().length;
+    const minFee = new BigNumber(BASE_FEE).mul(innerOps + 1);
```
tamirms, line 371: `minFee` is calculated using the fee from `innerTxEnvelope`

src/transaction.js
abuiles: It makes sense to have this on the builder. While testing it felt to me like this responsibility should be here mostly since I was testing  how `Transaction` behaves when you pass a fee bump envelope. However, I think we can move this back to the build and split the tests.

src/transaction.js
```diff
@@ -288,13 +320,27 @@ export class Transaction {
...
         );
         break;
+      case xdr.EnvelopeType.envelopeTypeTxFeeBump():
+        envelope = new xdr.TransactionEnvelope.envelopeTypeTxFeeBump(
+          new xdr.FeeBumpTransactionEnvelope({ tx, signatures })
+        );
+        break;
       default:
-        throw new Error('Invalid transaction');
+        throw new Error(`Invalid EnvelopeType: ${this.envelopeType}.`);
```
tamirms, line 329: should `this._envelopeType` be used instead of `this.envelopeType`? I don't think `this.envelopeType` is defined anywhere

abuiles: 👍

src/transaction_builder.js
```diff
@@ -219,19 +223,99 @@ export class TransactionBuilder {
...
+      );
+    }
+
+    const fee = base.mul(innerOps + 1);
+
+    const tx = new xdr.FeeBumpTransaction({
+      feeSource: feeSource.xdrAccountId(),
+      fee: xdr.Int64.fromString(fee.toString()),
+      innerTx: xdr.FeeBumpTransactionInnerTx.envelopeTypeTx(
+        innerTx.toEnvelope().value()
```
tamirms, line 302: shouldn't we be extracting the v1 arm of innerTx? I think it would also be good to validate that `innerTx` is a v1 transaction (not a v0 tx or a fee bump tx) since the error message we would provide is easier to understand then the one generated by the library crashing

abuiles: We can get the value directly or through the discriminant, the reason I went with the value was to do the validation at the end when converting to an XDR.  I'll update it to explicitly use v1 and validate that the tx is a v1 tx envelope

abuiles: handled in e0d734f

PR #325, opened 2020-04-14
Add support for CAP 0027 (MuxedAccounts)

### LINKED ISSUE #324: Implement CAP 27 support

abuiles:
## What

Add support for CAP0027.

## Why

Protocol 13 introduces multiplexed accounts, replacing some fields with the union `xdr.MuxedAccount` where `xdr.PublicKey` was previously used.

This pull requests:

- Extends `Keypair` with a new method `.xdrMuxedAccount` which creates a new `MuxedAccount` using the `ed25519` discriminant.
- Updates all the operations which changed their attribute from `PublicKey` to `MuxedAccount`.

Fixes #324 

## Breaking changes

This update doesn't include breaking changes for users of this library. However, if you are relying on XDRs, you'll get now a `MuxedAccount` where a `PublicKey` was returned. 

As long as you are not checking for the instance type, your code should continue to work since both `MuxedAccount` and `PublicKey` have the discriminant `ed25519` which returns the same data type. 

## What is not implemented in this pull request
- SEP0023 support - I'll handle that in a different PR.
- Passing an envelope to `new Transaction(envelope)` which contains `MuxedAccount`s with the `med25519` discriminant, won't work. For now, we expect an envelope to contain muxed accounts with the `ed25519` discriminant.  I'll update this code to be able to read the `med25519` discriminant in the same PR where I'll add SEP0023 support.

### DIFF
--- a/test/unit/keypair_test.js
+++ b/test/unit/keypair_test.js
@@ -1,52 +1,64 @@
-
 describe('Keypair.contructor', function() {
-  it("fails when passes secret key does not match public key", function() {
-    let secret = "SD7X7LEHBNMUIKQGKPARG5TDJNBHKC346OUARHGZL5ITC6IJPXHILY36";
+  it('fails when passes secret key does not match public key', function() {
+    let secret = 'SD7X7LEHBNMUIKQGKPARG5TDJNBHKC346OUARHGZL5ITC6IJPXHILY36';
     let kp = StellarBase.Keypair.fromSecret(secret);
 
     let secretKey = kp.rawSecretKey();
     let publicKey = StellarBase.StrKey.decodeEd25519PublicKey(kp.publicKey());
     publicKey[0] = 0; // Make public key invalid
 
-    expect(() => new StellarBase.Keypair({type: 'ed25519', secretKey, publicKey})).to.throw(/secretKey does not match publicKey/)
+    expect(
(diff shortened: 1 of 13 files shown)


test/unit/keypair_test.js
```diff
@@ -1,52 +1,64 @@
-
 describe('Keypair.contructor', function() {
-  it("fails when passes secret key does not match public key", function() {
-    let secret = "SD7X7LEHBNMUIKQGKPARG5TDJNBHKC346OUARHGZL5ITC6IJPXHILY36";
+  it('fails when passes secret key does not match public key', function() {
```
abuiles, line 2: oh, this is replaced automatically by some linter on this repo -- ignore this changes

test/unit/keypair_test.js
```diff
@@ -85,67 +96,103 @@ describe('Keypair.fromBase58Seed', function() {
...
 describe('Keypair.random', function() {
-
-  it("creates a keypair correctly", function() {
+  it('creates a keypair correctly', function() {
     let kp = StellarBase.Keypair.random();
     expect(kp).to.be.instanceof(StellarBase.Keypair);
   });
+});
 
+describe('Keypair.xdrMuxedAccount', function() {
```
abuiles, line 187: this is the new tests

PR #328, opened 2020-04-16
Add FeeBumpTransaction class

abuiles:
## What

Add the class `FeeBumpTransaction` to represent fee bump transactions.

## Why

This pull request updates some of the code introduced in #321,  creating a different class to represent fee bump transaction.

While having a single class could work, long term, having different classes to separate both type of transactions is less error prone and makes the interface on fee bump transactions more explicit.

When you create a new `FeeBumpTransaction` you can reach the inner transaction via `.innerTransaction`, this returns an instance of `Transaction` which gives access to all the properties on the inner transaction, like the original fee, sourceAccount, operations, and memo.

### DIFF
--- a/src/fee_bump_transaction.js
+++ b/src/fee_bump_transaction.js
@@ -37,13 +37,6 @@ export class FeeBumpTransaction extends TransactionBase {
 
     super(tx, signatures, fee, networkPassphrase);
 
-    const feeSource = txEnvelope
-      .tx()
-      .feeSource()
-      .ed25519();
-
-    this._feeSource = StrKey.encodeEd25519PublicKey(feeSource);
-
     const innerTxEnvelope = xdr.TransactionEnvelope.envelopeTypeTx(
       tx.innerTx().v1()
     );
@@ -67,7 +60,7 @@ export class FeeBumpTransaction extends TransactionBase {
    * @readonly
    */
   get feeSource() {
-    return this._feeSource;
+    return this._muxedToString(this.tx.feeSource());
   }
 
   /**
--- a/src/fee_bump_transaction.js
+++ b/src/fee_bump_transaction.js
@@ -54,10 +54,18 @@ export class FeeBumpTransaction extends TransactionBase {
     );
   }
 
+  /**
+   * @type {Transaction}
+   * @readonly
+   */
   get innerTransaction() {
     return this._innerTransaction;
   }
 
+  /**
+   * @type {string}
+   * @readonly
+   */
   get feeSource() {
     return this._feeSource;
   }
--- a/src/fee_bump_transaction.js
+++ b/src/fee_bump_transaction.js
@@ -5,106 +5,84 @@ import xdr from './generated/stellar-xdr_generated';
 import { hash } from './hashing';
 
 import { StrKey } from './strkey';
-import { Operation } from './operation';
-import { Network } from './network';
-import { Memo } from './memo';
 import { Keypair } from './keypair';
+import { Transaction } from './transaction';
 
 /**
- * Use {@link TransactionBuilder} to build a transaction object, unless you have
+ * Use {@link TransactionBuilder.buildFeeBumpTransaction} to build a FeeBumpTransaction object, unless you have
  * an object or base64-encoded string of the transaction envelope XDR.
- * Once a Transaction has been created, its attributes and operations
- * should not be changed. You should only add signatures (using {@link Transaction#sign}) to a Transaction object before
+ * Once a FeeBumpTransaction has been created, its attributes and operations
+ * should not be changed. You should only add signatures (using {@link FeeBumpTransaction#sign}) to a Transaction object before
  * submitting to the network or forwarding on to additional signers.
  * @constructor
  * @param {string|xdr.TransactionEnvelope} envelope - The transaction envelope object or base64 encoded string.
- * @param {string} [networkPassphrase] passphrase of the target stellar network (e.g. "Public Global Stellar Network ; September 2015").
+ * @param {string} networkPassphrase passphrase of the target stellar network (e.g. "Public Global Stellar Network ; September 2015").
  */
-export class Transaction {
+export class FeeBumpTransaction {
   constructor(envelope, networkPassphrase) {
     if (typeof envelope === 'string') {
       const buffer = Buffer.from(envelope, 'base64');
       envelope = xdr.TransactionEnvelope.fromXDR(buffer);
     }
 
-    // Deprecation warning. TODO: remove optionality with next major release.
-    if (networkPassphrase === null || networkPassphrase === undefined) {
-      console.warn(
-        'Global `Network.current()` is deprecated. Please pass explicit argument instead, e.g. `new Transaction(envelope, Networks.PUBLIC)` (see https://git.io/fj9fG for more info).'
-      );
-    } else if (typeof networkPassphrase !== 'string') {
+    if (typeof networkPassphrase !== 'string') {
       throw new Error(
         `Invalid passphrase provided to Transaction: expected a string but got a ${typeof networkPassphrase}`
       );
     }
     this._networkPassphrase = networkPassphrase;
 
-    const txEnvelope = envelope.value();
-    let sourceAccount;
-    this._envelopeType = envelope.switch();
-    switch (this._envelopeType) {
-      case xdr.EnvelopeType.envelopeTypeTxV0():
-        sourceAccount = txEnvelope.tx().sourceAccountEd25519();
-        break;
-      case xdr.EnvelopeType.envelopeTypeTx():
-        sourceAccount = this._getSourceAccount(txEnvelope.tx().sourceAccount());
-        break;
-      default:
-        throw new Error(
-          `Invalid TransactionEnvelope: expected an envelopeTypeTxV0 or envelopeTypeTx but received an ${this._envelopeType.name}.`
-        );
-    }
-
-    // since this transaction is immutable, save the tx
-    this.tx = txEnvelope.tx();
-    this.source = StrKey.encodeEd25519PublicKey(sourceAccount);
-    this.fee = this.tx.fee().toString();
-
-    this._memo = this.tx.memo();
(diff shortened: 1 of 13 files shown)


src/fee_bump_transaction.js
```diff
@@ -0,0 +1,255 @@
...
+
+    const envelopeType = envelope.switch();
+    if (envelopeType !== xdr.EnvelopeType.envelopeTypeTxFeeBump()) {
+      throw new Error(
+        `Invalid TransactionEnvelope: expected an envelopeTypeTxFeeBump but received an ${envelopeType.name}.`
+      );
+    }
+
+    const txEnvelope = envelope.value();
+    const sourceAccount = txEnvelope
```
tamirms, line 43: I think this variable should be renamed to `feeSourceAccount`

test/unit/transaction_test.js
```diff
@@ -489,131 +485,23 @@ describe('Transaction', function() {
...
-        name: 'envelopeTypeTxInvalid',
-        value: 99
-      };
-
-      expect(() => {
-        transaction.toEnvelope();
-      }).to.throw(/Invalid EnvelopeType: envelopeTypeTxInvalid./);
-
-      done();
+      expect(txWithMuxedAccount.source).to.equal(tx.source);
```
tamirms, line 504: why are they equal? isn't `txWithMuxedAccount.source` a muxed account while `tx.source` is not?

src/fee_bump_transaction.js
```diff
@@ -0,0 +1,255 @@
...
+   * for {@link FeeBumpTransaction#getKeypairSignature} to sign the transaction.
+   * - They should send you back their `publicKey` and the `signature` string
+   * from {@link FeeBumpTransaction#getKeypairSignature}, both of which you pass to
+   * this function.
+   *
+   * @param {string} publicKey The public key of the signer
+   * @param {string} signature The base64 value of the signature XDR
+   * @returns {TransactionBuilder}
+   */
+  addSignature(publicKey = '', signature = '') {
```
tamirms, line 151: there is some code which is duplicated between  `FeeBumpTransaction` and `Transaction`. is there is an easy and clean way to share common code between the 2 implementations?

src/fee_bump_transaction.js
```diff
@@ -0,0 +1,255 @@
...
+      this.tx.innerTx().v1()
+    );
+    this._innerTransaction = new Transaction(
+      innerTxEnvelope,
+      networkPassphrase
+    );
+    this.signatures = map(txEnvelope.signatures() || [], (s) => s);
+  }
+
+  get tx() {
```
tamirms, line 63: it seems like this property isn't exposed in `Transaction` should we be exposing it here?

tamirms:
if I have a xdr envelope string and I want to parse it into a `Transaction` or `FeeBumpTransaction`, how do I do that?

if I call `new Transaction(myXDRString)` I will get an exception if `myXDRString` is a fee bump transaction envelope. Similarly, if I call `new FeeBumpTransaction(myXDRString)` I will get an exception if `myXDRString` is a normal transaction envelope. The problem is I don't know which type of envelope I have unless I decode `myXDRString`.

It seems like we want a `FromEnvelope(envelope: string | xdr.TransactionEnvelope): Transaction | FeeBumpTransaction` function. But, assuming something like `FromEnvelope()` exists there is still the problem that  most users of the library will ignore it and call `new Transaction(myXDRString)` since that was the way you created `Transaction` objects prior to v3.0.0.

Is there a way to limit access to the `Transaction` and `FeeBumpTransaction` constructors so only the builders can invoke them? Ideally, there should only be 2 ways to create a `Transaction` / `FeeBumpTransaction` instance: (1) using a builder or (2) using  `FromEnvelope`.

We should compare the tradeoffs between this approach vs just having a single `Transaction` class with an `innerTransaction()` accessor.

test/unit/transaction_test.js
abuiles: source contains the strkey representation, my understanding is that for muxed accounts we'll show the public key not the `M` version

tamirms: I thought the strkey will print muxed accounts as "M...." instead of "G..."

src/fee_bump_transaction.js
abuiles: @tamirms in the `Transaction` constructor we do `this.tx = ` which is kind of the same and allow `tx` to be modified. By doing this we disable the setter on tx. I'll update `Transaction` to follow this same patter too.

src/fee_bump_transaction.js
abuiles: sure, let me explore this a little bit.

test/unit/transaction_test.js
abuiles: It does. What I mean is that we are showing the strkey representation of the ed25519 part of the muxed account, not the whole muxed account.

abuiles:
> if I have a xdr envelope string and I want to parse it into a Transaction or FeeBumpTransaction, how do I do that?

Good point! I didn't add that yet. I think we can add it in the builder so it does the right handling depending on the envelope type. 

> Is there a way to limit access to the Transaction and FeeBumpTransaction constructors so only the builders can invoke them? Ideally, there should only be 2 ways to create a Transaction / FeeBumpTransaction instance: (1) using a builder or (2) using FromEnvelope.

I don't think there is  an out the box way to achieve this, as long as people can get a hold on the class they can create new instances. I'm actually okay by allowing people to use this directly, however, we should add a helper to the builder so people can do `fromEnvelope(envelope, network)`.

test/unit/transaction_test.js
tamirms: what's the reasoning behind that? In horizon we convert muxed accounts to ed25519 account ids in ingestion so that we only ingest "G..." addresses. but I don't see why we should do the conversion for other cases

abuiles: @tamirms oh, my understanding is that we were going to show only `G..` for now.

ire-and-curses: yeah we shouldn't show this in an SDK yet. It has no functionality and is confusing. We want this change to be transparent. If an SDK returns an M address all kinds of things downstream could potentially break.

ire-and-curses: @tamirms makes some excellent points offline which I think guide the answer to this.

>for sep10 I think we want to ensure that no mux accounts are used in the source accounts. if we use the js sdk to convert a challenge string to an object we will not be able to verify that condition correctly

>     If an SDK returns an M address all kinds of things downstream could potentially break.

this is a valid concern and that's why horizon should only ingest G addresses. address fields in horizon responses will always contain G addresses . so the only case where you can encounter M addresses is when you're parsing xdr which actually contains an M address. If the user of the sdk created the xdr string that they are parsing then they're aware that it contains an M address. If the user of the sdk is parsing an xdr string generated by someone else it's unclear to me if there is a default behavior we should support. In many cases maybe the user doesn't care that it's an M address and they would have preferred the js sdk to convert automatically. But there might be other cases similar to sep10 where there are security implications to silently converting an M address to a G address

> alternatively, we could throw an exception if you try to parse an xdr which contains an M address

I think the exception idea could be a good one. It would mean that a caller would have to handle the theoretical possibility of receiving something other than a G, which seems like a good safety move.

src/fee_bump_transaction.js
abuiles: fixed

test/unit/transaction_test.js
abuiles: I fixed this, we now return the M or G address, also it uncovered a bug! https://github.com/stellar/js-stellar-base/commit/f050af83ba65cc305692a94bde6f679924836fcf

src/transaction_base.js
```diff
@@ -0,0 +1,202 @@
+import xdr from './generated/stellar-xdr_generated';
+import { hash } from './hashing';
+import { Keypair } from './keypair';
+
+/**
+ * @ignore
+ */
+export class TransactionBase {
+  constructor(tx, signatures, fee, networkPassphrase) {
+    if (typeof networkPassphrase !== 'string') {
```
abuiles, line 10: we were suppose to make this change on the previous major bump but we forgot, not this is a good time.

src/transaction.js
```diff
@@ -1,14 +1,11 @@
...
 
 import { StrKey } from './strkey';
 import { Operation } from './operation';
-import { Network } from './network';
 import { Memo } from './memo';
-import { Keypair } from './keypair';
+import { TransactionBase } from './transaction_base';
 
 /**
  * Use {@link TransactionBuilder} to build a transaction object, unless you have
```
tamirms, line 11: I think it is worth adding a comment that we generally discourage creating `Transaction` instances manually. Instead people should be using `builder.fromXdr()` to construct `Transaction` instances. The same comment applies to `FeeBumpTransaction` as well.

If this were implemented in typescript I think we'd make `Transaction` and `FeeBumpTransaction` would be interfaces which are exported but the implementations would not be exported and only be local to the builder code

src/transaction.js
```diff
@@ -19,238 +16,121 @@ import { Keypair } from './keypair';
...
-      // make inner transaction the source account and add field feeSource
-      this.feeSource = StrKey.encodeEd25519PublicKey(sourceAccount);
-
-      this.source = StrKey.encodeEd25519PublicKey(tx.sourceAccount().ed25519());
-      this.innerSignatures = map(innerTxEnvelope.signatures() || [], (s) => s);
-      this.innerFee = tx.fee().toString();
-    }
+    super(tx, signatures, fee, networkPassphrase);
 
+    this._envelopeType = envelopeType;
```
tamirms, line 47: apparently ecmascript supports private class variables but I'm not sure how widely supported the feature is in browsers:

https://stackoverflow.com/a/52237988

abuiles: This is still a proposal. We could use babel, but at the end of the day it will transpile down to today's JS. I'll add a ticket to this library to start exploring this.

src/transaction.js
abuiles: 👍

PR #330, opened 2020-04-17
Extend StrKey to handle xdr.MuxedAccount

abuiles:
The initial implementation was not compatible with an `xdr.MuxedAccount` but rather the `muxedEd25119` discriminant. 

This PR:

- Extends `encodeMuxedAccount` to handle the raw data for `xdr.MuxedAccount`, it can return a `G..` or `M...` account.
- Extends `decodeMuxedAccount` to handle `G..` or `M..` accounts, the result can be used to create a valid `xdr.MuxedAccount`.

### DIFF
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -49,6 +49,19 @@ export class Transaction extends TransactionBase {
     this._memo = tx.memo();
     this._sequence = tx.seqNum().toString();
 
+    switch (this._envelopeType) {
+      case xdr.EnvelopeType.envelopeTypeTxV0():
+        this._source = StrKey.encodeEd25519PublicKey(
+          this.tx.sourceAccountEd25519()
+        );
+        break;
+      default:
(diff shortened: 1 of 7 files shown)


src/transaction.js
```diff
@@ -99,7 +99,7 @@ export class Transaction extends TransactionBase {
         source = StrKey.encodeEd25519PublicKey(this.tx.sourceAccountEd25519());
         break;
       default:
-        source = this._muxedToString(this.tx.sourceAccount());
+        source = StrKey.encodeMuxedAccount(this.tx.sourceAccount().toXDR());
```
tamirms, line 102: should the source string be calculated in the constructor like some of the other properties in this class?

abuiles: @tamirms  I moved it here to experiment with a different pattern, but actually let's keep it in the source. It is consistent with the other places.

PR #332, opened 2020-04-20
Update dependencies which depend on minimist.

abuiles:
Fixes a security alert.


karma-sauce.conf.js
```diff
@@ -27,12 +27,6 @@ module.exports = function(config) {
       recordScreenshots: false,
       recordVideo: false
     },
-
-    browserDisconnectTimeout: 10000,
```
abuiles, line 31: this was causing  our test to hang, I did multiple test and we don't need this for now.

PR #366, opened 2020-09-18
Update XDR definitions to Protocol 14.

abuiles:
Regenerate XDR definitions with Protocol 14.

This includes the new operations from CAP23 and CAP33.

The XDR generated in this code includes breaking changes on the internal XDR library since a bug was fixed which was causing incorrect code to be generated (see https://github.com/stellar/xdrgen/pull/52)

The following functions were renamed:

- `xdr.OperationBody.setOption()` -> `xdr.OperationBody.setOptions()`
- `xdr.OperationBody.manageDatum()` -> `xdr.OperationBody.manageData()`
- `xdr.OperationType.setOption()` -> `xdr.OperationType.setOptions()`
- `xdr.OperationType.manageDatum()` -> `xdr.OperationType.manageData()`


types/xdr.d.ts
```diff
@@ -353,9 +353,33 @@ declare namespace xdr {
       | 'manageDatum'
       | 'bumpSequence'
       | 'manageBuyOffer'
-      | 'pathPaymentStrictSend';
+      | 'pathPaymentStrictSend'
+      | 'createClaimableBalance'
+      | 'claimClaimableBalance'
+      | 'beginSponsoringFutureReserf'
```
abuiles, line 359: oh -- this is weird, it should say `beingSponsoringFutureReserves` -- I'll check what's going on

src/operation.js
```diff
@@ -235,7 +235,7 @@ export class Operation {
         result.destination = encodeMuxedAccountToAddress(attrs);
         break;
       }
-      case 'manageDatum': {
+      case 'manageData': {
```
morleyzhi, line 238: These changes seem like they'll break old code; is that change documented anywhere? Is there a way to make this backwards-compatible?

morleyzhi (approved):
No red flags for me, just one note about potentially preserving backwards compatibility. If we can't, feel free to ship it.

src/operation.js
abuiles: @morleyzhi  People who rely on stellar-base and the helper we expose, won't have any problem since the change is transparent for them.

However, if someone is using the autogenerated code for the XDR library and it's relying on this, it's very likely their code will break. I'll update the changelog to reflect this change.

PR #367, opened 2020-09-23
Add Claimant class wrapper.

abuiles:
Add a new helper class to make it easier to create `xdr.Claimant` which will be used  to create and read the `CreateClaimableBalance` operations. 

By default it will create a claimant with an unconditional predicate if none is passed:

```
const claimant = new StellarBase.Claimant(
  'GCEZWKCA5VLDNRLN3RPRJMRZOX3Z6G5CHCGSNFHEYVXM3XOJMDS674JZ'
);
```

But additionally, it offers some helpers to make it easier to build new claim predicates:

```
Claimant.predicateUnconditional();
Claimant.predicateAnd(left, right);
Claimant.predicateOr(left, right);
Claimant.predicateNot(predicate);
Claimant.predicateBeforeAbsoluteTime(unixEpoch);
Claimant.predicateBeforeRelativeTime(seconds);
```

The helpers above will appear in the documentation. The following is a simple snippet showing how to create a claimant with an `or` predicate:

```
const left = StellarBase.Claimant.predicateBeforeRelativeTime('800');
const right = StellarBase.Claimant.predicateBeforeRelativeTime(
  '1200'
);
const predicate = StellarBase.Claimant.predicateOr(left, right);
const claimant = new StellarBase.Claimant(
  'GCEZWKCA5VLDNRLN3RPRJMRZOX3Z6G5CHCGSNFHEYVXM3XOJMDS674JZ',
  predicate
);
```

### DIFF
--- a/src/claimant.js
+++ b/src/claimant.js
@@ -1,0 +1,171 @@
+import xdr from './generated/stellar-xdr_generated';
+import { Keypair } from './keypair';
+import { StrKey } from './strkey';
+
+/**
+ * Claimant class represents an xdr.Claimant
+ *
+ * The claim predicate is optional, it defaults to unconditional if none is specified.
+ *
+ * @constructor
+ * @param {string} destination - The destination account ID.
+ * @param {xdr.ClaimPredicate} [predicate] - The claim predicate.
+ */
+export class Claimant {
+  constructor(destination, predicate) {
+    if (destination && !StrKey.isValidEd25519PublicKey(destination)) {
+      throw new Error('Destination is invalid');
+    }
+    this._destination = destination;
+
+    if (!predicate) {
+      this._predicate = xdr.ClaimPredicate.claimPredicateUnconditional();
+    } else if (predicate instanceof xdr.ClaimPredicate) {
+      this._predicate = predicate;
+    } else {
+      throw new Error('Predicate should be an xdr.ClaimPredicate');
+    }
+  }
+
+  /**
+   * Returns an unconditional claim predicate
+   * @Return {xdr.ClaimPredicate}
+   */
+  static claimPredicateUnconditional() {
+    return xdr.ClaimPredicate.claimPredicateUnconditional();
+  }
+
+  /**
+   * Returns an `and` claim predicate
+   * @param {xdr.ClaimPredicate} left an xdr.ClaimPredicate
+   * @param {xdr.ClaimPredicate} right an xdr.ClaimPredicate
+   * @Return {xdr.ClaimPredicate}
+   */
+  static claimPredicateAnd(left, right) {
+    if (!(left instanceof xdr.ClaimPredicate)) {
+      throw new Error('left Predicate should be an xdr.ClaimPredicate');
+    }
+    if (!(right instanceof xdr.ClaimPredicate)) {
+      throw new Error('right Predicate should be an xdr.ClaimPredicate');
+    }
+
+    return xdr.ClaimPredicate.claimPredicateAnd([left, right]);
+  }
+
+  /**
+   * Returns an `or` claim predicate
+   * @param {xdr.ClaimPredicate} left an xdr.ClaimPredicate
+   * @param {xdr.ClaimPredicate} right an xdr.ClaimPredicate
+   * @Return {xdr.ClaimPredicate}
+   */
+  static claimPredicateOr(left, right) {
+    if (!(left instanceof xdr.ClaimPredicate)) {
+      throw new Error('left Predicate should be an xdr.ClaimPredicate');
+    }
+    if (!(right instanceof xdr.ClaimPredicate)) {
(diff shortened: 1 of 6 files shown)


src/claimant.js
```diff
@@ -0,0 +1,171 @@
...
+    } else {
+      throw new Error('Predicate should be an xdr.ClaimPredicate');
+    }
+  }
+
+  /**
+   * Returns an unconditional claim predicate
+   * @Return {xdr.ClaimPredicate}
+   */
+  static claimPredicateUnconditional() {
```
Shaptic, line 34: Pretty nitpicky, but maybe `predicateUnconditional` is enough here (and in the others), given the `Claimant` prefix? Could reduce code verbosity, e.g.
```js
Claimant.predicateUnconditional()
// vs.
Claimant.claimPredicateUnconditional()
```
especially as they get combined. No strong opinion here, just a thought.

src/claimant.js
```diff
@@ -0,0 +1,171 @@
...
+  /**
+   * Returns a `BeforeAbsoluteTime` claim predicate
+   *
+   * This predicate will be fulfilled if the closing time of the ledger that
+   * includes the CreateClaimableBalance operation is less than this (absolute) Unix timestamp.
+   *
+   * @param {string} absBefore Unix epoch as a string
+   * @Return {xdr.ClaimPredicate}
+   */
+  static claimPredicateBeforeAbsoluteTime(absBefore) {
```
Shaptic, line 94: Any reason these are strings rather than integers?

src/claimant.js
```diff
@@ -0,0 +1,171 @@
...
+  static claimPredicateUnconditional() {
+    return xdr.ClaimPredicate.claimPredicateUnconditional();
+  }
+
+  /**
+   * Returns an `and` claim predicate
+   * @param {xdr.ClaimPredicate} left an xdr.ClaimPredicate
+   * @param {xdr.ClaimPredicate} right an xdr.ClaimPredicate
+   * @Return {xdr.ClaimPredicate}
+   */
```
Shaptic, line 43: This is more of a general question since I've seen these kinds of doc-strings in a lot of other places: _why do we have them?_ They don't really provide meaningful information that isn't apparent from the function signature. It makes some sense in JS since there are no types, but I've seen it in the Go SDK as well and it feels like  clutter. Something like 
```js
/// @Return a new claim predicate that is the logical AND of both parameters.
```
could convey the same idea and look cleaner

...or maybe I'm just tired of reading `ClaimClaimableBalance/ClaimPredicate` over and over as I work on integration tests :joy:

Shaptic (approved):
Definitely wait for a JS pro to review, but LGTM :+1:

src/claimant.js
abuiles: good idea! I've updated the function names.

src/claimant.js
abuiles: @Shaptic this can be an int64 value which JS doesn't support properly. So the workaround is to take the value as a String  - and then let the application calculate this value (they can use libraries like https://www.npmjs.com/package/long)

src/claimant.js
abuiles: this is JSDOC format which is used to automatically generate the docs :) https://jsdoc.app/

morleyzhi (approved):
The code itself looks good. Two things are sticking for me. 

Obviously, naming is a judgment call, so these changes aren't critical. And the issues can be mitigated with good documentation.

- The predicate functions are on the Claimant class, but they don't necessarily need to be. Maybe it's better that they're exported from a different namespace? (I don't feel as strongly about this as the next sticking point, but the next sticking point might make us change this one.)
- Unless I'm wrong, traditionally a predicate is a function that returns true or false. That doesn't seem to be the case here. Additionally, the word "predicate" is not very clear about what it does.

My suggestion for the second point is, in developer-facing code, to refer to them in code as "conditions" or "xdrConditions" or something that makes it clear to developers that you use them to set parameters on Claimable balances. So maybe something like this:

```ts
export namespace ClaimableConditions {
  static unconditional(): xdr.ClaimPredicate;
  static and(left: xdr.ClaimPredicate, right: xdr.ClaimPredicate): xdr.ClaimPredicate;
  static or(left: xdr.ClaimPredicate, right: xdr.ClaimPredicate): xdr.ClaimPredicate;
  static not(predicate: xdr.ClaimPredicate): xdr.ClaimPredicate;
  static beforeAbsoluteTime(absBefore: string): xdr.ClaimPredicate;
  static beforeRelativeTime(seconds: string): xdr.ClaimPredicate;
}
```

abuiles:
@morleyzhi after implementing the helper where this is going to be used https://github.com/stellar/js-stellar-base/pull/368 -- I've decided to keep predicate factories in the `Claimant` --  as for the meaning of the word predicate in this context it will be clear once people start to familiarize with cap23 and cap33 (which is where this concepts are coming from)

PR #368, opened 2020-09-23
Add helpers to create CAP23 operations.

abuiles:
Extend the `Operation` class with two new helpers which allow the creation of CAP23 related operations:

- Add `Operation.createClaimableBalance`.
Extend the operation class with a new helper to create claimable balance operations.

```js
const asset = new Asset(
  'USD',
  'GDGU5OAPHNPU5UCLE5RDJHG7PXZFQYWKCFOEXSXNMR6KRQRI5T6XXCD7'
);
const amount = '100.0000000';
const claimants = [
  new Claimant(
    'GCEZWKCA5VLDNRLN3RPRJMRZOX3Z6G5CHCGSNFHEYVXM3XOJMDS674JZ',
     Claimant.predicateBeforeAbsoluteTime("4102444800000")
  )
];

const op = Operation.createClaimableBalance({
  asset,
  amount,
  claimants
});
```

- Add `Operation.claimClaimableBalance`.
Extend the operation class with a new helper to create claim claimable balance operations. It receives the `balanceId` as exposed by Horizon in the `/claimable_balances` end-point.

```js
const op = Operation.createClaimableBalance({
  balanceId: '00000000da0d57da7d4850e7fc10d2a9d0ebc731f7afb40574c03395b17d49149b91f5be',
});
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unreleased
 
 ## Add
-- Add the `Claimant` class which helps the creation of claimable balances. ((#367)[https://github.com/stellar/js-stellar-base/pull/367]).
+- Add the `Claimant` class which helps the creation of claimable balances. ([#367](https://github.com/stellar/js-stellar-base/pull/367)).
 The default behavior of this class it to create claimants with an unconditional predicate if none is passed:
 
 ```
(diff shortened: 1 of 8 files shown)


src/operations/claim_claimable_balance.js
```diff
@@ -0,0 +1,33 @@
...
+ * @returns {xdr.Operation} Claim claimable balance operation
+ *
+ * @example
+ * const op = Operation.claimClaimableBalance({
+ *   balanceId: '00000000da0d57da7d4850e7fc10d2a9d0ebc731f7afb40574c03395b17d49149b91f5be',
+ * });
+ *
+ */
+export function claimClaimableBalance(opts) {
+  if (typeof opts.balanceId !== 'string' || opts.balanceId.length < 2) {
```
Shaptic, line 19: Wait, why length < 2? 

Also, the docstring above says `opts.claimableBalanceId` which doesn't match `opts.balanceId` here.

CHANGELOG.md
```diff
@@ -37,6 +37,38 @@ const claimant = new StellarBase.Claimant(
...
+  claimants
+});
+```
+
+- Add `Operation.claimClaimableBalance` ([#368](https://github.com/stellar/js-stellar-base/pull/368))
+Extend the operation class with a new helper to create claim claimable balance operations. It receives the `balanceId` as exposed by Horizon in the `/claimable_balances` end-point.
+
+```js
+const op = Operation.createClaimableBalance({
+  balanceId: '0da0d57da7d4850e7fc10d2a9d0ebc731f7afb40574c03395b17d49149b91f5be',
```
Shaptic, line 68: Since we decided not to compress the XDR, this should be updated with the leading 0s!

src/operations/create_claimable_balance.js
```diff
@@ -0,0 +1,63 @@
...
+    throw new Error(
+      'must provide an asset for create claimable balance operation'
+    );
+  }
+
+  if (!this.isValidAmount(opts.amount)) {
+    throw new TypeError(this.constructAmountRequirementsError('amount'));
+  }
+
+  if (!opts.claimants || (opts.claimants && opts.claimants.length === 0)) {
```
Shaptic, line 46: Doesn't the latter condition fulfill the former condition? 
That is, if `opts.claimants = []`, then I think `!opts.claimants` will be true.
Maybe a type check for `Array` makes sense here, instead?

Shaptic (commented):
Again, I'm not a JS expert by any means, but everything LGTM :ok_hand:

src/operations/claim_claimable_balance.js
abuiles: good catch! Actually we don't need the length check, we can delegate that to the XDR library.

src/operations/create_claimable_balance.js
abuiles: good point,  updated to. `(!Array.isArray(opts.claimants) || opts.claimants.length === 0)`

PR #370, opened 2020-09-25
Upgrade dependencies.

abuiles:


abuiles:
we'll need to delay this a bit since there are some conflicts with dtslint and the latest version of TS.

PR #375, opened 2020-10-22
Allow account create operation with zero balance.

### LINKED ISSUE #374: Ensure that it's possible to create accounts with zero starting balance.
With the introduction of Protocol 14 and specifically [CAP-33](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0033.md), accounts can now be created **without** a starting balance when they are sponsored.

The [CAP notes](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0033.md#operation-validity-changes):
> **Operation Validity Changes**  
> Any downstream system relying on the fact that `CreateAccountOp` is invalid if `startingBalance = 0` will be affected.

This subtlety is easy to overlook, so please ensure that the SDK allows this updated constraint.

**If this SDK validates the starting balance to prevent starting balances of zero, please update the SDK to allow it.** When a transaction is submitted the Stellar network will ensure the amount is non-zero if it needs to be in the specific situations it is required, but SDKs should no longer require it for all cases.

abuiles:
Fixes #374.

### DIFF
--- a/test/unit/operation_test.js
+++ b/test/unit/operation_test.js
@@ -46,7 +46,7 @@ describe('Operation', function() {
         source: 'GCEZWKCA5VLDNRLN3RPRJMRZOX3Z6G5CHCGSNFHEYVXM3XOJMDS674JZ'
       };
       expect(() => StellarBase.Operation.createAccount(opts)).not.to.throw(
-        /startingBalance argument must be of type String, represent a positive number and have at most 7 digits after the decimal/
+        /startingBalance must be of type String, represent a non-negative number and have at most 7 digits after the decimal/
       );
     });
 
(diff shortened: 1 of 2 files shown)


test/unit/operation_test.js
```diff
@@ -39,6 +39,17 @@ describe('Operation', function() {
...
     });
 
+    it('creates a createAccount operation with startingBalance equal to 0', function() {
+      let opts = {
+        destination: 'GCEZWKCA5VLDNRLN3RPRJMRZOX3Z6G5CHCGSNFHEYVXM3XOJMDS674JZ',
+        startingBalance: '0',
+        source: 'GCEZWKCA5VLDNRLN3RPRJMRZOX3Z6G5CHCGSNFHEYVXM3XOJMDS674JZ'
+      };
+      expect(() => StellarBase.Operation.createAccount(opts)).not.to.throw(
+        /startingBalance argument must be of type String, represent a positive number and have at most 7 digits after the decimal/
```
bartekn, line 49: Shouldn't we change this error msg too (include zero).

Shaptic: We can say "non-negative" instead of "positive" to include `>= 0`.

Shaptic (approved):

test/unit/operation_test.js
abuiles: good point.

abuiles: message updated https://github.com/stellar/js-stellar-base/commit/4efa3a853c25201643deb3b75b4cc830e5b8aa2e

PR #406, opened 2021-04-02
Add type hints for CAP-35 operations.

overcat:
resolve #405

PR #407, opened 2021-04-07
Add Typescript definitions for CAP-35 operations.

Shaptic:
This builds on @overcat's work in #406 (you can see the [branch diff here](https://github.com/overcat/js-stellar-base/compare/dev-types-cap35...stellar:cap35-hints)), introducing test cases for the type hints and allows certain flags to be left off.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -391,9 +391,9 @@ export namespace OperationOptions {
     trustor: string;
     asset: Asset;
     flags: {
-      authorized?: boolean | undefined;
-      authorizedToMaintainLiabilities?: boolean | undefined;
-      clawbackEnabled?: boolean | undefined;
+      authorized?: boolean;
+      authorizedToMaintainLiabilities?: boolean;
+      clawbackEnabled?: boolean;
     };
   }
 }
@@ -668,9 +668,9 @@ export namespace Operation {
     trustor: string;
     asset: Asset;
     flags: {
-      authorized?: boolean | undefined;
-      authorizedToMaintainLiabilities?: boolean | undefined;
-      clawbackEnabled?: boolean | undefined;
+      authorized?: boolean;
+      authorizedToMaintainLiabilities?: boolean;
+      clawbackEnabled?: boolean;
     };
   }
   function setTrustLineFlags(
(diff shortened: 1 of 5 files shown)


types/index.d.ts
```diff
@@ -373,6 +379,23 @@ export namespace OperationOptions {
...
+    from: string;
+  }
+  interface ClawbackClaimableBalance extends BaseOptions {
+    balanceId: string;
+  }
+  interface SetTrustLineFlags extends BaseOptions {
+    trustor: string;
+    asset: Asset;
+    flags: {
+      authorized?: boolean | undefined;
```
overcat, line 394: If we mark them as optional, should we make corresponding changes to the [documentation](https://github.com/stellar/js-stellar-base/blob/c687cf9411/src/operations/set_trustline_flags.js#L21-L27)?

```js
 * @param {bool}   [opts.flags.authorized]  - authorize account to perform transactions
 *     with its credit
 * @param {bool}   [opts.flags.authorizedToMaintainLiabilities] - authorize
 *     account to maintain and reduce liabilities for its credit
 * @param {bool}   [opts.flags.clawbackEnabled] - stop claimable balances on this
 *     trustlines from having clawbacks enabled (this flag can only be set to
 *     false!)
```

types/test.ts
```diff
@@ -78,6 +78,34 @@ const transaction = new StellarSdk.TransactionBuilder(account, {
         preAuthTx: "da0d57da7d4850e7fc10d2a9d0ebc731f7afb40574c03395b17d49149b91f5be"
       }
     })
+  ).addOperation(
```
overcat, line 81: This looks great, I didn't notice that there are corresponding tests for the types, which caused me to miss them.

marcelosalloum (approved):
LGTM 🎉 !

types/index.d.ts
Shaptic: Good call, done in 1f91b83fcf114e339f7c894e9397a5073ed6306d.

types/test.ts
Shaptic: Yep! Likewise.

PR #408, opened 2021-04-07
Release v5.1.0

Shaptic:
I was originally going to introduce a breaking change but finagled around it; we'll revisit it once `AllowTrustOp` goes beyond its deprecation period.


marcelosalloum (approved):
LGTM!

PR #413, opened 2021-04-15
Expose the AuthClawbackEnabled flag for use with the SetOptionsOp

Shaptic:
This should allow library consumers to do things like:

```js
const op = StellarSdk.Operation.setOptions({ 
    setFlags: StellarSdk.AuthFlag.clawbackEnabled,
});
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -18,7 +18,8 @@ export {
   Operation,
   AuthRequiredFlag,
   AuthRevocableFlag,
-  AuthImmutableFlag
+  AuthImmutableFlag,
+  AuthClawbackEnabledFlag
 } from './operation';
 export * from './memo';
 export { Account } from './account';
--- a/src/operation.js
+++ b/src/operation.js
@@ -22,26 +22,40 @@ const ONE = 10000000;
 const MAX_INT64 = '9223372036854775807';
 
 /**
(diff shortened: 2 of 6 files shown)


Shaptic:
I literally do not understand why the [build is failing](https://travis-ci.com/github/stellar/js-stellar-base/jobs/498317989), when the crypto_test.js file in question doesn't even use `let` (and hasn't for some time, #396). Ignore this; the PR is still RFR.

piyalbasu:
> I literally do not understand why the [build is failing](https://travis-ci.com/github/stellar/js-stellar-base/jobs/498317989), when the crypto_test.js file in question doesn't even use `let` (and hasn't for some time, #396). Ignore this; the PR is still RFR.

Yeah, this is weird. There isn't even a line 48 in `crypto_test.js` which tells me there are some other  issues here that are out of scope of this PR

Not sure who is technically in charge of this repo, but this needs further investigation

piyalbasu (approved):

paulbellamy:
AFAICT, build is failing because we are on firefox 39, which [doesn't support let](https://caniuse.com/let). Possibly we're transpiling to a newer browser target, then trying to run tests on an old firefox. This error being post-transpilation would also explain the on-existent line numbers.

paulbellamy (approved):

PR #416, opened 2021-04-22
Add muxed account (SEP-23) support

Shaptic:
You can refer to [CAP-27](https://stellar.org/protocol/cap-27) for details on muxed accounts, [SEP-23](https://stellar.org/protocol/sep-23) for details on the stringification of them, stellar/stellar-protocol#617 for the support rollout plan, and stellar/go#3527 for the Go SDK reference implementation.

Here's the list of requirements:

- [x] Be able to convert M... addresses to & from their XDR representation easily
- [x] Helpers to create muxed accounts efficiently
- [x] Update the required operations to support muxed accounts:
  * [x] Operation.sourceAccount
  * [x] Payment.destination
  * [x] PathPaymentStrictReceive.destination
  * [x] PathPaymentStrictSend.destination
  * [x] AccountMerge.destination
  * [x] Transaction.sourceAccount
  * [x] ~~FeeBumpTransaction.feeSource~~
- [x] Hide everything behind an opt-in flag
- [x] Test cases for the above:
  * [x] Converting to/from M.../XDR
  * [x] The required operations can take/parse muxed accounts to/from the XDR
  * [x] All of this successfully hides behind an explicit opt-in flag
 - [x] Update the CHANGELOG (minor bump since everything is opt-in)

## Known Limitations
The final missing implementation detail is adapting `FeeBumpTransaction.feeSource` to accept an M-address, but this will not (and _should not_) be a part of this pull request because it involves a breaking change.

Currently, `TransactionBuilder.buildFeeBumpTransaction` takes `feeSource` typed to `Keypair`, which doesn't support M-addresses. Either Keypair will need to support M-addresses (unlikely, as this breaks abstraction barriers) or the type will need to change (likely), so this will be part of a breaking 6.0 release.

Tracking this in #422.


CHANGELOG.md
```diff
@@ -2,9 +2,39 @@
 
 ## Unreleased
 
-### Fix 
+### Add
+- **Opt-in support for muxed accounts.** This introduces `M...` addresses from [CAP-27](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0027.md), which multiplex a Stellar `G...` address across IDs to eliminate the need for ad-hoc multiplexing via the Transaction.memo field (see the relevant [SEP-29](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0029.md) and [blog post](https://www.stellar.org/developers-blog/fixing-memo-less-payments) on the topic). The following operations now support muxed accounts ([#416](https://github.com/stellar/js-stellar-base/pull/416)):
```
2opremio, line 6: I would stick to mentioning SEP23 (unless raw XDR suport for muxed accounts wasn't there already).If you disagree (fair) at the very least I would mention SEP23

2opremio: To clarify, M-addresses are covered by SEP23, CAP27 introduces the XDR representation. Those are separate specifications.

src/muxed_account.js
```diff
@@ -0,0 +1,99 @@
...
+    );
+  }
+
+  accountId() {
+    return this._gAddress;
+  }
+
+  address() {
+    return this._mAddress;
+  }
```
2opremio, line 77: These two can be confusing, due to naming and it's not clear how they behave when the underlying XDR muxed account is `Ed25519PublicKey` or `Med25519PublicKey`.

I am unfamiliar with the js SDK but (assuming `accountId()` returns the `G` strkey representation of the Muxed account regardless of the XDR MuxedAccount type), wouldn't it be better to have a conversion method to an `AccountID` object (assuming that exists) and then convert that to an strkey?

test/unit/account_test.js
```diff
@@ -1,27 +1,43 @@
...
+  it('fails to create Account object from an invalid address', function() {
+    expect(() => new StellarBase.Account('GBBB')).to.throw(
+      /accountId is invalid/
+    );
+  });
+
+  it('fails to create Account object from an invalid sequence number', function() {
+    expect(() => new StellarBase.Account(ADDRESS, 100)).to.throw(
+      /sequence must be of type string/
+    );
```
paulbellamy, line 15: What about an invalid string?

test/unit/muxed_account_test.js
```diff
@@ -0,0 +1,59 @@
...
+    );
+    expect(innerMux.id()).to.eql(StellarBase.xdr.Uint64.fromString(id));
+
+    return muxXdr;
+  };
+
+  it('can construct from a G address', function() {
+    const mux = new StellarBase.MuxedAccount(PUBKEY);
+    performBasicInvariantChecks(mux, '0', MPUBKEY_ZERO);
+  });
```
paulbellamy, line 32: `G` aren't really muxed accounts though, right? They're just a normal account?

test/unit/muxed_account_test.js
```diff
@@ -0,0 +1,59 @@
...
+  it('can construct from an M address', function() {
+    const mux = new StellarBase.MuxedAccount(MPUBKEY_ZERO);
```
paulbellamy, line 34: Should 'id' be required?

paulbellamy (commented):
Might have missed it when reviewing, but looks like there's a few missing cases for the invalid parsing testing? https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0023.md#invalid-test-cases
Or are they tested in the xdr stuff?

CHANGELOG.md
Shaptic: Good call, I completely overlooked the existence of SEP-23.

src/muxed_account.js
Shaptic: The idea is that the underlying XDR for this will *always* be a `Med25519PublicKey`. I'll add docstrings accordingly. Either way, converting to an `Account` object instead here is a great idea.

test/unit/muxed_account_test.js
Shaptic: Right, so this is to create a muxed account (w/ an ID of zero) from a G account.

test/unit/muxed_account_test.js
Shaptic: It's hard to tell what the use-cases of these APIs will be just yet, but I envisioned that if somebody decided to create a bunch of muxed accounts from a single public key, they'd want some "master" muxed version that just has an ID of zero. There's merit to making it required, though... I'm not sure.

test/unit/account_test.js
Shaptic: This should be covered by BigNumber, but I've added a test case :heavy_check_mark:

src/account.js
```diff
@@ -5,36 +5,55 @@ import { StrKey } from './strkey';
...
+ *     `GB3KJPLFUYN5VL6R3GU3EGCGVCKFDSD7BEDX42HWG5BWFKB3KQGJJRMA` or
+ *     `MA7QYNF7SOWQ3GLR2BGMZEHXAVIRZA4KVWLTJJFC7MGXUA74P7UJUAAAAAAAAAAAAGZFQ`)
  * @param {string} sequence current sequence number of the account
  */
 export class Account {
   constructor(accountId, sequence) {
-    if (!StrKey.isValidEd25519PublicKey(accountId)) {
+    if (
+      !StrKey.isValidEd25519PublicKey(accountId) &&
+      !StrKey.isValidMed25519PublicKey(accountId)
```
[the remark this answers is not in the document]
Shaptic: You're right; I modified this code prior to writing the MuxedAccount abstraction.

src/util/decode_encode_muxed_account.js
```diff
@@ -2,30 +2,118 @@ import xdr from '../generated/stellar-xdr_generated';
...
+  if (muxedAccount.switch() === xdr.CryptoKeyType.keyTypeMuxedEd25519()) {
+    if (supportMuxing) {
+      return _encodeMuxedAccountFullyToAddress(muxedAccount);
+    }
+    muxedAccount = muxedAccount.med25519();
+  }
+  return StrKey.encodeEd25519PublicKey(muxedAccount.ed25519());
+}
+
+export function encodeMuxedAccount(address, id) {
```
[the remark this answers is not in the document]
Shaptic: yep, thanks for catching

test/unit/muxed_account_test.js
Shaptic: (in other words, this abstraction is opt-in by using it)

test/unit/muxed_account_test.js
Shaptic: They can always just `new MuxedAccount(gAddress).setId('123');`

CHANGELOG.md
Shaptic: Resolved in 857faad626da638a24282313b428e68a5b43833f.

src/muxed_account.js
Shaptic: Made MuxedAccount behave like Account in 904501b731ff01520ba526c17656525961e13205.

Shaptic:
@paulbellamy I added more StrKey tests in 5c6b244658c2d62310426a2dda89c6977f239c5a but some will require refactoring the decoder itself, since apparently it wasn't written with this kind of strictness in mind. I'm delegating that to a future PR.

src/account.js
```diff
@@ -1,40 +1,61 @@
...
  */
 export class Account {
   constructor(accountId, sequence) {
+    if (StrKey.isValidMed25519PublicKey(accountId)) {
+      accountId = encodeMuxedAccountToAddress(
+        decodeAddressToMuxedAccount(accountId, true),
+        false
+      );
+    }
+
```
paulbellamy, line 36: I guess I still feel like this should fail, instead of silently throwing away the account sub-id part of the muxed account. But not super strong either way, so... /shrug

src/muxed_account.js
```diff
@@ -0,0 +1,132 @@
...
+ *
+ * @constructor
+ *
+ * @param {Account}   account - the @{link Account} instance representing the
+ *     underlying G... address
+ * @param {string}    id      - a stringified uint64 value that represents the
+ *     ID of the muxed account
+ */
+export class MuxedAccount {
+  constructor(baseAccount, id) {
```
paulbellamy, line 47: Nit: The interface here seems a bit surprising (to me). I guess the interface expected is something like:
```js
// base account setup (as before)
const gAddr = "GB3KJPLFUYN5VL6R3GU3EGCGVCKFDSD7BEDX42HWG5BWFKB3KQGJJRMA";
const seq = 5;
const base = new Account(gAddr, seq);

// Create muxed accounts with pointer back to parent to track seq
const sub1 = base.createSubaccount(1);
const sub2 = new MuxedAccount(base, 2);

// Creating a sibling of a muxed account
const sub3 = sub2.baseAccount().createSubaccount(3);

// Parsing helper
const mAddr = "MA7QYNF7SOWQ3GLR2BGMZEHXAVIRZA4KVWLTJJFC7MGXUA74P7UJUAAAAAAAAAAAALIWQ";
const parsed = MuxedAccount.fromAddress(mAddr, seq); // or something

// Fails
new Account(mAddr, seq);
new MuxedAccount(gAddr, seq);
```

src/muxed_account.js
```diff
@@ -0,0 +1,132 @@
...
+
+  /**
+   * Creates a muxed account with the *same base* address but a *new* ID.
+   *
+   * @param  {string} newId - the ID of the new muxed account
+   * @return {MuxedAccount} a new instance w/ the specified parameters
+   */
+  createSubaccount(newId) {
+    return new MuxedAccount(this.account, newId);
+  }
```
paulbellamy, line 73: Given the above comment, not sure we should have this vs being more explicit with: `mux.baseAccount().createSubAccount(4)`

paulbellamy (approved):

src/muxed_account.js
Shaptic: I like this suggestion, as it drives home the fact that it's the underlying account that matters not the muxed version.

src/account.js
Shaptic: I think you're right; throwing makes more sense here. They should use a `MuxedAccount` for M-addresses.

src/muxed_account.js
Shaptic: I think I wrote it this way because to me, the ID=0 is the "master" account (so you'd want a muxed version), but it's entirely possible that users will prefer to keep the muxed versions separate from the master account.

src/muxed_account.js
Shaptic: Good suggestion re: `baseAccount().createSubAccount(id)`, but I'm not sure how feasible the `fromAddress` API is. The base account gets disconnected, so there's no way to associate it. Yeah we could have a warning of some sort, or do a fetch via `server.loadAccount()`, but those both feel hacky to me. 

However, there is clear value in being able to create an object from the M-address. The best approach in this case may be to throw on any sequence number-related methods if `this.account` isn't set, and introduce a `setBaseAccount()` sorta thing. Then, you'd need to do something like:

```js
const parsed = MuxedAccount.fromAddress(mAddr);
expect(() => parsed.incrementSequenceNumber()).to.throw();
const baseAccount = new Account(
    parsed.underlyingGAddress(), /* or equivalent, naming is hard */
    sequence);
parsed.setBaseAccount(baseAccount);
expect(() => parsed.incrementSequenceNumber()).not.to.throw();
```

Shaptic: Ohh wait, I see what you were suggesting: if you have `fromAddress(mAddr, seq)` then you *create* an underlying `Account` instance with the values set accordingly.. I like that!

paulbellamy: Exactly. it means that underlying base account is secretly not connected to anything. But I feel like that could be covered by a warning in the MuxedAccount.fromAddress docs.

src/muxed_account.js
paulbellamy: Oh, interesting that id=0 is the master. should there be some handling for that in `createSubaccount('0')`?

Shaptic: No no sorry, that's not spec-defined or anything. I just thought that if I was mucking around w/ muxed accounts, id=0 would be my "main" receiver, but I don't think this is something that we should impose in the SDK.

src/muxed_account.js
Shaptic: Unfortunately adapting the API to these changes (which I agree are better) introduces a circular dependency: Account now needs MuxedAccount to `createSubaccount()`, while MuxedAccount needs Account to `fromAddress`. I've merged them into a single file to avoid this, but I'm open to suggestions on how else to work around that.

paulbellamy: Eh. good enough for now. Let's clean it up later.

PR #418, opened 2021-04-27
Fixes Typescript errors referencing the SetOptions auth flags.

Shaptic:
The SDK uses the full name (i.e. `AuthClawbackEnabledFlag`) rather than going through the namespace (e.g. `AuthFlag.clawbackEnabled`), and the Typescript files have been updated accordingly to hint these types.

This should fix linter errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+### Fix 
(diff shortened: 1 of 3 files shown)


quietbits (approved):
These look good to me but I'm not 100% sure how they should be. 🤞

quietbits (approved):

PR #423, opened 2021-05-06
Fix XDR gen instructions

leighmcculloch:
### What
Add an npm install and missing env var setting to the xdr generation instructions.

### Why
I needed to add them to get the process to work.


Shaptic (approved):
tysm!

Issue #422, opened 2021-05-03
Allow creation of fee-bump transactions with muxed accounts as the feeSource.


Shaptic:
See Known Limitations in #416 for some context: this will involve a breaking change to `TransactionBuilder.buildFeeBumpTransaction()`. Related also to #412.

bartekn:
Can be closed?

Issue #433, opened 2021-06-03
Rename createSubaccount to createMuxedAccount


leighmcculloch:
The Account class has a function, `createSubaccount`.

I propose we rename it, safely with deprecation, etc, to `createMuxedAccount`.

The function name `createSubaccount` makes assumptions about how muxed accounts will be used. Muxed accounts are intended to solve a variety of problems, and only one of them is the ability to represents virtual accounts. They're also intended to support invoicing and really any memo transaction case where the receiver or sender wish to track a specific transaction.

Shaptic:
This was done in an effort to better-describe the meaning / purpose of a muxed account, but you're right in that it describes a too-specific use case. I support a rename (and will keep this issue in mind for the next major version bump); I just wish we had a better one. Multiplexing is such a deeply-technical term...

Shaptic:
In #485, I propose removing `Account.createSubAccount(id)` and introducing `MuxedAccount.createSibling(id)`.

The reason is that if you'd like to create a muxed account and you have an `Account`, you should just use `new MuxedAccount()`. If you just want to create an `M...` address, you should use `encodeMuxedAccount(gAddress, id)`. The only scenario in which a "create related muxed account" helper makes sense is when you already have a `MuxedAccount` instance.

leighmcculloch:
I think we could drop `createSubAccount` and rely on this existing behavior:

```js
MuxedAccount.fromAddress(account.accountId(), id);
```

`createSibling` would introduce a new concept that is probably unnecessary given that you could just do the above instead, which is still a short single line.

@Shaptic Wdyt?

leighmcculloch:
Ah yes, I misread the docs.

Are you asking if we should keep createSubaccount? I don't feel strongly. If we keep it I think we should name it `createMuxedAccount()` because that is what it is creating. Sibling introduces a new concept to learn.

I also think it would be fine to remove `createSubaccount` and require folks to do the following, which is fine:

```js
const mux3 = new StellarBase.MuxedAccount(mux2.baseAccount(), '3'); 
```

Something that occurs to me:

Is there a way to create an M address from a G+id without a sequence number?

It is the most confusing thing to me in all of this is that MuxedAccount and Account have sequence numbers. I realize this is a common thing in all the Stellar SDKs.

Shaptic:
> If we keep it I think we should name it createMuxedAccount() because that is what it is creating. Sibling introduces a new concept to learn.

[...]

PR #434, opened 2021-06-04
Add non-breaking muxed account support to fee-bump transactions.

Shaptic:
To avoid a breaking change, I've introduced a way to set the source of a fee-bump transaction via account strings (in both `G...` and `M...` form). To enable proper muxed accounts, you pass `withMuxing: true` as the last parameter of `TransactionBuilder.buildFeeBumpTransaction()` just like everywhere else to explicitly opt-in. Namely,

```js
const feeTx = StellarBase.TransactionBuilder.buildFeeBumpTransaction(
  feeSourceMAddress,
  bumpedFee,
  innerTx,
  networkPassphrase,
  true // optionally parses the source as a true muxed address
);
```

### DIFF
--- a/src/account.js
+++ b/src/account.js
@@ -151,6 +151,9 @@ export class MuxedAccount {
 
   /**
    * A helper method to turn an M-address into its underlying G-address.
+   *
+   * @param   {string} mAddress - muxed address to convert
+   * @returns {string} underlying G-address
    */
   static parseBaseAddress(mAddress) {
     const muxedAccount = decodeAddressToMuxedAccount(mAddress, true);
--- a/src/account.js
+++ b/src/account.js
@@ -150,6 +150,14 @@ export class MuxedAccount {
   }
 
   /**
+   * A helper method to turn an M-address into its underlying G-address.
+   */
+  static parseBaseAddress(mAddress) {
+    const muxedAccount = decodeAddressToMuxedAccount(mAddress, true);
+    return encodeMuxedAccountToAddress(muxedAccount, false);
+  }
+
+  /**
    * @return {Account} the underlying account object shared among all muxed
    *     accounts with this Stellar address
    */
@@ -193,6 +201,16 @@ export class MuxedAccount {
    */
   incrementSequenceNumber() {
     return this.account.incrementSequenceNumber();
+  }
+
+  /**
+   * Creates another muxed "sub"account from the base with a new ID set
+   *
+   * @param  {string} id - the ID of the new muxed account
+   * @return {MuxedAccount} a new instance w/ the specified parameters
+   */
+  createSubaccount(id) {
+    return new MuxedAccount(this.account, id);
   }
 
   /**
--- a/src/account.js
+++ b/src/account.js
@@ -150,17 +150,6 @@ export class MuxedAccount {
   }
 
   /**
-   * A helper method to turn an M-address into its underlying G-address.
-   *
-   * @param   {string} mAddress - muxed address to convert
-   * @returns {string} underlying G-address
-   */
-  static parseBaseAddress(mAddress) {
-    const muxedAccount = decodeAddressToMuxedAccount(mAddress, true);
-    return encodeMuxedAccountToAddress(muxedAccount, false);
-  }
-
-  /**
    * @return {Account} the underlying account object shared among all muxed
    *     accounts with this Stellar address
    */
--- a/src/transaction_builder.js
+++ b/src/transaction_builder.js
@@ -2,6 +2,7 @@ import { UnsignedHyper } from 'js-xdr';
 import BigNumber from 'bignumber.js';
 import clone from 'lodash/clone';
 import isUndefined from 'lodash/isUndefined';
+import isString from 'lodash/isString';
 
 import xdr from './generated/stellar-xdr_generated';
 import { Transaction } from './transaction';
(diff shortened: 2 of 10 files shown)


src/account.js
```diff
@@ -195,6 +206,16 @@ export class MuxedAccount {
     return this.account.incrementSequenceNumber();
   }
 
+  /**
+   * Creates another muxed "sub"account from the base with a new ID set
+   *
+   * @param  {string} id - the ID of the new muxed account
+   * @return {MuxedAccount} a new instance w/ the specified parameters
+   */
+  createSubaccount(id) {
```
leighmcculloch, line 215: If we need this method can we rename this to `createdMuxedAccount`? See https://github.com/stellar/js-stellar-base/issues/433 for why.

However, I think this function confusing. You can't create a muxed account that multiplexes another muxed account. The following code, although verbose, seems much clearer. Is there a reason we can't use it? or a reason we need to be able to create muxed accounts from muxed accounts?

```js
muxedAccount.baseAccount().createMuxedAccount(id)
```

src/transaction_builder.js
```diff
@@ -272,18 +272,36 @@ export class TransactionBuilder {
...
+   *
    * @returns {FeeBumpTransaction}
    */
   static buildFeeBumpTransaction(
     feeSource,
     baseFee,
     innerTx,
-    networkPassphrase
+    networkPassphrase,
+    muxedId
```
leighmcculloch, line 304: Is it possible for us to make feeSource a union of `Keypair` or `string`, where `string` is an G/M address?

That should be backwards compatible.

That would be more consistent with other places in the SDK where we can provide a source account in string form too.

leighmcculloch (commented):
A couple questions and thoughts shared inline.

src/account.js
Shaptic: Though I completely agree that it's odd to have this method on this object, I added it to `MuxedAccount` so that there was "interface" compatibility with `Account`. And :+1: for the rename, though unfortunate that it wasn't noted at its introduction.

src/transaction_builder.js
Shaptic: Unfortunately, no: it has to be a `Keypair` because it signs the transaction.

src/account.js
leighmcculloch: What is driving the need for this interface compatibility? I don't see anything new calling `createSubaccount` in this PR.

src/transaction_builder.js
leighmcculloch: Woah. Building a fee bump signs it? That's not good. 

When I look through the `buildFeeBumpTransaction` I don't see the keypair being used to sign. The signatures are empty here:
https://github.com/stellar/js-stellar-base/blob/d92bb4a04d89b4def0b9e1a93204ab8558579913/src/transaction_builder.js#L358

Shaptic: Uhm wow, yeah, you're right. I could've sworn.. In that case, I definitely support this idea! Will give it a shot now.

src/account.js
Shaptic: There's no true need (especially since nobody is using muxed accounts); I just had it in mind after js-stellar-sdk#653.

leighmcculloch: Ah, interesting. In this case, I don't think all accounts should have this capability.

Shaptic: I agree in theory, but then you cannot use `MuxedAccount` abstraction where `Account` is allowed, yet the two should be largely interchangeable. Fixing this imo fundamentally requires a bigger change (see [this comment](https://github.com/stellar/js-stellar-sdk/pull/655#discussion_r643486317), where `createSubaccount` exists only on the non-interface), but that's out of scope here, so for now I think making muxed accounts "quack like an Account" is a reasonable measure.

src/transaction_builder.js
Shaptic: Done in 0dd5b8d; let me know what you think!

src/account.js
leighmcculloch: Got it. Thanks for explaining.

For my own understanding, which is limited here, is there no way in JavaScript/TypeScript to specify an interface type that provides a set of functions, so AccountResponse and MuxedAccount don't always have to have the exact same function set as Account, but rather there could be some interface Account that is a common subset that doesn't include `createSubaccount`?

src/account.js
```diff
@@ -149,6 +149,17 @@ export class MuxedAccount {
     return new MuxedAccount(new Account(gAddress, sequenceNum), id);
   }
 
+  /**
+   * A helper method to turn an M-address into its underlying G-address.
+   *
+   * @param   {string} mAddress - muxed address to convert
+   * @returns {string} underlying G-address
+   */
+  static parseBaseAddress(mAddress) {
```
leighmcculloch, line 158: The name of this function is a little confusing. Often functions named `parseX` are understood to "parse X". In this case the function is named `parseBaseAddress` but what it does is "parse an M address".

I think it would be easier to understand the intent of this function if it was named `parseMuxedAddress` and it returned all the components of a muxed address. I thought we would have had something like this already, but I guess we don't?

leighmcculloch (approved):
On a final pass I noticed a function below that has a confusing name. I defer to you on it though. Otherwise the PR looks good.

src/account.js
Shaptic: You're essentially describing exactly what I want: an interface that covers the common subset of functionality across `AccountResponse`, `MuxedAccount`, and `Account`. This would involve replacing everywhere that says `Account` (an object) which _used_ to act as said "common subset" with some new interface.

src/account.js
Shaptic: I see what you're saying. We do have that: `MuxedAccount.fromAddress`, but doing M -> G strings becomes unwieldy:

```js
const muxed = new StellarBase.MuxedAccount.fromAddress(mAddress, '0');
const gAddress = muxed.baseAccount().accountId();
```

(see cfc6df0). How about `extract` over `parse`, @leighmcculloch? e.g. 

```js
const gAddress = MuxedAccount.extractBaseAddress(mAddress);
```

Shaptic:
Closes #422

PR #454, opened 2021-08-18
[v6.0.0] Update XDR to include CAP-38 definitions 

marcelosalloum:
### Add

- Introduced new CAP-38 operations `LiquidityPoolDepositOp` and `LiquidityPoolWithdrawOp`.
- Introduced two new types of assets, `TrustLineAsset` and `ChangeTrustAsset`.

### Update

- The XDR definitions have been updated to support CAP-38.
- Extended `Operation` class with the `Operation.revokeLiquidityPoolSponsorship` helper that allows revoking a liquidity pool sponsorship.
- Asset types now include `AssetType.liquidityPoolShares`.

### Breaking

- Must use `TrustLineAsset` instead of `Asset` in the function `Operation.revokeTrustlineSponsorship` and the XDR objects `TrustLineEntry` and `LedgerKeyTrustLine`.
- Must use `ChangeTrustAsset` instead of `Asset` in the function `Operation.changeTrust` and the XDR object `ChangeTrustOp`.
- The `AssetAlphaNum4` and `AssetAlphaNum12` objects were renamed to `AlphaNum4` and `AlphaNum12` respectively.

### DIFF
--- a/src/asset.js
+++ b/src/asset.js
@@ -79,10 +79,10 @@ export class Asset {
     let xdrType;
     let xdrTypeString;
     if (this.code.length <= 4) {
-      xdrType = xdr.AssetAlphaNum4;
+      xdrType = xdr.AlphaNum4;
       xdrTypeString = 'assetTypeCreditAlphanum4';
     } else {
-      xdrType = xdr.AssetAlphaNum12;
+      xdrType = xdr.AlphaNum12;
       xdrTypeString = 'assetTypeCreditAlphanum12';
     }
 
--- a/src/change_trust_asset.js
+++ b/src/change_trust_asset.js
@@ -1,0 +1,253 @@
+import clone from 'lodash/clone';
+import padEnd from 'lodash/padEnd';
+import trimEnd from 'lodash/trimEnd';
+import { Asset } from './asset';
+import xdr from './generated/stellar-xdr_generated';
+import { Keypair } from './keypair';
+import { LiquidityPoolFeeV18, liquidityPoolId } from './liquidity_pool_id';
+import { StrKey } from './strkey';
+
+// TODO: ChangeTrustAsset
+// - [ ] ChangeTrustOp -> Asset to ChangeTrustAsset
+
+/**
+ * ChangeTrustAsset class represents a trustline change to either a liquidity
+ * pool or an issued asset with an asset code / issuer account ID pair.
+ *
+ * The change trust asset can either represent a trustline change to the an
+ * issued asset or to a liquidity pool. For an issued asset, the code and issuer
+ * will be valid and liquidityPool parameters will be empty. For liquidity
+ * pools, the liquidityPool parameters will all be valid while the code and
+ * issuer will be empty.
+ *
+ * @constructor
+ * @param {string} code - The asset code.
+ * @param {string} issuer - The account ID of the asset issuer.
+ * @param {LiquidityPoolParams.ConstantProduct} liquidityPoolParams – the
+ * liquidity pool parameters.
+ * @param {Asset} liquidityPoolParams.asseta – the first asset in the Pool, it
+ * must respect the rule assetA < assetB.
+ * @param {Asset} liquidityPoolParams.assetB – the second asset in the Pool, it
+ * must respect the rule assetA < assetB.
+ * @param {number} liquidityPoolParams.fee – the liquidity pool fee. For now the
+ * only fee supported is `30`.
+ */
+export class ChangeTrustAsset {
+  constructor(code, issuer, liquidityPoolParams) {
+    if (!code && !issuer && !liquidityPoolParams) {
+      throw new Error(
+        'Must provide either code, issuer or liquidityPoolParams'
+      );
+    }
+
+    // Validate code & issuer.
+    if ((code || issuer) && !/^[a-zA-Z0-9]{1,12}$/.test(code)) {
+      throw new Error(
+        'Asset code is invalid (maximum alphanumeric, 12 characters at max)'
+      );
+    }
+    if (code && String(code).toLowerCase() !== 'xlm' && !issuer) {
+      throw new Error('Issuer cannot be null');
+    }
+    if (issuer && !StrKey.isValidEd25519PublicKey(issuer)) {
+      throw new Error('Issuer is invalid');
+    }
+
+    // Validate liquidity pool params.
+    if (!code && !issuer) {
+      if (!liquidityPoolParams) {
+        throw new Error('Must provide liquidityPoolParams or code & issuer');
+      }
+      const { asseta, assetB, fee } = liquidityPoolParams;
+      if (!asseta || !(asseta instanceof Asset)) {
+        throw new Error('asseta is invalid');
+      }
+      if (!assetB || !(assetB instanceof Asset)) {
+        throw new Error('assetB is invalid');
+      }
+      if (!fee || fee !== LiquidityPoolFeeV18) {
+        throw new Error('fee is invalid');
+      }
+    }
+
+    this.code = code;
+    this.issuer = issuer;
+    this.liquidityPoolParams = liquidityPoolParams;
+  }
+
+  /**
+   * Returns a trustline asset object for the native asset.
+   * @returns {ChangeTrustAsset}
+   */
+  static native() {
+    return new ChangeTrustAsset('XLM');
+  }
+
+  /**
+   * Returns a change trust asset object from its XDR object representation.
+   * @param {xdr.ChangeTrustAsset} ctAssetXdr - The asset xdr object.
+   * @returns {ChangeTrustAsset}
+   */
+  static fromOperation(ctAssetXdr) {
+    let anum;
+    let code;
+    let issuer;
+    let liquidityPoolParams;
+    switch (ctAssetXdr.switch()) {
+      case xdr.AssetType.assetTypeNative():
+        return this.native();
+      case xdr.AssetType.assetTypeCreditAlphanum4():
+        anum = ctAssetXdr.alphaNum4();
+      /* falls through */
+      case xdr.AssetType.assetTypeCreditAlphanum12():
+        anum = anum || ctAssetXdr.alphaNum12();
+        issuer = StrKey.encodeEd25519PublicKey(anum.issuer().ed25519());
+        code = trimEnd(anum.assetCode(), '\0');
+        return new this(code, issuer);
+      case xdr.AssetType.assetTypePoolShare():
+        // TODO: review if this is correct
+        liquidityPoolParams = ctAssetXdr.liquidityPool().constantProduct();
+        return new this(null, null, {
+          asseta: liquidityPoolParams.asseta(),
+          assetB: liquidityPoolParams.assetB(),
+          fee: liquidityPoolParams.fee()
+        });
+      default:
+        throw new Error(`Invalid asset type: ${ctAssetXdr.switch().name}`);
+    }
+  }
+
+  /**
+   * Returns the xdr object for this asset.
+   * @returns {xdr.ChangeTrustAsset} XDR ChangeTrustAsset object
+   */
+  toXDRObject() {
+    if (this.isNative()) {
+      return xdr.ChangeTrustAsset.assetTypeNative();
+    }
+
+    if (this.isLiquidityPool()) {
+      // TODO: review if this is correct
+      const lpConstantProductParamsXdr = new xdr.LiquidityPoolConstantProductParameters(
+        this.getLiquidityPoolParams()
+      );
+      const lpParamsXdr = new xdr.LiquidityPoolParameters(
+        'liquidityPoolConstantProduct',
+        lpConstantProductParamsXdr
+      );
+      return new xdr.ChangeTrustAsset('assetTypePoolShare', lpParamsXdr);
+    }
+
+    let xdrType;
+    let xdrTypeString;
+    if (this.code.length <= 4) {
+      xdrType = xdr.AlphaNum4;
+      xdrTypeString = 'assetTypeCreditAlphanum4';
+    } else {
+      xdrType = xdr.AlphaNum12;
+      xdrTypeString = 'assetTypeCreditAlphanum12';
+    }
+
+    // pad code with null bytes if necessary
+    const padLength = this.code.length <= 4 ? 4 : 12;
+    const paddedCode = padEnd(this.code, padLength, '\0');
+
+    // eslint-disable-next-line new-cap
+    const assetType = new xdrType({
+      assetCode: paddedCode,
+      issuer: Keypair.fromPublicKey(this.issuer).xdrAccountId()
+    });
+
+    return new xdr.ChangeTrustAsset(xdrTypeString, assetType);
+  }
+
+  /**
+   * @returns {string} Asset code
+   */
+  getCode() {
+    return clone(this.code);
+  }
+
+  /**
+   * @returns {string} Asset issuer
+   */
+  getIssuer() {
+    return clone(this.issuer);
+  }
+
+  /**
+   * @returns {string} Liquidity pool ID
+   */
+  getLiquidityPoolParams() {
+    return clone(this.liquidityPoolParams);
+  }
+
+  /**
+   * @see [Assets concept](https://www.stellar.org/developers/guides/concepts/assets.html)
+   * @returns {string} Asset type. Can be one of following types:
+   *
+   * * `native`
+   * * `credit_alphanum4`
+   * * `credit_alphanum12`
+   * * `liquidity_pool_shares`
+   */
+  getAssetType() {
+    if (this.isNative()) {
+      return 'native';
+    }
+    if (this.isLiquidityPool()) {
+      return 'liquidity_pool_shares';
+    }
+    if (this.code.length >= 1 && this.code.length <= 4) {
+      return 'credit_alphanum4';
+    }
+    if (this.code.length >= 5 && this.code.length <= 12) {
+      return 'credit_alphanum12';
+    }
+
+    return null;
+  }
+
+  /**
+   * @returns {boolean}  true if this trustline asset object is the native asset.
+   */
+  isNative() {
+    return this.code && !this.issuer;
+  }
+
+  /**
+   * @returns {boolean} true if this trustline asset object is a liquidity pool.
+   */
+  isLiquidityPool() {
+    return !!this.liquidityPoolParams;
+  }
+
+  /**
+   * @param {ChangeTrustAsset} asset Asset to compare
+   * @returns {boolean} true if this asset equals the given asset.
+   */
+  equals(asset) {
+    return (
+      this.code === asset.getCode() &&
+      this.issuer === asset.getIssuer() &&
+      this.liquidityPoolParams === asset.getLiquidityPoolParams()
+    );
+  }
+
+  toString() {
+    if (this.isNative()) {
+      return 'native';
+    }
+
+    if (this.isLiquidityPool()) {
+      const poolId = liquidityPoolId(
+        xdr.LiquidityPoolType.liquidityPoolConstantProduct(),
+        this.getLiquidityPoolParams()
+      ).toString('hex');
+      // TODO: review if this is correct
+      return `liquidity_pool:${poolId}`;
+    }
+
+    return `${this.getCode()}:${this.getIssuer()}`;
+  }
+}
--- a/src/change_trust_asset.js
+++ b/src/change_trust_asset.js
@@ -108,8 +108,8 @@ export class ChangeTrustAsset {
         // TODO: review if this is correct
         liquidityPoolParams = ctAssetXdr.liquidityPool().constantProduct();
         return new this(null, null, {
-          asseta: liquidityPoolParams.asseta(),
-          assetB: liquidityPoolParams.assetB(),
+          asseta: Asset.fromOperation(liquidityPoolParams.asseta()),
+          assetB: Asset.fromOperation(liquidityPoolParams.assetB()),
           fee: liquidityPoolParams.fee()
         });
       default:
@@ -128,8 +128,13 @@ export class ChangeTrustAsset {
 
     if (this.isLiquidityPool()) {
       // TODO: review if this is correct
+      const { asseta, assetB, fee } = this.getLiquidityPoolParams();
       const lpConstantProductParamsXdr = new xdr.LiquidityPoolConstantProductParameters(
-        this.getLiquidityPoolParams()
+        {
+          asseta: asseta.toXDRObject(),
+          assetB: assetB.toXDRObject(),
+          fee
+        }
       );
       const lpParamsXdr = new xdr.LiquidityPoolParameters(
         'liquidityPoolConstantProduct',
--- a/src/change_trust_asset.js
+++ b/src/change_trust_asset.js
@@ -1,39 +1,36 @@
 import clone from 'lodash/clone';
 import padEnd from 'lodash/padEnd';
 import trimEnd from 'lodash/trimEnd';
-import { Asset } from './asset';
 import xdr from './generated/stellar-xdr_generated';
+import { Asset } from './asset';
 import { Keypair } from './keypair';
 import { LiquidityPoolFeeV18, getLiquidityPoolId } from './liquidity_pool_id';
 import { StrKey } from './strkey';
 
 /**
  * ChangeTrustAsset class represents a trustline change to either a liquidity
- * pool or an issued asset with an asset code / issuer account ID pair.
+ * pool, the native asset (`XLM`) or an issued asset with an asset code / issuer
+ * account ID pair.
  *
- * The change trust asset can either represent a trustline change to the an
- * issued asset or to a liquidity pool. For an issued asset, the code and issuer
- * will be valid and liquidityPool parameters will be empty. For liquidity
- * pools, the liquidityPool parameters will all be valid while the code and
- * issuer will be empty.
+ * In case of the native asset, the code will represent `XLM` while the issuer
+ * and liquidityPoolParameters will be empty. For an issued asset, the code and
+ * issuer will be valid and liquidityPoolParameters will be empty. For liquidity
+ * pools, the liquidityPoolParameters will be valid and the remaining two fields
+ * will be empty.
  *
  * @constructor
  * @param {string} code - The asset code.
  * @param {string} issuer - The account ID of the asset issuer.
- * @param {LiquidityPoolParams.ConstantProduct} liquidityPoolParams – the
- * liquidity pool parameters.
- * @param {Asset} liquidityPoolParams.asseta – the first asset in the Pool, it
- * must respect the rule assetA < assetB.
- * @param {Asset} liquidityPoolParams.assetB – the second asset in the Pool, it
- * must respect the rule assetA < assetB.
- * @param {number} liquidityPoolParams.fee – the liquidity pool fee. For now the
- * only fee supported is `30`.
+ * @param {LiquidityPoolParameters} liquidityPoolParameters – The liquidity pool parameters.
+ * @param {Asset} liquidityPoolParameters.asseta – The first asset in the Pool, it must respect the rule assetA < assetB.
+ * @param {Asset} liquidityPoolParameters.assetB – The second asset in the Pool, it must respect the rule assetA < assetB.
+ * @param {number} liquidityPoolParameters.fee – The liquidity pool fee. For now the only fee supported is `30`.
  */
 export class ChangeTrustAsset {
-  constructor(code, issuer, liquidityPoolParams) {
-    if (!code && !issuer && !liquidityPoolParams) {
+  constructor(code, issuer, liquidityPoolParameters) {
+    if (!code && !issuer && !liquidityPoolParameters) {
       throw new Error(
-        'Must provide either code, issuer or liquidityPoolParams'
+        'Must provide either code, issuer or liquidityPoolParameters'
       );
     }
 
@@ -50,30 +47,29 @@ export class ChangeTrustAsset {
       throw new Error('Issuer is invalid');
     }
 
-    // Validate liquidity pool params.
-    if (!code && !issuer) {
-      if (!liquidityPoolParams) {
-        throw new Error('Must provide liquidityPoolParams or code & issuer');
-      }
-      const { asseta, assetB, fee } = liquidityPoolParams;
-      if (!asseta || !(asseta instanceof Asset)) {
-        throw new Error('asseta is invalid');
-      }
-      if (!assetB || !(assetB instanceof Asset)) {
-        throw new Error('assetB is invalid');
-      }
-      if (!fee || fee !== LiquidityPoolFeeV18) {
-        throw new Error('fee is invalid');
-      }
-    }
-
     this.code = code;
     this.issuer = issuer;
-    this.liquidityPoolParams = liquidityPoolParams;
+    if (code) {
+      // return early if it's an asset with code.
+      return;
+    }
+
+    // Validate liquidity pool params.
+    const { asseta, assetB, fee } = liquidityPoolParameters;
+    if (!asseta || !(asseta instanceof Asset)) {
+      throw new Error('asseta is invalid');
+    }
+    if (!assetB || !(assetB instanceof Asset)) {
+      throw new Error('assetB is invalid');
+    }
+    if (!fee || fee !== LiquidityPoolFeeV18) {
+      throw new Error('fee is invalid');
+    }
+    this.liquidityPoolParameters = liquidityPoolParameters;
   }
 
   /**
-   * Returns a trustline asset object for the native asset.
+   * Returns a change trust asset object for the native asset.
    * @returns {ChangeTrustAsset}
    */
   static native() {
@@ -82,14 +78,14 @@ export class ChangeTrustAsset {
 
   /**
    * Returns a change trust asset object from its XDR object representation.
-   * @param {xdr.ChangeTrustAsset} ctAssetXdr - The asset xdr object.
+   * @param {xdr.ChangeTrustAsset} ctAssetXdr - The asset XDR object.
    * @returns {ChangeTrustAsset}
    */
   static fromOperation(ctAssetXdr) {
     let anum;
     let code;
     let issuer;
-    let liquidityPoolParams;
+    let liquidityPoolParameters;
     switch (ctAssetXdr.switch()) {
       case xdr.AssetType.assetTypeNative():
         return this.native();
@@ -103,11 +99,11 @@ export class ChangeTrustAsset {
         return new this(code, issuer);
       case xdr.AssetType.assetTypePoolShare():
         // TODO: review if this is correct
-        liquidityPoolParams = ctAssetXdr.liquidityPool().constantProduct();
+        liquidityPoolParameters = ctAssetXdr.liquidityPool().constantProduct();
         return new this(null, null, {
-          asseta: Asset.fromOperation(liquidityPoolParams.asseta()),
-          assetB: Asset.fromOperation(liquidityPoolParams.assetB()),
-          fee: liquidityPoolParams.fee()
+          asseta: Asset.fromOperation(liquidityPoolParameters.asseta()),
+          assetB: Asset.fromOperation(liquidityPoolParameters.assetB()),
+          fee: liquidityPoolParameters.fee()
         });
       default:
         throw new Error(`Invalid asset type: ${ctAssetXdr.switch().name}`);
@@ -115,8 +111,8 @@ export class ChangeTrustAsset {
   }
 
   /**
-   * Returns the xdr object for this asset.
-   * @returns {xdr.ChangeTrustAsset} XDR ChangeTrustAsset object
+   * Returns the XDR object for this change trust asset.
+   * @returns {xdr.ChangeTrustAsset} XDR ChangeTrustAsset object.
    */
   toXDRObject() {
     if (this.isNative()) {
@@ -125,7 +121,7 @@ export class ChangeTrustAsset {
 
     if (this.isLiquidityPool()) {
       // TODO: review if this is correct
-      const { asseta, assetB, fee } = this.getLiquidityPoolParams();
+      const { asseta, assetB, fee } = this.getLiquidityPoolParameters();
       const lpConstantProductParamsXdr = new xdr.LiquidityPoolConstantProductParameters(
         {
           asseta: asseta.toXDRObject(),
@@ -160,29 +156,29 @@ export class ChangeTrustAsset {
   }
 
   /**
-   * @returns {string} Asset code
+   * @returns {string} Asset code.
    */
   getCode() {
     return clone(this.code);
   }
 
   /**
-   * @returns {string} Asset issuer
+   * @returns {string} Asset issuer.
    */
   getIssuer() {
     return clone(this.issuer);
   }
 
   /**
-   * @returns {string} Liquidity pool ID
+   * @returns {LiquidityPoolParameters} Liquidity pool parameters.
    */
-  getLiquidityPoolParams() {
-    return clone(this.liquidityPoolParams);
+  getLiquidityPoolParameters() {
+    return clone(this.liquidityPoolParameters);
   }
 
   /**
    * @see [Assets concept](https://www.stellar.org/developers/guides/concepts/assets.html)
-   * @returns {string} Asset type. Can be one of following types:
+   * @returns {string} Asset type. Can be one of the following:
    *
    * * `native`
    * * `credit_alphanum4`
@@ -207,28 +203,28 @@ export class ChangeTrustAsset {
   }
 
   /**
-   * @returns {boolean}  true if this trustline asset object is the native asset.
+   * @returns {boolean} `true` if this change trust asset object is the native asset.
    */
   isNative() {
     return this.code && !this.issuer;
   }
 
   /**
-   * @returns {boolean} true if this trustline asset object is a liquidity pool.
+   * @returns {boolean} `true` if this change trust asset object is a liquidity pool.
    */
   isLiquidityPool() {
-    return !!this.liquidityPoolParams;
+    return !!this.liquidityPoolParameters;
   }
 
   /**
-   * @param {ChangeTrustAsset} asset Asset to compare
-   * @returns {boolean} true if this asset equals the given asset.
+   * @param {ChangeTrustAsset} asset the ChangeTrustAsset to compare
+   * @returns {boolean} `true` if this asset equals the given asset.
    */
   equals(asset) {
     return (
       this.code === asset.getCode() &&
       this.issuer === asset.getIssuer() &&
-      this.liquidityPoolParams === asset.getLiquidityPoolParams()
+      this.liquidityPoolParameters === asset.getLiquidityPoolParameters()
     );
   }
 
@@ -240,9 +236,8 @@ export class ChangeTrustAsset {
     if (this.isLiquidityPool()) {
       const poolId = getLiquidityPoolId(
         'constant_product',
-        this.getLiquidityPoolParams()
+        this.getLiquidityPoolParameters()
       ).toString('hex');
-      // TODO: review if this is correct
       return `liquidity_pool:${poolId}`;
     }
 
--- a/src/change_trust_asset.js
+++ b/src/change_trust_asset.js
@@ -4,11 +4,8 @@ import trimEnd from 'lodash/trimEnd';
 import { Asset } from './asset';
 import xdr from './generated/stellar-xdr_generated';
 import { Keypair } from './keypair';
-import { LiquidityPoolFeeV18, liquidityPoolId } from './liquidity_pool_id';
+import { LiquidityPoolFeeV18, getLiquidityPoolId } from './liquidity_pool_id';
 import { StrKey } from './strkey';
-
-// TODO: ChangeTrustAsset
-// - [ ] ChangeTrustOp -> Asset to ChangeTrustAsset
 
 /**
  * ChangeTrustAsset class represents a trustline change to either a liquidity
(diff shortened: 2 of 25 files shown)


types/xdr.d.ts
```diff
@@ -1478,6 +1612,9 @@ export namespace xdr {
 
   const DataValue: VarOpaque;
 
+  // The generated code was `type PoolId = Hash;` but that's not a valid type.
+  type PoolId = typeof Hash;
```
marcelosalloum, line 1616: The generated was actually `type PoolId = Hash;` but that's not a valid type.

types/xdr.d.ts
```diff
@@ -2056,6 +2227,124 @@ export namespace xdr {
     static validateXDR(input: string, format: 'hex' | 'base64'): boolean;
   }
 
+  class LiquidityPoolConstantProductParameters {
+    constructor(attributes: { asseta: Asset; assetB: Asset; fee: number });
+
+    asseta(value?: Asset): Asset;
+
+    assetB(value?: Asset): Asset;
```
marcelosalloum, line 2235: All parameters with the pattern `*assetA` were generated with the camelCase wrong:
* asseta should be assetA
* maxAmounta should be maxAmountA
* minAmounta should be minAmountA

marcelosalloum (commented):
❗  There were a few places where the generated code wasn't generated correctly. I've highlighted them and I'll need help figuring out how to fix them.

src/change_trust_asset.js
```diff
@@ -0,0 +1,246 @@
...
+        code = trimEnd(anum.assetCode(), '\0');
+        return new this(code, issuer);
+      case xdr.AssetType.assetTypePoolShare():
+        // TODO: review if this is correct
+        liquidityPoolParameters = ctAssetXdr.liquidityPool().constantProduct();
+        return new this(null, null, {
+          asseta: Asset.fromOperation(liquidityPoolParameters.asseta()),
+          assetB: Asset.fromOperation(liquidityPoolParameters.assetB()),
+          fee: liquidityPoolParameters.fee()
+        });
```
marcelosalloum, line 107: 👀

src/change_trust_asset.js
```diff
@@ -0,0 +1,246 @@
...
+          assetB: assetB.toXDRObject(),
+          fee
+        }
+      );
+      const lpParamsXdr = new xdr.LiquidityPoolParameters(
+        'liquidityPoolConstantProduct',
+        lpConstantProductParamsXdr
+      );
+      return new xdr.ChangeTrustAsset('assetTypePoolShare', lpParamsXdr);
+    }
```
marcelosalloum, line 137: 👀

src/trustline_asset.js
```diff
@@ -0,0 +1,206 @@
...
+      /* falls through */
+      case xdr.AssetType.assetTypeCreditAlphanum12():
+        anum = anum || tlAssetXdr.alphaNum12();
+        issuer = StrKey.encodeEd25519PublicKey(anum.issuer().ed25519());
+        code = trimEnd(anum.assetCode(), '\0');
+        return new this(code, issuer);
+      case xdr.AssetType.assetTypePoolShare():
+        // TODO: review if this is correct
+        liquidityPoolId = tlAssetXdr.liquidityPoolId().toString('hex');
+        return new this(null, null, liquidityPoolId);
```
marcelosalloum, line 82: 👀

src/trustline_asset.js
```diff
@@ -0,0 +1,206 @@
...
+  toXDRObject() {
+    if (this.isNative()) {
+      return xdr.TrustLineAsset.assetTypeNative();
+    }
+
+    if (this.isLiquidityPool()) {
+      // TODO: review if this is correct
+      const xdrPoolId = xdr.PoolId.fromXDR(this.liquidityPoolId, 'hex');
+      return new xdr.TrustLineAsset('assetTypePoolShare', xdrPoolId);
+    }
```
marcelosalloum, line 101: 👀

types/test.ts
```diff
@@ -269,3 +288,19 @@ const trust = StellarSdk.xdr.SetTrustLineFlagsOp.fromXDR(
...
+);
+lpDeposit; // $ExpectType LiquidityPoolDepositOp
+
+// TODO: check if this is correct
+const lpWithdraw = StellarSdk.xdr.LiquidityPoolWithdrawOp.fromXDR(
+  // tslint:disable:max-line-length
+  '3XsauDHCczEN2+xvl4cKqDwvvXjOIq3tN+y/TzOA+scAAAAAAvrwgAAAAAAF9eEAAAAAAAvrwgA=',
+  'base64'
+);
+lpWithdraw; // $ExpectType LiquidityPoolWithdrawOp
```
marcelosalloum, line 306: 👀  I'm not sure these tests are being executed at all.

marcelosalloum (commented):
👀  I've highlighted a few places where I convert between `object <> xdrObject`. I'd appreciate a closer look at my implementation there since it's my first time doing the XDR conversion.

types/xdr.d.ts
marcelosalloum: Seems like the problem is the ruby `pluralize` function in [stellar/xdrgen](https://github.com/stellar/xdrgen/blob/9dacf24a73b3e73105e9272ccf333704dcc82333/lib/xdrgen/generators/javascript.rb#L181-L183).

Everything ending with `ta` (or `tA`) is not pluralized, so the assertion below will always be true:
```rb
plural = named.name.downcase.pluralize == named.name.downcase  # true, beacuse `asseta == asseta`
```

marcelosalloum: Adding a fix in https://github.com/stellar/xdrgen/pull/62

paulbellamy (commented):
I think, overall I'm a bit surprised you had to add whole new classes for `TrustLineAsset`, and `ChangeTrustAsset`, I guess in typescript, I expected them to be implemented more like:

```ts
type Asset = NativeAsset | CreditAsset;
type TrustLineAsset = Asset | LiquidityPoolId;
type ChangeTrustAsset = Asset | LiquidityPoolParameters;
```

But I'm not actually sure what the JS version of that would look like, and I suppose this might make the `toXDRObject` a bit messy?

src/operations/revoke_sponsorship.js
```diff
@@ -195,6 +195,42 @@ export function revokeClaimableBalanceSponsorship(opts = {}) {
...
+ * @param {string} [opts.source] - The source account for the operation. Defaults to the transaction's source account.
+ * @returns {xdr.Operation} xdr Operation.
+ *
+ * @example
+ * const op = Operation.revokeLiquidityPoolSponsorship({
+ *   liquidityPoolId: 'dd7b1ab831c273310ddbec6f97870aa83c2fbd78ce22aded37ecbf4f3380fac7',
+ * });
+ *
+ */
+export function revokeLiquidityPoolSponsorship(opts = {}) {
```
paulbellamy, line 214: Might be a bit surprising to have a separate operation here that isn't mirrored in the actual operations? Not sure, tbh...

src/change_trust_asset.js
```diff
@@ -0,0 +1,246 @@
...
+    this.code = code;
+    this.issuer = issuer;
+    if (code) {
+      // return early if it's an asset with code.
+      return;
+    }
+
+    // Validate liquidity pool params.
+    const { asseta, assetB, fee } = liquidityPoolParameters;
+    if (!asseta || !(asseta instanceof Asset)) {
```
paulbellamy, line 59: now that xdrgen is fixed, should these be assetA?

src/trustline_asset.js
paulbellamy: Not sure if we are doing hex here?

src/change_trust_asset.js
```diff
@@ -0,0 +1,246 @@
...
+ * @constructor
+ * @param {string} code - The asset code.
+ * @param {string} issuer - The account ID of the asset issuer.
+ * @param {LiquidityPoolParameters} liquidityPoolParameters – The liquidity pool parameters.
+ * @param {Asset} liquidityPoolParameters.asseta – The first asset in the Pool, it must respect the rule assetA < assetB.
+ * @param {Asset} liquidityPoolParameters.assetB – The second asset in the Pool, it must respect the rule assetA < assetB.
+ * @param {number} liquidityPoolParameters.fee – The liquidity pool fee. For now the only fee supported is `30`.
+ */
+export class ChangeTrustAsset {
+  constructor(code, issuer, liquidityPoolParameters) {
```
tamirms, line 30: I think it would be better to get rid of the constructor and and two static methods to create ChangeTrustAsset instances:

`static creditAsset(code, issuer)`
and
`static poolShareAsset(liquidityPoolParameters)`

src/liquidity_pool_id.js
```diff
@@ -0,0 +1,121 @@
...
+  }
+  if (!assetB || !(assetB instanceof Asset)) {
+    throw new Error('assetB is invalid');
+  }
+  if (!fee || fee != LiquidityPoolFeeV18) {
+    throw new Error('fee is invalid');
+  }
+
+  if (!validateLexicographicalAssetsOrder(asseta, assetB)) {
+    throw new Error('assets are not in lexicografichal order');
```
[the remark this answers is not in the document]
marcelosalloum: Fixed typo in https://github.com/stellar/js-stellar-base/pull/454/commits/92b8e0d2f00cecd1b263dd66ee4c715163a6a61d.

src/change_trust_asset.js
marcelosalloum: I re-generated the variable names in https://github.com/stellar/js-stellar-base/pull/454/commits/714c66b12c4eae4f40dd874f650f75cd248206b9.

types/xdr.d.ts
marcelosalloum: Re-generated the code and fixed those typos in https://github.com/stellar/js-stellar-base/pull/454/commits/714c66b12c4eae4f40dd874f650f75cd248206b9.

src/trustline_asset.js
marcelosalloum: That was a tricky one for me too and I'm not sure this is correct... I did it based on the transaction ID whose representation is a hash-hex of the transaction.

How do you represent liquidity pool IDs in Core @jonjove?

src/change_trust_asset.js
marcelosalloum: That's a great idea! I'm not very experienced in typescript types so I didn't find a clean way to avoid the code repetition... I'll get back to you after I've improved that.

src/trustline_asset.js
jonjove: We represent it as the raw bytes (https://datatracker.ietf.org/doc/html/rfc4506.html#section-4.9) of the hash. It should be represented the same as `ClaimableBalanceID.v0()`. I have basically no context on this code base so I can't easily comment on whether this is right or wrong. I don't know if it will be helpful, but we have tests for liquidity pool IDs in https://github.com/stellar/stellar-core/pull/3165/commits/5e8d7cfe81cc79978a43d39f075f719dae5e2a09.

marcelosalloum: Sharing those samples will be very helpful 🙌 , thank you!

IDs like `c17f36fbd210e43dca1cda8edc5b6c0f825fcb72b39f0392fd6309844d77ff7d` do look like a hex representation but I'll double-check it later. Also, I'll make sure to add tests with the same assets and expected poolID to make sure the JS SDK is using the same logic as core.

jonjove: The tests convert to hex for human readability and writability (it's hard to write binary strings directly into a program). You should convert the hex strings back to raw bytes.

marcelosalloum:
> I think, overall I'm a bit surprised you had to add whole new classes for `TrustLineAsset`, and `ChangeTrustAsset`, I guess in typescript, I expected them to be implemented more like:
> 
> ```ts
> type Asset = NativeAsset | CreditAsset;
> type TrustLineAsset = Asset | LiquidityPoolId;
> type ChangeTrustAsset = Asset | LiquidityPoolParameters;
> ```
> 
> But I'm not actually sure what the JS version of that would look like, and I suppose this might make the `toXDRObject` a bit messy?
> 
> Edit: Ah, it's `ChangeTrustAsset.fromOperation(...)`, that this gives us.. 🤔

I couldn't find a clean way to prevent this much code repetition in Javascript. The objects `Asset`, `TrustLineAsset` and `ChangeTrustAsset` need to be related 1:1 with their `xdr.(.*)Asset` counterparts, which makes it impossible to reuse similar methods like `fromOperation` or `toXDRObject`. Thanks to this, little code would be reusable across those classes and it would come at the cost of increased complexity.

src/change_trust_asset.js
marcelosalloum: Creating the static methods makes a lot of sense to me! If I remove the constructor though, Javascript will use the default constructor:
```js
constructor() {}
```
Since the class won't throw an error if you call it with an empty constructor, my idea is to leave the constructor as is but stop exposing it in [index.d.ts](https://github.com/stellar/js-stellar-base/blob/714c66b12c4eae4f40dd874f650f75cd248206b9/types/index.d.ts#L67), so typescript users will only have access to the static methods. WDYT?

src/trustline_asset.js
marcelosalloum: Hey @jonjove, I've added a few tests in https://github.com/stellar/js-stellar-base/pull/454/commits/54e0070d772fc02745096530c06ddd998860123c mirroring the same tests from stellar-core. We're reaching the same results so I'm confident that the way we're calculating the liquidity pool ID is the same 🙌 .

About hex vs raw bytes, I'm handling the pool ID the same way we already handle claimable balance IDs so I believe we're fine there. The PoolID is represented as a 'hex' string to make it more readable but we convert it to raw bytes when converting the objects to XDR.

src/trustline_asset.js
```diff
@@ -0,0 +1,214 @@
...
+  constructor(code, issuer, liquidityPoolId) {
+    if (!code && !issuer && !liquidityPoolId) {
+      throw new Error('Must provide either code, issuer or liquidityPoolId');
+    }
+    if (code && !/^[a-zA-Z0-9]{1,12}$/.test(code)) {
+      throw new Error(
+        'Asset code is invalid (maximum alphanumeric, 12 characters at max)'
+      );
+    }
+    if (String(code).toLowerCase() !== 'xlm' && !issuer) {
```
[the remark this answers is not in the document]
marcelosalloum: I've copied that from asset.js:
https://github.com/stellar/js-stellar-base/blob/6e0fa3e1a25910e193041d1f377b71f125ec4d1c/src/asset.js#L26

I've updated the condition to check if the variable `code` is not empty:
https://github.com/stellar/js-stellar-base/blob/17bd9f350084501a8ecb628aa0d86f00d38a0033/src/trustline_asset.js#L33

src/change_trust_asset.js
```diff
@@ -0,0 +1,271 @@
...
+ * @param {string} code - The asset code.
+ * @param {string} issuer - The account ID of the asset issuer.
+ * @param {LiquidityPoolParameters} liquidityPoolParameters – The liquidity pool parameters.
+ * @param {Asset} liquidityPoolParameters.assetA – The first asset in the Pool, it must respect the rule assetA < assetB.
+ * @param {Asset} liquidityPoolParameters.assetB – The second asset in the Pool, it must respect the rule assetA < assetB.
+ * @param {number} liquidityPoolParameters.fee – The liquidity pool fee. For now the only fee supported is `30`.
+ */
+export class ChangeTrustAsset {
+  constructor(code, issuer, liquidityPoolParameters) {
+    if (!code && !issuer && !liquidityPoolParameters) {
```
tamirms, line 31: if code is not null then liquidityPoolParameters must be null. If liquidityPoolParameters is not null then code and issuer must be null

src/change_trust_asset.js
tamirms: sounds good to me. I don't think there's any nice way to make a constructor private in javascript

src/change_trust_asset.js
```diff
@@ -0,0 +1,271 @@
...
+    if (issuer && !StrKey.isValidEd25519PublicKey(issuer)) {
+      throw new Error('Issuer is invalid');
+    }
+
+    this.code = code;
+    this.issuer = issuer;
+    if (code) {
+      // return early if it's an asset with code.
+      return;
+    }
```
Shaptic, line 55: This is ripped straight from asset.js, right? We should abstract this into a separate validation layer now that it's duplicated in both places. Maybe `ChangeTrustAsset` can hold an `Asset` object internally?

src/change_trust_asset.js
```diff
@@ -0,0 +1,271 @@
...
+
+    // Validate liquidity pool params.
+    const { assetA, assetB, fee } = liquidityPoolParameters;
```
Shaptic, line 58: woah :o

src/change_trust_asset.js
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    return null;
+  }
+
+  /**
+   * @returns {boolean} `true` if this change trust asset object is the native asset.
+   */
+  isNative() {
+    return this.code && !this.issuer;
```
Shaptic, line 234: Should we validate on `this.code.toLowerCase() === 'xlm'` here?

src/change_trust_asset.js
```diff
@@ -0,0 +1,271 @@
...
+ * issuer will be valid and liquidityPoolParameters will be empty. For liquidity
+ * pools, the liquidityPoolParameters will be valid and the remaining two fields
+ * will be empty.
+ *
+ * @constructor
+ * @param {string} code - The asset code.
+ * @param {string} issuer - The account ID of the asset issuer.
+ * @param {LiquidityPoolParameters} liquidityPoolParameters – The liquidity pool parameters.
+ * @param {Asset} liquidityPoolParameters.assetA – The first asset in the Pool, it must respect the rule assetA < assetB.
+ * @param {Asset} liquidityPoolParameters.assetB – The second asset in the Pool, it must respect the rule assetA < assetB.
```
Shaptic, line 26: Can we drop a reference to something explaining what `<` means here? I see `validateLexicographicAssetsOrder()` which I think is relevant.

test/unit/change_trust_asset_test.js
```diff
@@ -0,0 +1,355 @@
...
+      expect(gotPoolParams.assetA).to.be.equal(assetA);
+      expect(gotPoolParams.assetB).to.be.equal(assetB);
+      expect(gotPoolParams.fee).to.be.equal(fee);
+    });
+  });
+
+  describe('getAssetType()', function() {
+    it('returns "native" for native assets', function() {
+      const asset = StellarBase.ChangeTrustAsset.native();
+      expect(asset.getAssetType()).to.eq('native');
```
Shaptic, line 137: nit: why `.to.be.equal(...)` above and `.to.eq(...)` here?

test/unit/change_trust_asset_test.js
```diff
@@ -0,0 +1,355 @@
...
+        '12345',
+        'GCEZWKCA5VLDNRLN3RPRJMRZOX3Z6G5CHCGSNFHEYVXM3XOJMDS674JZ'
+      );
+      const xdr = asset.toXDRObject();
+
+      expect(xdr).to.be.instanceof(StellarBase.xdr.ChangeTrustAsset);
+      expect(() => xdr.toXDR('hex')).to.not.throw();
+
+      expect(xdr.arm()).to.equal('alphaNum12');
+      expect(xdr.value().assetCode()).to.equal('12345\0\0\0\0\0\0\0');
```
Shaptic, line 214: Where does this extra `\0` suffix come from?

test/unit/change_trust_asset_test.js
```diff
@@ -0,0 +1,355 @@
...
+        '123456789012',
+        'GCEZWKCA5VLDNRLN3RPRJMRZOX3Z6G5CHCGSNFHEYVXM3XOJMDS674JZ'
+      );
+      const xdr = asset.toXDRObject();
+
+      expect(xdr).to.be.instanceof(StellarBase.xdr.ChangeTrustAsset);
+      expect(() => xdr.toXDR('hex')).to.not.throw();
+
+      expect(xdr.arm()).to.equal('alphaNum12');
+      expect(xdr.value().assetCode()).to.equal('123456789012');
```
Shaptic, line 228: I don't like how there are 4 near-identical tests here; can we consolidate them somehow, e.g. into a loop?

test/unit/change_trust_asset_test.js
```diff
@@ -0,0 +1,355 @@
...
+      expect(gotPoolParams.assetA).to.be.equal(assetA);
+      expect(gotPoolParams.assetB).to.be.equal(assetB);
+      expect(gotPoolParams.fee).to.be.equal(fee);
+    });
+  });
+
+  describe('fromOperation()', function() {
+    it('parses a native asset XDR', function() {
+      const xdr = new StellarBase.xdr.ChangeTrustAsset.assetTypeNative();
+      const asset = StellarBase.ChangeTrustAsset.fromOperation(xdr);
```
Shaptic, line 252: This isn't the standard way to do this in the SDK, as far as I know. It should be doable via `Operation.fromXDRObject(xdr)`; do we support that?

Shaptic: Oh, I see `Asset.fromOperation` now, so this is akin to that, gotcha

test/unit/liquidity_pool_id_test.js
```diff
@@ -0,0 +1,265 @@
...
+      )
+    );
+    const issuer2 = StellarBase.StrKey.encodeEd25519PublicKey(
+      Buffer.from(
+        'abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789',
+        'hex'
+      )
+    );
+
+    // NATIVE and ALPHANUM4 (short and full length)
```
Shaptic, line 87: It would be more descriptive if these comments were test names, e.g.

```js
it("works when mixing native and alpha4 assets",  ...)
```

Thoughts?

types/index.d.ts
```diff
@@ -58,6 +60,26 @@ export class Asset {
...
+  isNative(): boolean;
+  isLiquidityPool(): boolean;
+  equals(other: ChangeTrustAsset): boolean;
+  toXDRObject(): xdr.ChangeTrustAsset;
+
+  code: string;
+  issuer: string;
+  liquidityPoolParameters: LiquidityPoolParameters;
+}
+
```
Shaptic, line 82: Again, maybe this should inherit from `Asset`, or otherwise have some common interface?

Shaptic (commented):
Catching up here after vacation and left some notes.

My biggest architectural concern is the seemingly large amount of duplication between `Asset` and the new `*Asset` types; if we can consolidate their shared code I'd be much happier.

Secondary question: are we sure this is a breaking change? From the way I read things, the new asset types are compatible with the old type (e.g. if I had `Asset(code, issuer)` before then it will continue to work, and `ChangeTrustAsset(code, issuer)` will also work). You can see for example that `change_trust.js` didn't change at all (except the docstring), which tells me this isn't actually a breaking change. If the intent *is* a hard break, we should introduce type-checks or some other validation here, then.

test/unit/change_trust_asset_test.js
marcelosalloum: No particular reason. Seems like this test suite allows many different ways of writing and I got multiple ones due to cp[y & paste. I'll update the code to use only one variation.

test/unit/change_trust_asset_test.js
marcelosalloum: This test is already part of asset_test.js:
https://github.com/stellar/js-stellar-base/blob/6e0fa3e1a25910e193041d1f377b71f125ec4d1c/test/unit/asset_test.js#L143

I'm not sure about the motivation for it but `\0` is used to pad the asset code when converting it to XDR:
https://github.com/stellar/js-stellar-base/blob/6e0fa3e1a25910e193041d1f377b71f125ec4d1c/src/asset.js#L91

test/unit/change_trust_asset_test.js
marcelosalloum: Fixed in https://github.com/stellar/js-stellar-base/pull/454/commits/fbec065664b70070037e631190ddb1e9157391a8.

src/change_trust_asset.js
marcelosalloum: Do you think this works? https://github.com/stellar/js-stellar-base/pull/454/commits/35771f2bb76ef89e08100bcab55fadde27c544f8

I didn't want to duplicate the information here so I just referenced `validateLexicographicAssetsOrder `.

src/change_trust_asset.js
marcelosalloum: Fixed in https://github.com/stellar/js-stellar-base/pull/454/commits/19e30f661a8ecdfe48bd548167630d2cccfd50e0.

src/change_trust_asset.js
marcelosalloum: I'll give it some thought... 🤔

src/change_trust_asset.js
Shaptic: Yeah something like that is all I was looking for, or even a `@see validateLexicographicAssetsOrder`; thanks!

src/change_trust_asset.js
marcelosalloum: Included that suggestion in https://github.com/stellar/js-stellar-base/pull/454/commits/2de22a943f293acdef42b9bec0a4478359bc7421.

types/index.d.ts
marcelosalloum: I'm trying inheritance in this commit: https://github.com/stellar/js-stellar-base/pull/454/commits/83ab66bc0ba406607c06b4328f98a51ad14bc448. I'm not very happy with it though because the `Asset` class has become a bit dirty as some method signatures were updated to accept not only Asset but also ChangeTrustAsset.

The other options could include generics (I think it'd get even dirtier) and some level of feature composition... these are not topics I'm experienced in JS or TS though so I wanted to hear from you before diving into that.

test/unit/liquidity_pool_id_test.js
marcelosalloum: Good idea! Done in https://github.com/stellar/js-stellar-base/pull/454/commits/b0060f2c510cf51bb1048265a326d138aab67c6c

test/unit/change_trust_asset_test.js
marcelosalloum: When we're talking about tests I prefer being verbose and have more readability. Using a for loop here would save us 2 or 3 lines per test at the cost of reduced readability so I don't think it's worth it for this particular case. WDYT?

marcelosalloum:
> You can see for example that `change_trust.js` didn't change at all (except the docstring), which tells me this isn't actually a breaking change. If the intent _is_ a hard break, we should introduce type-checks or some other validation here, then.

If you pass an `Asset` to the changeTrust method it will break because the code below throws an error if the asset XDR representation is not `xdr.ChangeTrustAsset`:
https://github.com/stellar/js-stellar-base/blob/b0060f2c510cf51bb1048265a326d138aab67c6c/src/operations/change_trust.js#L40

The error returned is:
Screen Shot 2021-08-24 at 12 04 48

src/change_trust_asset.js
```diff
@@ -0,0 +1,185 @@
...
+   * Returns a change trust asset object of the type "liquidity pool shares".
+   * @static
+   * @param {LiquidityPoolParameters} liquidityPoolParameters – The liquidity pool parameters.
+   * @param {Asset} liquidityPoolParameters.assetA – The first asset in the Pool, it must respect the rule assetA < assetB.
+   * @param {Asset} liquidityPoolParameters.assetB – The second asset in the Pool, it must respect the rule assetA < assetB.
+   * @param {number} liquidityPoolParameters.fee – The liquidity pool fee. For now the only fee supported is `30`.
+   * @return {ChangeTrustAsset}
+   * @memberof ChangeTrustAsset
+   */
+  static liquidityPoolAsset(liquidityPoolParameters) {
```
paulbellamy, line 87: WDYT of `liquidityPoolShare`?
```suggestion
  static liquidityPoolShare(liquidityPoolParameters) {
```

src/asset.js
```diff
@@ -44,8 +45,8 @@ export class Asset {
 
   /**
    * Returns an asset object from its XDR object representation.
-   * @param {xdr.Asset} assetXdr - The asset xdr object.
-   * @returns {Asset}
+   * @param {xdr.Asset | xdr.ChangeTrustAsset} assetXdr - The asset xdr object.
```
paulbellamy, line 48: This, no?
```suggestion
   * @param {xdr.Asset | xdr.ChangeTrustAsset | xdr.TrustLineAsset} assetXdr - The asset xdr object.
```

src/trustline_asset.js
```diff
@@ -0,0 +1,236 @@
+import clone from 'lodash/clone';
+import padEnd from 'lodash/padEnd';
+import trimEnd from 'lodash/trimEnd';
+import xdr from './generated/stellar-xdr_generated';
+import { Keypair } from './keypair';
+import { StrKey } from './strkey';
+
+/**
+ * TrustLineAsset class represents a trustline to either a liquidity pool, the
```
paulbellamy, line 9:
```suggestion
 * TrustLineAsset class represents the asset referenced by a trustline to either a liquidity pool, the
```

types/index.d.ts
```diff
@@ -107,6 +122,28 @@ export class Keypair {
   xdrMuxedAccount(id: string): xdr.MuxedAccount;
 }
 
+export const LiquidityPoolFeeV18 = 30;
+
+export function getLiquidityPoolId(liquidityPoolType: LiquidityPoolType, liquidityPoolParameters: LiquidityPoolParameters): Buffer;
+
+export function validateLexicographicAssetsOrder(assetA: Asset, assetB: Asset): boolean;
```
paulbellamy, line 129: I would be tempted to do something like:
`Asset.isSorted(assets ...Asset): boolean`, and something like a `Asset.compare(a: Asset, b: Asset): number` so you could do `[a, b].sortBy(Asset.compare)` as well.

Maybe that's overengineering it... wdyt?

paulbellamy (commented):
Looking good, IMO. Will leave final approval to @Shaptic, though.

src/trustline_asset.js
marcelosalloum: https://github.com/stellar/js-stellar-base/pull/454/commits/8a22a6deaa092d1d5d88ce97a424dfbe047045e3

src/change_trust_asset.js
marcelosalloum: Good suggestion. Done in https://github.com/stellar/js-stellar-base/pull/454/commits/185002ed3c232306b01057dd31d01d513574960c

src/asset.js
marcelosalloum: This is incomplete, TrustLineAsset is not inheriting from Asset yet. I'm waiting for @.Shaptic to reply in https://github.com/stellar/js-stellar-base/pull/454#discussion_r694852344 before doing that.

types/index.d.ts
marcelosalloum: `Asset.compare` is a much better place for that kind of logic, thank you! 
I've updated that in https://github.com/stellar/js-stellar-base/commit/11537633dfa7eed32f92931418165d23882fd56b.

I don't think we need `Asset.isSorted` though, `Asset.compare` should be enough since it's a common pattern in javascript objects.

types/index.d.ts
Shaptic: This is a tough call.. on one hand there is much less duplicated code in this commit (nice!) but as you say it does require a little bit of shoe-horning in terms of the type signatures. I do think this version will be easier to maintain (since all validation is now consolidated), and I left a few more thoughts in the commit itself.

Shaptic:
Oh jeez, GitHub's interface for commenting on commits is *not* very user-friendly... It should be easier to grok if you click the commit I'm commenting on (8f960519536161d877e3ff34f0d715912fbed65a and 3e354aac856deab034f453f1bdec586c3c015e15), sorry.

types/index.d.ts
marcelosalloum: We agreed (offline) to handle the code repetition in a separate PR to make it easier to review and discuss.

src/change_trust_asset.js
```diff
@@ -0,0 +1,280 @@
...
+    const { assetA, assetB, fee } = liquidityPoolParameters;
+    if (!assetA || !(assetA instanceof Asset)) {
+      throw new Error('assetA is invalid');
+    }
+    if (!assetB || !(assetB instanceof Asset)) {
+      throw new Error('assetB is invalid');
+    }
+    if (Asset.compare(assetA, assetB) !== -1) {
+      throw new Error('Assets are not in lexicographic order');
+    }
```
Shaptic, line 73: Out of scope but worth asking: Should we enforce this, or would it be better to just be helpful and order them ourselves?

src/change_trust_asset.js
```diff
@@ -0,0 +1,280 @@
...
+   * @static
+   * @param {LiquidityPoolParameters} liquidityPoolParameters – The liquidity pool parameters.
+   * @param {Asset} liquidityPoolParameters.assetA – The first asset in the Pool, it must respect the rule assetA < assetB.
+   * @param {Asset} liquidityPoolParameters.assetB – The second asset in the Pool, it must respect the rule assetA < assetB.
+   * @param {number} liquidityPoolParameters.fee – The liquidity pool fee. For now the only fee supported is `30`.
+   * @return {ChangeTrustAsset}
+   * @memberof ChangeTrustAsset
+   */
+  static liquidityPoolShare(liquidityPoolParameters) {
+    return new ChangeTrustAsset(undefined, undefined, liquidityPoolParameters);
```
Shaptic, line 104: Okay, I'm not sure why I didn't consider this before, but can we step back for a second: **why are we even trying to mimic the XDR like this?** Just because the XDR makes this bizarre design choice in which it can hold the native asset, 4-char assets, 12-char assets, **or** liquidity pool parameters doesn't mean we need to follow that. 

We should provide user-friendly APIs, instead, so if `Asset(code, issuer)` is *exactly the same* as `ChangeTrustAsset(code, issuer)` (right?), why not *only* have something like `LiquidityPoolAsset(assetA, assetB, fee)` and just make people use `Asset` when they want, well, an asset (either native or code:issuer)? 

And if binary compatibility is a concern, we can have something like an `Asset.asChangeTrustXDR()` helper, instead.

src/change_trust_asset.js
```diff
@@ -0,0 +1,280 @@
...
+      return 'liquidity_pool_shares';
+    }
+    if (this.code.length >= 1 && this.code.length <= 4) {
+      return 'credit_alphanum4';
+    }
+    if (this.code.length >= 5 && this.code.length <= 12) {
+      return 'credit_alphanum12';
+    }
+
+    return null;
```
Shaptic, line 236: Thoughts on simplifying this?

```suggestion
    if (this.code.length <= 4) {
      return 'credit_alphanum4';
    }
    return 'credit_alphanum12';
```

src/change_trust_asset.js
```diff
@@ -0,0 +1,280 @@
...
+
+  /**
+   * @param {ChangeTrustAsset} asset the ChangeTrustAsset to compare
+   * @returns {boolean} `true` if this asset equals the given asset.
+   */
+  equals(asset) {
+    return (
+      this.code === asset.getCode() &&
+      this.issuer === asset.getIssuer() &&
+      this.liquidityPoolParameters === asset.getLiquidityPoolParameters()
```
Shaptic, line 261: Does this do a deep compare like you'd expect?

src/liquidity_pool_id.js
```diff
@@ -0,0 +1,51 @@
...
+  const { assetA, assetB, fee } = liquidityPoolParameters;
+  if (!assetA || !(assetA instanceof Asset)) {
+    throw new Error('assetA is invalid');
+  }
+  if (!assetB || !(assetB instanceof Asset)) {
+    throw new Error('assetB is invalid');
+  }
+  if (!fee || fee !== LiquidityPoolFeeV18) {
+    throw new Error('fee is invalid');
+  }
```
Shaptic, line 37: Are you sure we should we enforce the fee value?

src/operations/change_trust.js
```diff
@@ -2,24 +2,30 @@ import isUndefined from 'lodash/isUndefined';
...
  *                                If the limit is set to "0" it deletes the trustline.
  * @param {string} [opts.source] - The source account (defaults to transaction source).
  * @returns {xdr.ChangeTrustOp} Change Trust operation
  */
 export function changeTrust(opts) {
   const attributes = {};
+
+  if (!(opts.asset instanceof ChangeTrustAsset)) {
+    throw new TypeError('options.asset must be a ChangeTrustAsset');
+  }
```
Shaptic, line 27: Since this error will happen in the XDR anyway I'm actually okay without introducing this check here. We don't do this in other places so it's better to be consistent.

src/operations/liquidity_pool_deposit.js
```diff
@@ -0,0 +1,59 @@
...
+ * @returns {xdr.Operation} The resulting operation (xdr.LiquidityPoolDepositOp).
+ */
+export function liquidityPoolDeposit(opts = {}) {
+  const attributes = {};
+  if (!opts.liquidityPoolId) {
+    throw new TypeError('liquidityPoolId argument is required');
+  }
+  attributes.liquidityPoolId = xdr.PoolId.fromXDR(opts.liquidityPoolId, 'hex');
+
+  if (isUndefined(opts.maxAmountA)) {
```
Shaptic, line 32: I think using `Operation.isValidAmount` would be a better validation here.

src/operations/liquidity_pool_withdraw.js
```diff
@@ -0,0 +1,49 @@
...
+ * @returns {xdr.Operation}   The resulting operation (xdr.LiquidityPoolWithdrawOp).
+ */
+export function liquidityPoolWithdraw(opts = {}) {
+  const attributes = {};
+  if (!opts.liquidityPoolId) {
+    throw new TypeError('liquidityPoolId argument is required');
+  }
+  attributes.liquidityPoolId = xdr.PoolId.fromXDR(opts.liquidityPoolId, 'hex');
+
+  if (isUndefined(opts.amount)) {
```
Shaptic, line 27: Same here as above: `this.isValidAmount(opts.amount)` is the preferred way to validate asset quantities.

src/operations/revoke_sponsorship.js
```diff
@@ -63,7 +63,7 @@ export function revokeTrustlineSponsorship(opts = {}) {
   if (!StrKey.isValidEd25519PublicKey(opts.account)) {
     throw new Error('account is invalid');
   }
-  if (!(opts.asset instanceof Asset)) {
+  if (!(opts.asset instanceof TrustLineAsset)) {
```
Shaptic, line 66: Oh hm I guess we *do* do a type-check on assets in other places... Okay, up to you to keep it above then!

src/trustline_asset.js
```diff
@@ -0,0 +1,236 @@
...
+      );
+    }
+    if (code && String(code).toLowerCase() !== 'xlm' && !issuer) {
+      throw new Error('Issuer cannot be null');
+    }
+    if (issuer && !StrKey.isValidEd25519PublicKey(issuer)) {
+      throw new Error('Issuer is invalid');
+    }
+
+    // Test if liquiditry pool ID is a hash.
```
Shaptic, line 48: nit: typo
```suggestion
    // Test if liquidity pool ID is a hash.
```

src/trustline_asset.js
```diff
@@ -0,0 +1,236 @@
...
+    if (liquidityPoolId && !/^[a-fA-F0-9]{64}$/.test(liquidityPoolId)) {
```
Shaptic, line 49: Can you confirm that there won't be casing issues here? The XDR always uses lowercase letters in its hashes and I just want to make sure that doesn't cause problems. To be extra safe, we can do 
```js
this.liquidityPoolId = liquidityPoolId.toLowerCase();
```
below, instead.

src/trustline_asset.js
```diff
@@ -0,0 +1,236 @@
...
+
+  /**
+   * Returns a trustline asset object of the type "liquidity pool ID".
+   * @static
+   * @param {string} liquidityPoolId - The ID of the liquidity pool in string 'hex'.
+   * @return {TrustLineAsset}
+   * @memberof TrustLineAsset
+   */
+  static liquidityPoolShare(liquidityPoolId) {
+    return new TrustLineAsset(undefined, undefined, liquidityPoolId);
```
Shaptic, line 79: Same comment applies here as above: maybe we don't care about providing a 1-to-1 XDR mapping and represent liquidity pools however we want while also having converters for binary compatibility.

types/index.d.ts
```diff
@@ -305,7 +352,7 @@ export namespace OperationOptions {
     authorize?: boolean | TrustLineFlag;
   }
   interface ChangeTrust extends BaseOptions {
-    asset: Asset;
+    asset: ChangeTrustAsset;
```
Shaptic, line 355: Need a similar change for `TrustLineAsset` and `RevokeTrustlineSponsorship`, right?

Shaptic (approved):
Okay, left another bunch of brain-dump-y comments but I'm generally on board with this change. We can iterate on the bigger points in subsequent PRs! Please address some of the minor nits and then merge when you're ready :heavy_check_mark:

src/change_trust_asset.js
marcelosalloum: In [this offline thread](https://stellarfoundation.slack.com/archives/C01PNKPSY3H/p1629393624105600) we decided not to change the order of the assets when building liquidity pool deposits, withdrawals or when building the liquidity pool ID. Since the logic here is pretty similar to the getLiquidityPoolId method I kept the logic consistent for both of them.

The argument was that changing the assets orders at any point could bring unforeseen downsides. For instance, if you build a transaction with: `[LP ChangeTrust, LP Deposit, LP Withdrawal]` and you use the same assets for all these operations you should not change the asset order for any of them at the risk of failing the transaction or messing with the deposit/withdrawal amounts.

src/liquidity_pool_id.js
marcelosalloum: According to [CAP-38] a change trust operation is considered invalid if the fee is different from LiquidityPoolFeeV18. I chose to fail earlier here rather than let the method `toXDR()` fail at a later time.

[CAP-38]: https://github.com/stellar/stellar-protocol/blob/master/core/cap-0038.md#changetrustop

src/trustline_asset.js
marcelosalloum: https://github.com/stellar/js-stellar-base/pull/454/commits/24c520f4f3be6fdf541ea89035a6021b6d40c799

src/change_trust_asset.js
marcelosalloum: Improved in https://github.com/stellar/js-stellar-base/pull/454/commits/24c520f4f3be6fdf541ea89035a6021b6d40c799

src/asset.js
```diff
@@ -157,4 +157,63 @@ export class Asset {
...
+
+    // Compare asset codes.
+    switch (assetA.getCode().localeCompare(assetB.getCode())) {
+      case -1: // assetA < assetB
+        return -1;
+      case 1: // assetA > assetB
+        return 1;
+      default:
+        break;
+    }
```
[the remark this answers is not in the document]
marcelosalloum: Improved in https://github.com/stellar/js-stellar-base/pull/454/commits/24c520f4f3be6fdf541ea89035a6021b6d40c799.

types/index.d.ts
marcelosalloum: Good catch! Added the missing type change in https://github.com/stellar/js-stellar-base/pull/454/commits/6bba221f910a322d1979ab1737c17296622aa731.

src/trustline_asset.js
marcelosalloum: Added this new validation in https://github.com/stellar/js-stellar-base/pull/454/commits/f32e27277343e6cc4240c95ac241878ce9e50781.

src/operations/liquidity_pool_deposit.js
marcelosalloum: TIL! Changed to `Operation.isValidAmount` in https://github.com/stellar/js-stellar-base/pull/454/commits/43de4f1ed5d0abd37faa3b28401d8f5b632e52a3.

src/operations/liquidity_pool_withdraw.js
marcelosalloum: Cool! changed to `Operation.isValidAmount` in https://github.com/stellar/js-stellar-base/pull/454/commits/43de4f1ed5d0abd37faa3b28401d8f5b632e52a3.

src/operations/revoke_sponsorship.js
marcelosalloum: That's where I got the inspiration to add this validation 😅 . I've decided to let the validations there but I'm open to removing them if you feel strongly towards it.

src/operations/change_trust.js
marcelosalloum: I've decided to let it there but I'm open to discussing it in https://github.com/stellar/js-stellar-base/pull/454#discussion_r695798768.

src/change_trust_asset.js
marcelosalloum: Wow, that's a great idea 👏 ! I'll play around with that and ask for your review later today.

Also, thanks a lot for putting some extra thought into this, it will make the implementation and usage much cleaner and remove the breaking changes! 🎉

src/operations/revoke_sponsorship.js
Shaptic: Yep, makes sense! I'm totally fine with keeping it.

src/liquidity_pool_id.js
Shaptic: Okay, great! Thanks for the reference to the CAP; it's interesting that they chose to do it this way but since they do, we should adhere to it :+1:

jonjove: @Shaptic why are you interested/surprised by the choice in the CAP?

PR #457, opened 2021-08-23
Create GitHub Action to test the repo

marcelosalloum:


karma.conf.js
```diff
@@ -5,7 +5,7 @@ delete webpackConfig.output;
 module.exports = function(config) {
   config.set({
     frameworks: ['mocha', 'sinon-chai'],
-    browsers: ['Firefox'],
+    browsers: ['ChromeHeadless'],
```
Shaptic, line 8: Rationale?

marcelosalloum: I'm having a hard time running browser tests in Github Actions. `karma` was failing when trying to launch Firefox and I saw a SO thread recommending to use `ChromeHeadless` instead for CI.

marcelosalloum: Seems karma can only launch headless versions of browsers in Github Actions. I've updated it to FirefoxHeadless and it works well too.

.github/workflows/gulp.yml
```diff
@@ -0,0 +1,36 @@
...
+
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+
+    - name: Build
+      run: |
+        npm install
+        gulp
```
leighmcculloch, line 27: If you can it would be clearer if these can be separate run commands.

Also, the readme says to use yarn to install, so I think the deps install should be `yarn`.

```suggestion
    - name: Install deps
      run: yarn

    - name: Build
      run: gulp
```

README.md
```diff
@@ -1,6 +1,5 @@
 # JS Stellar Base
 
-[![Build Status](https://travis-ci.com/stellar/js-stellar-base.svg?branch=master)](https://travis-ci.com/stellar/js-stellar-base)
```
leighmcculloch, line 3: FYI you can create a new badge for GitHub Actions at this URL, you click the `...` in the top right:

https://github.com/stellar/js-stellar-base/actions/workflows/gulp.yml

leighmcculloch (commented):
Thanks for updating the CI in this repo! Couple comments, otherwise looks great to me.

README.md
marcelosalloum: Cool! Done in https://github.com/stellar/js-stellar-base/pull/457/commits/a03ba59f35e894747f5f45d18188b7d8510fc6ef

.github/workflows/gulp.yml
marcelosalloum: https://github.com/stellar/js-stellar-base/pull/457/commits/ae2b2d66a7a3d12c38ee76a7477e533a9d5ac44b

.github/workflows/tests.yml
```diff
@@ -0,0 +1,37 @@
...
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+
+    - name: Install deps
+      run: yarn
```
leighmcculloch, line 25: You could also cache and restore the dependencies by using [actions/cache](https://github.com/actions/cache/blob/main/examples.md#node---yarn-2), but I think it is only worth doing if the dep install is lengthy and slow.

leighmcculloch (approved):
Looks great! Thanks!

karma.conf.js
Shaptic: Makes sense, thanks for resolving this! Do you think we should run it on both FF and Chrome?

Shaptic (approved):
I know this is just in draft but I think this would be great to merge! One thing, though: can we keep around the Travis CI file? I know it's not in use anymore, but since we haven't converted *all* of the functionality, it'll be easier to reference if it's still in the repo.

leighmcculloch:
> Missing commands (CD):
> * `gulp submit-coverage`
> * `./bower_publish.sh`
> * `./after_deploy.sh`

-- https://github.com/stellar/js-stellar-base/pull/457#issue-718129079

> since we haven't converted _all_ of the functionality

-- https://github.com/stellar/js-stellar-base/pull/457#pullrequestreview-737627838

There was an internal discussion last week where I think we worked out we aren't using the Sauce or Code Coverage, and so they can be ignored/dropped. The only thing left, I think, is the deploy to npm and bowerjs.

karma.conf.js
marcelosalloum: Sure, done in https://github.com/stellar/js-stellar-base/pull/457/commits/9623cb5742ca1cafb6c9f3242420967df4a2764e.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,43 @@
...
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+
+    - name: Cache node_modules
+      uses: actions/cache@v2
+      with:
+        path: '**/node_modules'
+        key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
```
leighmcculloch, line 28: If you're going to cache, this needs a little tweaking. You need each node version to use a different cache, so the matrix node version needs including into the key. Also, ideally you pull a recent cache if the current cache isn't available. You can do that by using the restore key.
```suggestion
        key: ${{ runner.os }}-${{ matrix.node-version }}-modules-${{ hashFiles('**/yarn.lock') }}
        restore-keys: ${{ runner.os }}-${{ matrix.node-version }}-modules
```

leighmcculloch (commented):
Still looks great, but I suggest a tweak to the caching.

.github/workflows/tests.yml
leighmcculloch: Actually, it looks like you're caching node_modules. According to GitHub Actions docs:
>Note: It is not recommended to cache node_modules, as it can break across Node versions and won't work with npm ci

I thought because we were using yarn that we could use the setup here to cache the packages in the yarn cache?

https://github.com/actions/cache/blob/main/examples.md#node---yarn-2

.github/workflows/tests.yml
```diff
@@ -0,0 +1,51 @@
...
+      with:
+        node-version: ${{ matrix.node-version }}
+
+    - name: Get yarn cache directory path
+      id: yarn-cache-dir-path
+      run: echo "::set-output name=dir::$(yarn config get cacheFolder)"
+
+    - name: Get cached dependencies
+      uses: actions/cache@v2
+      id: yarn-cache # use this to check for `cache-hit` (`steps.yarn-cache.outputs.cache-hit != 'true'`)
```
leighmcculloch, line 30: You don't need this because you're not referring to this step elsewhere.
```suggestion
```

.github/workflows/tests.yml
marcelosalloum: I didn't manage to make the cache work and since the build step is only taking ~30s I think we could merge this without the cache part. WDYT?

leighmcculloch: 💯

PR #465, opened 2021-09-08
Build docs after every push to `master` for CI testing purposes.

Shaptic:


.github/workflows/gh_pages.yml
```diff
@@ -1,30 +1,30 @@
 name: Github Pages
 on:
-  release:
-    types: [published]
+  push:
+    branches:
+      - master
```
marcelosalloum, line 5: ❗  With the old trigger (release published) the gh-pages would be updated only when we launch a release. With this change though it will be updated every time a new code is merged into master. Is that your intention?

Nowadays you can re-run this script simply by converting a release to draft and then re-publish it again.

.github/workflows/gh_pages.yml
```diff
@@ -1,30 +1,30 @@
...
 jobs:
   build-and-deploy:
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout
```
marcelosalloum, line 9: Any reason for removing the step name?

.github/workflows/gh_pages.yml
```diff
@@ -1,30 +1,30 @@
...
+
       - name: Build
         run: gulp
 
       - name: Generate JSDoc
-        run: yarn jsdoc -c .jsdoc.json --verbose
+        run: yarn docs
 
       - name: Deploy to GH Pages
-        uses: JamesIves/github-pages-deploy-action@0f24da7de3e7e135102609a4c9633b025be8411b
```
marcelosalloum, line 27: ❗  Since this is not a GH official action we should be pointing to the commit hash rather than a tag version, as recommended by [github](https://docs.github.com/en/actions/learn-github-actions/security-hardening-for-github-actions#using-third-party-actions):

>**Pin actions to a full length commit SHA**
>
>Pinning an action to a full length commit SHA is currently the only way to use an action as an immutable release. Pinning to a particular SHA helps mitigate the risk of a bad actor adding a backdoor to the action's repository, as they would need to generate a SHA-1 collision for a valid Git object payload.
>
>**Pin actions to a tag only if you trust the creator**
>
>Although pinning to a commit SHA is the most secure option, specifying a tag is more convenient and is widely used. If you’d like to specify a tag, then be sure that you trust the action's creators. The ‘Verified creator’ badge on GitHub Marketplace is a useful signal, as it indicates that the action was written by a team whose identity has been verified by GitHub. Note that there is risk to this approach even if you trust the author, because a tag can be moved or deleted if a bad actor gains access to the repository storing the action.

marcelosalloum: This is a security measure as it is possible for a tag to change the commit sha it's pointing to.

.github/workflows/gh_pages.yml
```diff
@@ -1,30 +1,30 @@
...
+        uses: JamesIves/github-pages-deploy-action@4.1.5
         with:
           branch: gh-pages # The branch the action should deploy to.
           folder: jsdoc # The folder the action should deploy.
```
marcelosalloum, line 30: I think what we're missing here is a Github Personal Access Token (PAT), according with the action documentation: https://github.com/JamesIves/github-pages-deploy-action#required-setup

marcelosalloum (commented):
I think what we really need is to add a `token: ${{ secrets.GITHUB_TOKEN }}` as a `with` parameter under the action `github-pages-deploy-action`

.github/workflows/gh_pages.yml
marcelosalloum: Update: I was wrong about the GH token being needed here, I've updated the gh-pages from my local machine so the permissions should be right. The problem could be in the action we're using.

TBH we shouldn't need an Action here... we could probably just use our own script to do something like this:
https://github.com/stellar/js-stellar-base/blob/a37e12558c04d26dc356ede476ac0ecb4036843f/after_deploy.sh

.github/workflows/gh_pages.yml
Shaptic: That was the intent, yeah, so that we could iterate on the file without publishing a release (since that would trigger the whole `npm_publish` routine and is generally more involved).

.github/workflows/gh_pages.yml
Shaptic: Hm, for some reason, I thought that if you look at a [particular build breakdown](https://github.com/stellar/js-stellar-base/runs/3437811362?check_suite_focus=true) and opened the "Checkout" tab, there were no results. I figured that made it makes sense to drop the name since there would never be a reason to look at it, but now I see that there *are* logs... maybe they were slow to load, or something. I'll put it back!

.github/workflows/gh_pages.yml
Shaptic: Good call! I should've asked why we used a commit hash instead of changing it haha. I'll point it to the v4.1.5 SHA, instead!

.github/workflows/gh_pages.yml
Shaptic: Tradeoffs, right? With an action, we "outsource" the process and don't need to worry about publishing, but OTOH we have less control and it's harder to debug (like here)...

For `after_deploy`, how do we set up the `stellar-jenkins@` credential properly? I'll follow-up with ops.

marcelosalloum: We shouldn't need that for the GH actions, it's legacy from the Travis CI. We could just use run the commands directly from the yml file, something in the lines of https://github.com/actions/checkout#push-a-commit-using-the-built-in-token:

```yml
on: push
jobs:
  build:
    runs-on: ubuntu-latest
    steps:
      - name: Prepare & Publish Github Pages
        uses: actions/checkout@v2
        run: |
          git config user.name github-actions
          git config user.email github-actions@github.com
          git clone -b gh-pages "git@github.com:stellar/js-stellar-base.git" jsdoc
          yarn docs
          git commit -m `git describe --abbrev=0 --tags`
          git push origin
```

Shaptic: Oh wow, very nice. Have a look at the change now? Done in 7d42fa6, then broken down into discrete steps in 0c75b47.

.github/workflows/gh_pages.yml
```diff
@@ -1,30 +1,41 @@
...
-        with:
-          branch: gh-pages # The branch the action should deploy to.
-          folder: jsdoc # The folder the action should deploy.
+      - name: Generate JS docs
+        run: yarn docs
+
+      - name: Publish JS docs
+        uses: actions/checkout@v2
+        run: |
+          DESCRIPTION=`git describe --tags`
```
marcelosalloum, line 37:
```suggestion
          DESCRIPTION=`git describe --abbrev=0 --tags`
```

If you don't use the `--abbrev=0` option, the DESCRIPTION will contain a little more information than just the latest tag.

```
git describe --abbrev=0 --tags  // returns "v6.0.1"
git describe --tags             // returns "v6.0.1-5-gd073684"
```

Is that your intention? The commit messages from branch [gh-pages](https://github.com/stellar/js-stellar-base/commits/gh-pages) are in the format `v6.0.1`.

marcelosalloum (approved):
It LGTM! 

There's a chance we're missing the [`permissions`] configuration, which may cause the `git push` command to fail, according to GitHub [docs].

I suggest we move forward with this and test if it works.

 [`permissions`]: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#setting-the-permissions-of-the-github_token-for-your-repository
[docs]: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#setting-the-permissions-of-the-github_token-for-your-repository

.github/workflows/gh_pages.yml
Shaptic: Yeah, I should've mentioned that I did this intentionally. I'm obviously open to critique, but I really like that this creates a tighter relationship between the library and the docs. This was partially inspired by what you said above: tag commits can change.

PR #466, opened 2021-09-13
Add a `withMuxing` parameter to all operation options in Typescript.

Shaptic:
### What
This allows all operations to set a muxed source account, as required by [CAP-27](https://stellar.org/protocol/cap-27).

### Why
While JavaScript consumers of this library are unaffected, as passing extra parameters to functions is always allowed, Typescript libraries will not compile when passing `withMuxing: true` to any operation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+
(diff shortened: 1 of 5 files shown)


JakeUrban (approved):
LGTM, thanks for the quick turnaround

PR #467, opened 2021-09-13
Fix logical errors in the GitHub Action for docgen.

Shaptic:
Whelp, that broke fast ... (see [build](https://github.com/stellar/js-stellar-base/actions/runs/1230815646)) :laughing: 

A misinterpretation of [the docs](https://github.com/actions/checkout#push-a-commit-using-the-built-in-token) on my part: The `uses` and `run` steps should be separate, and the former is already part of the first build step.


marcelosalloum (approved):
LGTM!

PR #482, opened 2022-01-05
Adds helper to predict claimable balance IDs given a transaction.

Shaptic:
This change introduces a helper method 

```typescript
Transaction.getClaimableBalanceId(opIndex: number): string;
```

which lets you predetermine the hex claimable balance ID of a `createClaimableBalance` operation prior to submission to the network.

Closes part (1) of stellar/js-stellar-sdk#584.

### DIFF
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -221,16 +221,15 @@ export class Transaction extends TransactionBase {
   /**
    * Calculate the claimable balance ID for an operation within the transaction.
    *
-   * This helper throws under the following circumstances:
-   *    - invalid `opIndex` value
-   *    - the operation at the `opIndex` is not a CreateClaimableBalance op
-   *    - the source account doesn't have a sequence number
-   *    - general XDR un/marshalling failures
-   *
    * @param   {integer}  opIndex   the index of the CreateClaimableBalance op
    * @returns {string}   a hex string representing the claimable balance ID
    *
-   * @see https://github.com/stellar/go/blob/master/txnbuild/transaction.go#L392
+   * @throws {RangeError}   for invalid `opIndex` value
+   * @throws {TypeError}    if op at `opIndex` is not `CreateClaimableBalance`
+   * @throws for general XDR un/marshalling failures
+   *
+   * @see https://github.com/stellar/go/blob/d712346e61e288d450b0c08038c158f8848cc3e4/txnbuild/transaction.go#L392-L435
+   *
    */
   getClaimableBalanceId(opIndex) {
     // Validate and then extract the operation from the transaction.
@@ -251,9 +250,8 @@ export class Transaction extends TransactionBase {
       );
     }
 
-    // Use the operation's source account or the transaction's source if not. In
-    // either case, use the unmuxed version.
-    let account = decodeAddressToMuxedAccount(this.source, true);
+    // Always use the transaction's *unmuxed* source.
+    let account = decodeAddressToMuxedAccount(this.source);
     if (account.switch() === xdr.CryptoKeyType.keyTypeMuxedEd25519()) {
       account = account.med25519();
     }
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -8,7 +8,7 @@ import { Operation } from './operation';
 import { Memo } from './memo';
 import { TransactionBase } from './transaction_base';
 import {
-  decodeAddressToMuxedAccount,
+  extractBaseAddress,
   encodeMuxedAccountToAddress
 } from './util/decode_encode_muxed_account';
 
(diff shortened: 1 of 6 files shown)


src/transaction.js
```diff
@@ -222,4 +226,57 @@ export class Transaction extends TransactionBase {
...
   }
+
+  /**
+   * Calculate the claimable balance ID for an operation within the transaction.
+   *
+   * This helper throws under the following circumstances:
+   *    - invalid `opIndex` value
+   *    - the operation at the `opIndex` is not a CreateClaimableBalance op
+   *    - the source account doesn't have a sequence number
+   *    - general XDR un/marshalling failures
```
marcelosalloum, line 237: 💡  I _believe_ you could use multiple `@throws` tags here, something like:
```js
/*
 * (...)
 * @throws {RangeError}  if opIndex is invalid
 * @throws {TypeError}   if the operation at the `opIndex` is not a CreateClaimableBalance op
 * (...)
/*
```

At least that works great in javadoc, so I believe it's the same for jsdoc.

This change is not required BTW, it's just a suggestion.

src/transaction.js
```diff
@@ -222,4 +226,57 @@ export class Transaction extends TransactionBase {
...
+        `expected createClaimableBalance, got ${op.type}: ${err}`
+      );
+    }
+
+    // Use the operation's source account or the transaction's source if not. In
+    // either case, use the unmuxed version.
+    let account = decodeAddressToMuxedAccount(this.source, true);
+    if (account.switch() === xdr.CryptoKeyType.keyTypeMuxedEd25519()) {
+      account = account.med25519();
+    }
```
marcelosalloum, line 268: 👀  I don't see you trying to use the op source account here, it seems you always get the transaction source. Is this comment accurate? Shouldn't it be referring to the muxed account G... address?

test/unit/operation_test.js
```diff
@@ -2478,6 +2478,20 @@ describe('Operation', function() {
         StellarBase.Operation.setTrustLineFlags({});
       }).to.throw();
     });
+    it('fails with a muxed account trustor', function() {
+      expect(() => {
+        const mAddress =
+          'MA7QYNF7SOWQ3GLR2BGMZEHXAVIRZA4KVWLTJJFC7MGXUA74P7UJUAAAAAAAAAAAAGZFQ';
+        console.log(mAddress);
```
marcelosalloum, line 2485: 👀  I think you can remove this log.

test/unit/operation_test.js
```diff
@@ -2478,6 +2478,20 @@ describe('Operation', function() {
...
+        StellarBase.Operation.setTrustLineFlags({
+          trustor: mAddress,
+          asset: StellarBase.Asset.native(),
+          flags: {
+            clawbackEnabled: false
+          }
+        });
+      }).to.throw();
+    });
```
marcelosalloum, line 2494: 🤔  this test doesn't make sense to me. Do you need to test a trustline to the native asset? Won't it always fail due to it being the native asset?

Which kind of error are you expecting to throw here?

marcelosalloum (commented):
I've marked 2 places to be fixed 👀 , a few suggestions 💡 and a question 🤔 .

src/transaction.js
Shaptic: Great call, I love it!

src/transaction.js
```diff
@@ -222,4 +226,57 @@ export class Transaction extends TransactionBase {
...
+   * This helper throws under the following circumstances:
+   *    - invalid `opIndex` value
+   *    - the operation at the `opIndex` is not a CreateClaimableBalance op
+   *    - the source account doesn't have a sequence number
+   *    - general XDR un/marshalling failures
+   *
+   * @param   {integer}  opIndex   the index of the CreateClaimableBalance op
+   * @returns {string}   a hex string representing the claimable balance ID
+   *
+   * @see https://github.com/stellar/go/blob/master/txnbuild/transaction.go#L392
```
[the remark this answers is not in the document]
Shaptic: nice catch

test/unit/operation_test.js
Shaptic: Nice, thanks!

src/transaction.js
Shaptic: Good catch, the comment is outdated! We should always use the G-address.

test/unit/operation_test.js
Shaptic: This change shouldn't even be here; reverted!

marcelosalloum (approved):
LGTM!

PR #483, opened 2022-01-05
Expose muxed account validators to Typescript

### LINKED ISSUE #472: `StrKey.isValidMed25519PublicKey` is not exposed in `types/index.d.ts`
**Describe the bug**

`StrKey.isValidMed25519PublicKey` is not exposed in `types/index.d.ts`, see the [function definition](https://github.com/stellar/js-stellar-base/blob/master/src/strkey.js#L106) and [types declaration](https://github.com/stellar/js-stellar-base/blob/master/types/index.d.ts#L843).

**What version are you on?**

6.0.3

Shaptic:
The following muxed account validation functions are now available from Typescript:

```typescript
namespace StrKey {
  function encodeMed25519PublicKey(data: Buffer): string;
  function decodeMed25519PublicKey(data: string): Buffer;
  function isValidMed25519PublicKey(publicKey: string): boolean;
}

function decodeAddressToMuxedAccount(address: string, supportMuxing: boolean): xdr.MuxedAccount;
function encodeMuxedAccountToAddress(account: xdr.MuxedAccount, supportMuxing: boolean): string;
function encodeMuxedAccount(gAddress: string, id: string): xdr.MuxedAccount;
```

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -846,6 +846,10 @@ export namespace StrKey {
   function decodeEd25519SecretSeed(data: string): Buffer;
   function isValidEd25519SecretSeed(seed: string): boolean;
 
+  function encodeMed25519PublicKey(data: Buffer): string;
+  function decodeMed25519PublicKey(data: string): Buffer;
+  function isValidMed25519PublicKey(publicKey: string): boolean;
+
   function encodePreAuthTx(data: Buffer): string;
   function decodePreAuthTx(data: string): Buffer;
 
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -944,4 +944,7 @@ export function verify(
   rawPublicKey: Buffer
 ): boolean;
 
+export function decodeAddressToMuxedAccount(address: string, supportMuxing: boolean): xdr.MuxedAccount;
+export function encodeMuxedAccountToAddress(account: xdr.MuxedAccount, supportMuxing: boolean): string;
+export function encodeMuxedAccount(gAddress: string, id: string): xdr.MuxedAccount;
 export function extractBaseAddress(address: string): string;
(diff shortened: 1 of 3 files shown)


sreuland (approved):
looks ok, I don't know typescript mechanics, probably not great review from me, can types/test.ts verify the interface change, didn't see ref to other methods from this interface in there, perhaps not.

Shaptic:
Good call @sreuland; I've added linting validation to `test.ts`.

types/test.ts
```diff
@@ -311,3 +311,23 @@ const lpWithdraw = StellarSdk.xdr.LiquidityPoolWithdrawOp.fromXDR(
   'base64'
 );
 lpWithdraw; // $ExpectType LiquidityPoolWithdrawOp
+
+const pubkey = masterKey.rawPublicKey(); // $ExpectType Buffer
+const seckey = masterKey.rawSecretKey(); // $ExpectType Buffer
+const muxed = StellarSdk.encodeMuxedAccount(masterKey.publicKey(), '1'); // $ExpectType MuxedAccount
+const muxkey = muxed.toXDR("raw"); // $ExpectType Buffer
+
+let result = StellarSdk.StrKey.encodeEd25519PublicKey(pubkey);  // $ExpectType string
```
paulbellamy, line 320: Not sure I really get the point of `ExpectType`.. The type signature is `(Buffer) -> string`, so if it didn't return that it wouldn't compile. but /shrug, doesn't hurt anything.

types/index.d.ts
```diff
@@ -849,6 +849,10 @@ export namespace StrKey {
   function decodeEd25519SecretSeed(data: string): Buffer;
   function isValidEd25519SecretSeed(seed: string): boolean;
 
+  function encodeMed25519PublicKey(data: Buffer): string;
```
paulbellamy, line 852: Yum, Medical25519 accounts. 💊

paulbellamy (commented):
Aren't we in the process of removing any explicit muxed-specific function signatures, so that muxed-account support is the default?

Shaptic:
@paulbellamy yes, but people still need to be able to distinguish between / validate M-addresses vs. G-addresses.

PR #484, opened 2022-01-05
Adds `Buffer` as type option for public/secret key parameter in the `Keypair` constructor

### LINKED ISSUE #456: Incorrect Type Defintion For KeyPair
The type definition for the KeyPair class constructor expects a string, where as the constructor in the KeyPair source file expects a buffer. 

Type Defintion:
`types/index.d.ts line 88`

Constructor:
`src/keypair.js line 26`

Version
`5.3.2`

1. Create a TS class that extends from `StellarSDK.Keypair`
2. Call `super` in the constructor of the child class with a Buffer as the argument for `secretKey`
3. TS lint throws an error for invalid argument type

The type definition should allow a `Buffer` as an argument for the secret key.

Shaptic:
Closes #456.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,11 @@ There are several other breaking changes:
   let mux2 = new MuxedAccount(mux1.baseAccount(), '2');
 ```
 
+### Fix
+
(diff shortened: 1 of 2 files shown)


marcelosalloum (approved):
LGTM!

PR #485, opened 2022-01-06
Turn on support for muxed addresses by default.

Shaptic:
see stellar/go#4164 for context


marcelosalloum (approved):
I've added a few suggestions and highlighted one breaking change❗about the base fee.

Please feel free to merge when you think it's ready.

src/transaction_builder.js
```diff
@@ -112,11 +112,13 @@ export class TransactionBuilder {
     this.source = sourceAccount;
     this.operations = [];
 
-    this.baseFee = isUndefined(opts.fee) ? BASE_FEE : opts.fee;
+    this.baseFee = opts.fee;
```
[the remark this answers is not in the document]
Shaptic: Heh, it's actually not :relieved: if you look above, the constructor will throw if `opts.fee` is undefined. That means the fee has always been required, and that this ternary condition has been never triggered:

```js
    if (isUndefined(opts.fee)) {
      throw new Error('must specify fee for the transaction (in stroops)');
    }
```

src/transaction_builder.js
```diff
@@ -112,11 +112,13 @@ export class TransactionBuilder {
...
     this.timebounds = clone(opts.timebounds) || null;
     this.memo = opts.memo || Memo.none();
     this.networkPassphrase = opts.networkPassphrase || null;
-    this.supportMuxedAccounts = opts.withMuxing || false;
+    this.supportMuxedAccounts = isUndefined(opts.withMuxing)
+      ? true
+      : opts.withMuxing;
```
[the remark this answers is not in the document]
Shaptic: Wow awesome, didn't know about `??`. Took me a sec to find docs; it's the ["nullish coalescing operator"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) :joy:

test/unit/operation_test.js
```diff
@@ -430,36 +414,43 @@ describe('Operation', function() {
...
+        'GBBM6BKZPEHWYO3E3YKREDPQXMS4VK35YLNU7NFBRI26RAN7GI5POFBB'
+      )
+    ];
 
-    it('does not support muxed accounts by default', function() {
+    it('supports muxed accounts by default', function() {
       expect(() => {
         StellarBase.Operation.pathPaymentStrictSend(opts);
-      }).to.throw(/destination is invalid/);
+      }).to.not.throw();
```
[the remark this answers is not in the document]
Shaptic: We're going from throw -> not throw, though. I want to be sure that this *never* throws, regardless of what the underlying error might be question. Is it okay in that case?

src/transaction_builder.js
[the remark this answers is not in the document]
Shaptic: Since we're dropping opt-out, this isn't necessary!

Shaptic:
@marcelosalloum new requirements rolled in: We are not going to allow "opting out" of muxing support. Basically, the default for all of the `withMuxing`, `supportMuxing`, etc. is now `true` and there's no way to pass `false`. The latest set of pushes will do this. I've resolved your comments that talk about the flags since they're just gone now. 

Please take another look!

--------

There is ONE case in which the above (no opt-out) will not hold:

```js
function encodeMuxedAccountToAddress(muxedAccount, forceEd25519) {}
```

The purpose of leaving it here is because sometimes you only want to extract the underlying `G...` address when you have an `xdr.MuxedAccount` option. Most notably, this is done by the `MuxedAccount.fromAddress` helper function. I was considering moving this specific functionality to a separate helper (e.g. `encodeMuxedAccountAsEd25519`) to separate it, but maybe that's excessive. Wdyt?

src/muxed_account.js
```diff
@@ -0,0 +1,155 @@
...
+  /**
+   * Returns a *new* muxed account with the same address but a new ID
+   *
+   * @param  {string} id    - a stringified uint64 representing the new ID
+   * @return {MuxedAccount} - a new account that shares the underlying base
+   *     account (incl. sequence number) but a different ID set
+   */
+  createSibling(id) {
+    return new MuxedAccount(this.account, id);
+  }
```
Shaptic, line 96: Thoughts on naming, @leighmcculloch? This is related to #433.

Shaptic: Rationale posted in https://github.com/stellar/js-stellar-base/issues/433#issuecomment-1009349938

src/operations/account_merge.js
```diff
@@ -9,22 +9,25 @@ import { decodeAddressToMuxedAccount } from '../util/decode_encode_muxed_account
...
+  opts.withMuxing = opts.withMuxing === undefined ? true : opts.withMuxing;
+
   const opAttributes = {};
   try {
     opAttributes.body = xdr.OperationBody.accountMerge(
       decodeAddressToMuxedAccount(opts.destination, opts.withMuxing)
     );
   } catch (e) {
-    throw new Error('destination is invalid');
+    throw new Error('destination is invalid; did you disable muxing?');
```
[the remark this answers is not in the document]
Shaptic: Resolving as the flag is no longer relevant.

src/muxed_account.js
leighmcculloch: I think we should remove the createSibling function and reconsider it separately if we want to re-add it. It introduces a new concept, sibling, that is one more thing to learn. It is relatively easy to create a muxed account by getting its base account and constructing a new one. Conceptually this seems sounds so I don't think we need createSibling or a function that does what it does. Wdyt?

src/muxed_account.js
```diff
@@ -0,0 +1,155 @@
+import isString from 'lodash/isString';
```
leighmcculloch, line 1: Could you keep this refactor, the separation into a different file, for another PR that is focused only on the refactor? As a reader it is difficult to see what has changed and what hasn't and to understand what is functional and what is refactor.

src/operation.js
```diff
@@ -114,20 +111,17 @@ export class Operation {
    * was used to create the operation (i.e. the `opts` parameter to most ops).
    *
    * @param {xdr.Operation}   operation - An XDR Operation.
-   * @param {boolean}         [withMuxing] - Indicates that if the operation
-   *     contains M... addresses, they should be interpreted fully as muxed
-   *     accounts. By default, this option is disabled until muxed accounts are
-   *     mature.
+   * @param {bool}    [=true] - Indicates any properties in this
```
leighmcculloch, line 114: Why do we keep this boolean property even though it is being removed?

src/operations/clawback.js
```diff
@@ -8,25 +8,30 @@ import { decodeAddressToMuxedAccount } from '../util/decode_encode_muxed_account
...
     throw new TypeError(this.constructAmountRequirementsError('amount'));
   }
   attributes.amount = this._toXDRAmount(opts.amount);
   attributes.asset = opts.asset.toXDRObject();
-  attributes.from = decodeAddressToMuxedAccount(opts.from);
+  try {
+    attributes.from = decodeAddressToMuxedAccount(opts.from);
+  } catch (e) {
+    throw new Error('from address is invalid');
+  }
```
leighmcculloch, line 34: I'm not following how these changes related to enabling muxed addresses by default. Are these changes required for enabling muxed addresses by default, or some other bug being fixed?

src/transaction_builder.js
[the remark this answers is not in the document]
leighmcculloch: This doesn't look related to turning on muxed addresses by default though. If this is an unrelated bug/improvement, could it be moved to its own PR to keep this PR hyper-focused on muxed address enabling?

src/util/decode_encode_muxed_account.js
```diff
@@ -35,27 +29,30 @@ export function decodeAddressToMuxedAccount(address, supportMuxing) {
...
- *     muxing ID, but *ONLY* when the ID is present.
+ * @param   {xdr.MuxedAccount} muxedAccount   Raw account to stringify
+ * @param   {bool} [forceEd25519=false]       If this is set to `true`, this
+ *     will always return the underlying G... address of the `muxedAccount`.
  *
  * @returns {string}  stringified G... (corresponding to the underlying pubkey)
  *     or M... address (corresponding to both the key and the muxed ID)
  */
-export function encodeMuxedAccountToAddress(muxedAccount, supportMuxing) {
+export function encodeMuxedAccountToAddress(muxedAccount, forceEd25519) {
```
leighmcculloch, line 48: Do we need to keep the forceEd25519? This function is named `encodedMuxedAccountToAddress`, and the only valid address for a muxed account is a M address. If someone wants to encode the base account of a muxed accounmt to a G address, they should probably call mux.baseAccount() and go from there?

Is this required for compatibility?

leighmcculloch (commented):
Really excited to see this happening, and I like that we're going with the unconditional support route.

I have some questions and concerns inline. In a few places it feels like this PR is doing multiple things such as refactoring areas of the code and fixing other unrelated issues/cleanup. It is possible they are all related and I don't see the connections. Could this PR be hyperfocused on the functional changes required for muxed address by default?

src/operation.js
Shaptic: Ctrl+F mistake :smile:

src/operations/clawback.js
Shaptic: This was done to line up how clawback assigns the muxed address with all of the other mux-enabled ops (like payment).

src/util/decode_encode_muxed_account.js
Shaptic: It's a separation of concerns: these functions are for operating between `xdr.MuxedAccount` <--> `string`, and then there's also the higher-level `MuxedAccount` object that implements the `Account` interface.

I touched on this a bit above (https://github.com/stellar/js-stellar-base/pull/485#issuecomment-1009347554), but basically we need to be able to forceEd25519 when actually building a `MuxedAccount` object from an M-address in order to get its composite G-address and ID.

src/muxed_account.js
Shaptic: Agreed, done in e9610c077707699e62d6f323ba9063adbc296ef5.

src/transaction_builder.js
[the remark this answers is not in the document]
Shaptic: Yep, you're right. Reverted in 2bbc5109dad913d0b7930b7f5368369fdedecf42.

src/muxed_account.js
Shaptic: Yep, done in 9fe601a5024ba47dd3eed8435bbe04ec469fc64c and 2f7da085bf49f2b2ebe682b1217f56dc543c2c40.

src/operation.js
Shaptic: Fixed in 2f7da085bf49f2b2ebe682b1217f56dc543c2c40.

Shaptic:
Thanks for your review, @leighmcculloch! The latest push should hyper-focus on the new muxing defaults delta.

src/util/decode_encode_muxed_account.js
leighmcculloch: Ah I see. I still think we should follow a deconstruct and a construct approach, rather than a parameter that makes a construction behave differently. By deconstruct I mean get the base address/account id out of the xdr.MuxedAccount. We should probably have a function on MuxedAccount to do this if we don't already.

Shaptic: I wrote up an overly-verbose comment on the design decisions I was making and all of the abstraction layers we need to deal with (the TL;DR is that we have `xdr.MuxedAccount` :left_right_arrow: M-address, `xdr.MuxedAccount` :left_right_arrow: G-address, and `MuxedAccount` :left_right_arrow: M-address), but I think the latest commit explains all that is needed: 10f14a377b049c1e6d94617cea4e8df16f2d325b.

I've dropped the `forceEd25519` flag and moved that subset of functionality to a new helper,
```ts
function extractBaseAddress(mAddress: string): string;
```

Does that help?

test/unit/operation_test.js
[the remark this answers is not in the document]
marcelosalloum: Yeah, my mistake here. Thanks for taking the time.

src/util/decode_encode_muxed_account.js
```diff
@@ -4,28 +4,17 @@ import xdr from '../generated/stellar-xdr_generated';
...
- * Note that this behaviour will eventually be the default.
- *
- * @function
- *
- * @param   {string}  address         G... or M... address to encode into XDR
- * @param   {bool}   [supportMuxing]  allows decoding of the muxed
- *     representation of the address, extracting the underlying ID from the M...
- *     address
+ * This supports full muxed accounts by default, where an M... address will
+ * resolve to both its underlying G... address and an ID.
```
leighmcculloch, line 11: 💡 Since it isn't configurable it is maybe misleading to say it is the default as if it is configurable.
```suggestion
 * This supports full muxed accounts, where an M... address will
 * resolve to both its underlying G... address and an ID.
```

src/util/decode_encode_muxed_account.js
```diff
@@ -86,6 +70,20 @@ export function encodeMuxedAccount(address, id) {
...
 
+/**
+ * Extracts the underlying base (G...) address from an M-address.
+ * @param  {string} address  a muxed account address (M...)
+ * @return {string} a Stellar public key address (G...)
+ */
+export function extractBaseAddress(address) {
+  if (!StrKey.isValidMed25519PublicKey(address)) {
+    throw new TypeError('address should be a muxed account (M...)');
+  }
```
leighmcculloch, line 81: ❔ Is there a way to test if an address is a muxed address before passing it to this function? I'm wondering about the flow where someone has a value that might be an M or a G and they want to get the G. If they call this function on its own it might error in the G cases. How do they distinguish? Do they just check the first character? Or maybe this function should just return the G if the input is a G. I don't feel strongly what way we go, but curious what the story is for this use case.

leighmcculloch (approved):
Looks great 👏🏻.

One suggested doc change (💡), very minor. And one question about how a new function will be used in one scenario (❔).

src/util/decode_encode_muxed_account.js
leighmcculloch: Haha, I realize the line I'm commenting on literally has the function for testing if an address is a muxed address.

leighmcculloch: I still wonder if there is value in not erroring in the case a G is passed and just returning the G. It might make the API a little easier to use, albeit with no errors in a case that might be error worthy.

Shaptic: That's an interesting question. I can't foresee when it would come up, but I don't have a strong opinion either way.

Callers can use the `StrKey` validators (`isValidEd25519PublicKey` and `isValidMed25519PublicKey`, just like is done here) before calling the function. I added a check in 9e1ee1f to just pass G-addresses through, for convenience.

src/util/decode_encode_muxed_account.js
Shaptic: Nice, done in 9e1ee1f!

src/util/decode_encode_muxed_account.js
Shaptic: Hah, should have refreshed before replying :ninja:. But yeah, added the pass-through.

PR #487, opened 2022-01-10
Move `MuxedAccount` to a separate file and drop the `createSubaccount` method.

### LINKED ISSUE #433: Rename createSubaccount to createMuxedAccount

Shaptic:
This closes #433 by removing the `createSubaccount` method entirely from `Account`, `MuxedAccount`, and their shared interface. This allows the two objects to be in separate files.

See #433 for discussion.


CHANGELOG.md
```diff
@@ -3,6 +3,20 @@
 
 ## Unreleased
 
+### Breaking Change
+
+- Remove the `createSubaccount` method from `Account` and `MuxedAccount` ([#487](https://github.com/stellar/js-stellar-base/pull/487)). You should prefer to do this manually:
```
leighmcculloch, line 8: When I suggested breaking it out I had imagined this part, a functional change, would stay in the other PR, but either way this was easy to review. 👏🏻

leighmcculloch (approved):

CHANGELOG.md
Shaptic: In retrospect that would've been cleaner, but thanks for the review! :heavy_check_mark:

Issue #404, opened 2021-04-01
Compile issues due to sodium-native


electic:
https://github.com/stellar/js-stellar-base/blob/c687cf941133814a40d3e49a13794900453bfed8/package.json#L115

Sodium-native's version is causing issues because the node-gyp version is old. That means it does not have support for VS 2019 causing compiler compiler errors because the enum value is not in node-gyp for python look ups. Can we bump the version? 

From the code, it seems it is only used for singing. Do we need this complexity?

gre:
This is indeed a problem that we also felt on our side (Ledger Live) on our path of migrating from Node 12 to Node 14. Is there any progress done on the matter? Is there a possible workaround to disable sodium-native globally? (it seems yarn only allow to globally disable optional, but you can't target a specific one)

gre:
Please consider upgrading or changing sodium to something else. 🙏

Shaptic:
This will be a long reply but I think worthwhile both for current and future context, particularly for myself as a newer maintainer. 

**TL;DR**: There are no suitable, secure alternatives to the library from my investigation. However, try the [`upgrade-nacl`](https://github.com/stellar/js-stellar-base/tree/upgrade-nacl) branch to see if your build errors are resolved. It bumps `sodium-native` to 2.4.9 and `node-gyp-build` to 4.3.0, which is its latest version. 

### Prior work
There have been discussions and changesets before on the crypto dependencies:

 - stellar/js-stellar-sdk#242
 - #339
 - #171

Historically, it appears we went from: [`ed25519`](https://github.com/dazoe/ed25519) to [`sodium-native`]() after considering `elliptic` and `crypto`, opting for the former for a number of reasons.

### On `sodium-native`
It's first worth driving home the fact that, as the README [states](https://github.com/stellar/js-stellar-base/#js-stellar-base), this is an *optional* dependency used for performance, so nobody should be blocked by not addressing this request. That should address this, from above:

> Is there a possible workaround to disable sodium-native globally?

#### Performance concerns
Obviously, performance is important, so we'd prefer a replacement that is performant and also alleviates people's build issues.

[...]

Issue #339, opened 2020-04-27
Consider removal of `sodium-native` optional dependency, leaving only `tweetnacl`


grempe:
**Is your feature request related to a problem? Please describe.**

`sodium-native`, while significantly faster than `tweetnacl`, introduces complexity in the form of:

- It requires native code compilation tools to be present in the runtime environment
- It causes significant complexity when trying to use within a Docker environment
- It confuses developers who need to determine if their use-case needs the additional speed.
- It may (?) be solving a performance problem that doesn't exist in the real world?
- `sodium-native` has not been put under a security audit to my knowledge (`tweetnacl` has been formally audited by Cure53 and is FAR less complex code-wise).

**Describe the solution you'd like**

No additional install-time or compilation time dependencies or decision making.

I've created and run a simple benchmark suite to satisfy my own curiosity as to the speed differences between these two libraries. There is no contest for simple signing operations. `sodium-native` is far faster. Here is the benchmark repo and the output:

https://github.com/truestamp/ed25519-bench

```
❯ node benchmark.js
Running "tweetnacl vs. sodium-native" suite...
Progress: 100%

  sign w/ tweetnacl:
    286 ops/s, ±2.24%      | slowest, 99.47% slower

  sign w/ sodium-native:
    54 008 ops/s, ±0.62%   | fastest

Finished 2 cases!
  Fastest: sign w/ sodium-native
  Slowest: sign w/ tweetnacl

Saved to: benchmark/results/sign.json

Saved to: benchmark/results/sign.chart.html
```

TL;DR : `54,008` vs. `286` operations per second. Crushed it!

However, I am left wondering if there is ever any use case for using `js-stellar-base` where this difference is meaningful? Would any use case of this library **ever** require thousands of signing operations per second vs. hundreds of operations/second? Is 3.49ms/signature ever too slow for normal use?

If so, and I am just unaware of these high performance use-cases, then by all means `sodium-native` should stay as an optional dependency. If not, then I feel it should be removed as it adds needless complexity and it introduces additional operational and security overhead and code complexity. This may fall under the category of pre-mature optimization.

**Describe alternatives you've considered**

The alternatives are already baked into this library as primary and optional dependencies.

**Additional context**

cc: @abuiles 

Related issue:

https://github.com/stellar/js-stellar-sdk/issues/534

s-a-y:
+1 to make sodium-native optional
[...]

PR #495, opened 2022-01-26
Remove sodium-native optional dependency

grempe:
This commit addresses the following issues:

https://github.com/stellar/js-stellar-base/issues/339
https://github.com/stellar/js-stellar-base/issues/404

Changes:

- removal of optional sodium-native native compiled module
- promotion of existing version of
tweetnacl-js to handle all sign/verify duties

Removal of the optional dependency greatly simplifies the
code in `src/signing.js` and removes all native compilation
issues that have negatively impacted developers for at least
two years when using modern versions of NodeJS.

This commit does not choose to prefer a new method of signing,
it simply delegates that task to the existing primary signature
library (tweetnacl-js) in all cases. This also has the pleasant
side-effect of greatly simplifying the signature code removing
what had been described in the code comments as
being "a little strange".

The actual signature generate/sign/verify functions remain
completely unchanged from prior code and have been refactored
only to simplify the code.  This also has the pleasant side
effect of allowing any security audits of this code, or the
associated tweetnacl-js library, to have far less surface
area to examine.

Cryptographic 'agility', as previously existed here to address
theoretical performance issues, is considered a security anti-pattern.

All existing gulp test suites pass when tested with Node version 14.18.2

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,14 +12,6 @@ implementation in JavaScript that can be used on either Node.js or web browsers.
 
 - **[API Reference](https://stellar.github.io/js-stellar-base/)**
 
-> **Warning!** Node version of this package is using [`sodium-native`](https://www.npmjs.com/package/sodium-native) package, a native implementation of [Ed25519](https://ed25519.cr.yp.to/) in Node.js, as an [optional dependency](https://docs.npmjs.com/files/package.json#optionaldependencies).
-> This means that if for any reason installation of this package fails, `stellar-base` will fallback to the much slower implementation contained in [`tweetnacl`](https://www.npmjs.com/package/tweetnacl).
->
-> If you are using `stellar-base` in a browser you can ignore this. However, for production backend deployments you should definitely be using `sodium-native`.
-> If `sodium-native` is successfully installed and working
-> `StellarBase.FastSigning` variable will be equal `true`. Otherwise it will be
-> `false`.
-
 ## Quick start
 
 Using yarn to include js-stellar-base in your own project:
--- a/package.json
+++ b/package.json
@@ -111,9 +111,6 @@
     "sha.js": "^2.3.6",
     "tweetnacl": "^1.0.0"
   },
-  "optionalDependencies": {
-    "sodium-native": "^2.3.0"
-  },
   "resolutions": {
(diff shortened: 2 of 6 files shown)


Shaptic (commented):
Thanks for taking the initiative on this; I really appreciate it. It resurfaced the pain point and I tried to investigate its impact to our organization a little more. An important factoid came up: we actually do have an important use case involving low-end hardware. 

As noted in [your own benchmarks](https://github.com/stellar/js-stellar-base/issues/339#issuecomment-879512062), removing sodium-native would cause a **significant** slowdown. It would balloon signing times on the order of **seconds** on certain low-end hardware we support.

As such, I can't approve this PR. We need another strategy for working around sodium-native, though I'm still not sure why that workaround can't just involve, well, not installing it :yum:

grempe:
Can you please share with us what the exact low end hardware platforms you support that are of concern? I can find no information that indicates minimum hardware specs for this lib. This is a question that has gone unanswered for two years.

Can you please also share a simple example that shows signing operations would take "seconds" to perform on that low spec platform using tweetnacl (the current default) as claimed? That would be literally thousands of times slower than observed performance on other low end platforms including Raspberry Pi. Do you support some hardware of lower spec than RPi?

To answer your question about just skipping installation I think there is a lot of history in these discussions that indicate that this is not an easy viable option using the current dependency setup (e.g docker builds where there is no compilation toolchain fail, also many edge compute serverless platforms will also fail).

If the removal of this optional dependency via pull request will not be approved, would an alternative build/install script for the optional sodium based installation be a possible route? However it is done, node-gyp it would seem needs to be completely removed from the normal 'npm install' path to solve this issue which is still apparently very much alive (and getting worse as node-gyp and sodium fall further behind).

E.g.

https://github.com/nodejs/node-gyp/issues/1177

Would this type of approach be approved? Where you perhaps need to run a different install command if you want native compiled support?

We can't currently help out to try and resolve the situation since we can't get anyone to commit to any information on supported hardware minimums or signing use cases that demand the absolute max performance. So there is no target we can hit for a solution to help improve the developer experience.

Shaptic:
Solved by #802

PR #497, opened 2022-02-01
Sodium native fix Diff

alejomendoza:


Shaptic:
This will be superceded by #498, which will upgrade sodium-native on the mainline.

PR #509, opened 2022-03-31
Update XDR definitions and relevant code for Protocol 19

Shaptic:
This is a **bare-minimum** set of code updates to work with the Protocol 19 XDR. That means there is no new functionality introduced, particularly around the new transaction preconditions in [CAP-21](https://stellar.org/protocol/cap-21). A robust refactor to this will come in a separate PR.

The purpose of this PR is to have a shared, working ground to work on the disjoint tasks outlined in stellar/js-stellar-sdk#763.

### DIFF
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -72,13 +72,41 @@ export class Transaction extends TransactionBase {
         break;
     }
 
-    const timeBounds = tx.timeBounds();
+    let timeBounds = null;
+    switch (this._envelopeType) {
+      case xdr.EnvelopeType.envelopeTypeTxV0():
+        timeBounds = tx.timeBounds();
+        break;
+
+      case xdr.EnvelopeType.envelopeTypeTx():
+        switch (tx.cond().switch()) {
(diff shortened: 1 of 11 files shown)


paulbellamy (approved):

src/transaction.js
```diff
@@ -72,13 +72,41 @@ export class Transaction extends TransactionBase {
...
+              .timeBounds();
+            break;
+
+          default:
+            break;
+        }
+        break;
+
+      default:
+        break;
```
leighmcculloch, line 100: ‼️I think this default case should throw an error. If we get into the default case it means we either have an invalid tx, or a tx we do not completely understand. A tx we do not completely understand may be particularly dangerous because we may completely misinterpret what the transaction means. (This is one reason why we use a format like XDR where we can explicitly pick and choose what we are okay with being unknown.)

leighmcculloch (commented):
@Shaptic One thing that we should address (‼️) below.

PR #510, opened 2022-03-31
Bump minimum Node version to 14.x.

Shaptic:
12.x will be EOL [in April](https://nodejs.org/en/about/releases/), so this is a good time to move to the next LTS release.

**Reviewers:** please read the commit message in b9c261dd0b30263fbe1963ced33d9a706848639f just for the sake of spreading obscure, latent tribal knowledge.


paulbellamy:
Can we bump .nvmrc in js-stellar-sdk as well?

paulbellamy (approved):

PR #511, opened 2022-03-31
Add support for signed payload signer keys to `StrKey`

### LINKED ISSUE #507: Support signed payloads (CAP-40) in StrKey implementations
This is the JavaScript-specific issue for stellar/go#4262.

Shaptic:
This adds support for encode signed payload signers ([CAP-40](https://stellar.org/protocol/cap-40)) to their StrKey representation ([SEP-23](https://github.com/stellar/stellar-protocol/pull/1014)):

```
+--------------------+
|   signer address   |
|                    | <------------> "PA7QYNF7SOWQ3GL..."
|      payload       |
+--------------------+
```

Closes #507.

### DIFF
--- a/src/keypair.js
+++ b/src/keypair.js
@@ -1,3 +1,5 @@
+/* eslint no-bitwise: ["error", {"allow": ["^"]}] */
+
 import nacl from 'tweetnacl';
 import isUndefined from 'lodash/isUndefined';
 import isString from 'lodash/isString';
@@ -235,5 +237,31 @@ export class Keypair {
     const hint = this.signatureHint();
 
     return new xdr.DecoratedSignature({ hint, signature });
+  }
+
+  /**
+   * Returns the signature hint for a signed payload signer.
+   *  This is defined as the last 4 bytes of the signer key XORed with last 4
+   *  bytes of the payload (zero-left-padded if necessary).
+   *
+   * @param  {Buffer} data    data to both sign and treat as the payload
+   * @return {xdr.DecoratedSignature}
+   *
+   * @see https://github.com/stellar/stellar-protocol/blob/master/core/cap-0040.md#signature-hint
+   */
+  signPayloadDecorated(data) {
+    const signature = this.sign(data);
+    const keyHint = this.signatureHint();
+
+    let hint = Buffer.from(data.slice(-4));
+    if (hint.length < 4) {
+      // front-pad with zeroes if necessary
+      hint = Buffer.concat([Buffer.alloc(4 - data.length, 0), hint]);
+    }
+
+    return new xdr.DecoratedSignature({
+      hint: hint.map((byte, i) => byte ^ keyHint[i]),
+      signature
+    });
   }
 }
--- a/src/strkey.js
+++ b/src/strkey.js
@@ -1,4 +1,4 @@
-/* eslint-disable no-bitwise */
+/* eslint no-bitwise: ["error", {"allow": ["<<"]}] */
 
 import base32 from 'base32.js';
 import crc from 'crc';
--- a/src/strkey.js
+++ b/src/strkey.js
@@ -233,10 +233,12 @@ function isValid(versionByteName, encoded) {
       return decoded.length === 32;
 
     case 'med25519PublicKey':
-      return decoded.length === 40;
+      return decoded.length === 40; // +8 bytes for the ID
 
     case 'signedPayload':
-      return decoded.length >= 32 + 4 && decoded.length <= 32 + 4 + 64;
+      return (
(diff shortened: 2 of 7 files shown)


src/strkey.js
```diff
@@ -136,31 +137,110 @@ export class StrKey {
...
+    case 'ed25519SecretSeed': // falls through
+    case 'preAuthTx': // falls through
+    case 'sha256Hash':
+      return decoded.length === 32;
+
+    case 'med25519PublicKey':
+      return decoded.length === 40;
+
+    case 'signedPayload':
+      return decoded.length >= 32 + 4 && decoded.length <= 32 + 4 + 64;
```
piyalbasu, line 239: My guess is that `4` is related to the length of the hint. Maybe a comment would be helpful to make this seem a little less magic number-y

piyalbasu (approved):

src/strkey.js
Shaptic: Yep, good call!

src/strkey.js
```diff
@@ -64,11 +65,11 @@ export class StrKey {
...
   /**
    * Decodes strkey ed25519 seed to raw data.
-   * @param {string} data data to decode
+   * @param {string} address data to decode
    * @returns {Buffer}
    */
-  static decodeEd25519SecretSeed(data) {
-    return decodeCheck('ed25519SecretSeed', data);
+  static decodeEd25519SecretSeed(address) {
+    return decodeCheck('ed25519SecretSeed', address);
```
leighmcculloch, line 72: 💡 Fwiw the secret seed strkey is not an address, it's a seed encoded in a strkey. Using the term `encoded` instead of `data` may be more accurate if you're wanting to get away from the generic `data`.

src/strkey.js
```diff
@@ -118,11 +119,11 @@ export class StrKey {
...
   /**
    * Decodes strkey PreAuthTx to raw data.
-   * @param {string} data data to decode
+   * @param {string} address data to decode
    * @returns {Buffer}
    */
-  static decodePreAuthTx(data) {
-    return decodeCheck('preAuthTx', data);
+  static decodePreAuthTx(address) {
+    return decodeCheck('preAuthTx', address);
```
leighmcculloch, line 126: 💡 Same here, this is also not an address, it's a tx hash encoded in a strkey.

src/strkey.js
```diff
@@ -136,31 +137,112 @@ export class StrKey {
...
   /**
    * Decodes strkey sha256 hash to raw data.
-   * @param {string} data data to decode
+   * @param {string} address data to decode
    * @returns {Buffer}
    */
-  static decodeSha256Hash(data) {
-    return decodeCheck('sha256Hash', data);
+  static decodeSha256Hash(address) {
+    return decodeCheck('sha256Hash', address);
```
leighmcculloch, line 144: 💡 Same here.

src/strkey.js
```diff
@@ -136,31 +137,112 @@ export class StrKey {
...
+    return encodeCheck('signedPayload', data);
+  }
+
+  /**
+   * Decodes strkey signed payload (P...) to raw data.
+   * @param   {string} address  address to decode
+   * @returns {Buffer}
+   */
+  static decodeSignedPayload(address) {
+    return decodeCheck('signedPayload', address);
```
leighmcculloch, line 162: 💡 Same here.

leighmcculloch (commented):
Looks like there's one issue in this PR (❗). I also left some less critical notes about the terminology we use for strkey encoded values.

src/keypair.js
```diff
@@ -236,4 +238,30 @@ export class Keypair {
...
+   * @see https://github.com/stellar/stellar-protocol/blob/master/core/cap-0040.md#signature-hint
+   */
+  signPayloadDecorated(data) {
+    const signature = this.sign(data);
+    const keyHint = this.signatureHint();
+
+    let hint = Buffer.from(data.slice(-4));
+    if (hint.length < 4) {
+      // front-pad with zeroes if necessary
+      hint = Buffer.concat([Buffer.alloc(4 - data.length, 0), hint]);
```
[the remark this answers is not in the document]
Shaptic: Yep, caught this later in #523. Wish GH had a better notification system! Def worth a ping in my case if you add comments post-merge :+1:

src/strkey.js
Shaptic: Yeah that's a pretty good point... I wanted to get away from the generic `data` to indicate the _type_. Calling it `encoded` doesn't help, since that's implied by `decodeEd25...`. I find myself checking the docstring every single time I use this set of helpers, because "am I encoding to strkey or am I decoding from strkey or am I decoding from binary or am I ..." etc.

leighmcculloch: Naming it `str` may help then.

PR #512, opened 2022-04-01
Adding builder methods for new preconditions.

### LINKED ISSUE #508: Support add'l preconditions (CAP-21) when building transactions
This is the JavaScript-specific issue for stellar/go#4281.

paulbellamy:
Fixes #508 

Note: this is pending on xdr updates to add the new fields to the txn envelope. I've made a guess at their location, but the txn constructor might need updating.


paulbellamy:
Should `setMinTime/setmaxTime`, take epoch milliseconds instead of epoch seconds? It would make `Date.now()`, and `new Date().valueOf()` type stuff more intuitive in js. But we'd be throwing away some precision. Is there a good example somewhere else we could follow for consistency?

PR #513, opened 2022-04-01
Adding builder methods for new preconditions.

paulbellamy:
Fixes https://github.com/stellar/js-stellar-base/issues/508

Replaced: https://github.com/stellar/js-stellar-base/pull/512

Note: this is pending on xdr updates to add the new fields to the txn envelope. I've made a guess at their location, but the txn constructor might need updating.

Should setMinTime/setmaxTime, take epoch milliseconds instead of epoch seconds? It would make Date.now(), and new Date().valueOf() type stuff more intuitive in js. But we'd be throwing away some precision. Is there a good example somewhere else we could follow for consistency?


src/transaction_builder.js
```diff
@@ -110,6 +124,11 @@ export class TransactionBuilder {
 
     this.baseFee = opts.fee;
     this.timebounds = clone(opts.timebounds) || null;
+    this.ledgerbounds = clone(opts.ledgerbounds) || null;
+    this.minAccountSequence = opts.minAccountSequence || null;
+    this.minAccountSequenceAge = opts.minAccountSequenceAge || null;
+    this.minAccountSequenceLedgerGap = opts.minAccountSequenceLedgerGap || null;
+    this.extraSigners = opts.extraSigners || null;
```
Shaptic, line 131: `clone()` equivalent? dunno

src/transaction_builder.js
```diff
@@ -153,33 +172,252 @@ export class TransactionBuilder {
...
+   * @see TransactionPreconditions
+   */
+  setMinTime(dateOrEpochSeconds) {
+    // Force it to a date type
+    if (typeof dateOrEpochSeconds === 'number') {
+      dateOrEpochSeconds = new Date(dateOrEpochSeconds * 1000);
+    }
+
+    // Convert that date to the epoch seconds
+    const t = Math.floor(dateOrEpochSeconds / 1000);
```
Shaptic, line 198: Erm is `Date` divided by `int` well-defined? :detective:

src/transaction_builder.js
```diff
@@ -153,33 +172,252 @@ export class TransactionBuilder {
...
+    }
+
+    if (t < 0) {
+      throw new Error('min_time cannot be negative');
+    }
+
+    if (this.timebounds === null) {
+      this.timebounds = {
+        minTime: 0,
+        maxTime: 0
```
Shaptic, line 212: maybe we use `maxTime: TIMEOUT_INFINITE` here to self-document? but also I think this is breaking behavior: previously it was impossible to create a tx w/o timebounds unless you explicitly called `setTimeout(INF)` or `setTimeBounds(minTime: x, maxTime: INF)`, but now setting `setMinTime(x)` gives you inf max time _implicitly_ :thinking: 

No strong opinion, but what're your thoughts?

src/transaction_builder.js
```diff
@@ -153,33 +172,252 @@ export class TransactionBuilder {
...
+    }
+
+    if (this.ledgerbounds === null) {
+      this.ledgerbounds = {
+        minLedger: 0,
+        maxLedger: 0
+      };
+    }
+
+    this.ledgerbounds.maxLedger = maxLedger;
```
Shaptic, line 320: Can / Should we check to `assert maxLedger === 0 || maxLedger >= minLedger`?

src/transaction_builder.js
```diff
@@ -153,33 +172,252 @@ export class TransactionBuilder {
...
+
+    return this;
+  }
+
+  /**
+   * If you want to prepare a transaction which will be valid only while the account sequence number is `minAccountSequence <= sourceAccountSequence < tx.seqNum`. Note that after execution the account's sequence number is always raised to `tx.seqNum`.
```
Shaptic, line 326: nit: can we wrap these all to 80 or 100 chars? :sweat_smile: surprised the linter didn't do it

src/transaction_builder.js
```diff
@@ -153,33 +172,252 @@ export class TransactionBuilder {
...
+   * be at least minAccountSequenceAge greater than sourceAccount's sequenceTime.
+   * Internally this will set the <code>minAccountSequenceAge</code> precondition.
+   *
+   * @param {number} durationInSeconds The minimum amount of time between source account sequence time and the ledger time when this transaction will become valid.
+   * If the value is `0`, the transaction is unrestricted by the account sequence age. Cannot be negative.
+   * @return {TransactionBuilder}
+   * @see TransactionPreconditions
+   */
+  setMinAccountSequenceAge(durationInSeconds) {
+    if (this.minAccountSequenceAge !== null && this.minAccountSequenceAge > 0) {
```
Shaptic, line 357: `>=`

src/transaction_builder.js
```diff
@@ -153,33 +172,252 @@ export class TransactionBuilder {
...
+   * If you want to prepare a transaction which will be valid before some ledger number in the future, you can set a maximum ledger precondition.
+   * Internally this will set the <code>maxLedger</code> precondition.
+   *
+   * @param {number} maxLedger The maximum ledger this transaction is valid before. Cannot be negative.
+   * If the value is `0`, the transaction is valid indefinitely.
+   * @return {TransactionBuilder}
+   * @see TransactionPreconditions
+   */
+  setMaxLedger(maxLedger) {
+    if (this.ledgerbounds !== null && this.ledgerbounds.maxLedger > 0) {
```
Shaptic, line 303: `>=`, since `0` is valid

src/transaction_builder.js
```diff
@@ -153,33 +172,252 @@ export class TransactionBuilder {
...
+   * If you want to prepare a transaction which will be valid at or after some ledger number in the future, you can set a minimum ledger precondition.
+   * Internally this will set the <code>minLedger</code> precondition.
+   *
+   * @param {number} minLedger The minimum ledger this transaction is valid at or after. Cannot be negative.
+   * If the value is `0`, the transaction is valid immediately.
+   * @return {TransactionBuilder}
+   * @see TransactionPreconditions
+   */
+  setMinLedger(minLedger) {
+    if (this.ledgerbounds !== null && this.ledgerbounds.minLedger > 0) {
```
Shaptic, line 271: `>=`

src/transaction_builder.js
```diff
@@ -153,33 +172,252 @@ export class TransactionBuilder {
...
+   * @see TransactionPreconditions
+   */
+  setExtraSigners(extraSigners) {
+    if (this.extraSigners !== null && this.extraSigners.length > 0) {
+      throw new Error(
+        'extra_signers has been already set - setting extra_signers would overwrite it.'
+      );
+    }
+
+    this.extraSigners = extraSigners;
```
Shaptic, line 419: Do we copy the array? Dunno

src/transaction_builder.js
```diff
@@ -227,14 +465,39 @@ export class TransactionBuilder {
...
 
+    attrs.preconditions = {};
+
     this.timebounds.minTime = UnsignedHyper.fromString(
       this.timebounds.minTime.toString()
     );
     this.timebounds.maxTime = UnsignedHyper.fromString(
       this.timebounds.maxTime.toString()
     );
+    attrs.preconditions.timeBounds = new xdr.TimeBounds(this.timebounds);
```
Shaptic, line 476: This will be clearer once you merge in the XDR changes, but the behavior we need is: if **only** timebounds are set, then set `attrs.cond = xdr.Preconditions.precondTime(this.timebounds)` (that code will already exist after rebase), otherwise, use `precondV2`

src/transaction_builder.js
Shaptic: Also can sanity check that `extraSigners.length <= 2` since more will blow up in the XDR encode step

Shaptic:
> Should setMinTime/setmaxTime, take epoch milliseconds instead of epoch seconds? It would make Date.now(), and new Date().valueOf() type stuff more intuitive in js. But we'd be throwing away some precision. Is there a good example somewhere else we could follow for consistency?

Good question... We use seconds in the other SDKs, yet it makes the intuition for SDK users clunkier here. If they haphazardly write `setMaxTime(Date.now() + 5000)` expecting to get 5 seconds from now, they will in reality have a timeout that's far, **far** larger. That's not very good, so maybe we should keep it intuitive?

src/transaction_builder.js
```diff
@@ -153,33 +172,252 @@ export class TransactionBuilder {
...
+  /**
+   * If you want to prepare a transaction which will become valid at some point in the future, you can set a minimum time precondition.
+   * Internally this will set the <code>minTime</code> precondition.
+   *
+   * @param {Date|number} dateOrEpochSeconds Either a js Date object, or a number of unix epoch seconds. Can't be negative.
+   * If the value is `0`, the transaction is valid immediately.
+   * @return {TransactionBuilder}
+   * @see TransactionPreconditions
+   */
+  setMinTime(dateOrEpochSeconds) {
```
Shaptic, line 191: Wait let's keep it simple and non-breaking: don't allow setting min or max time individually. Let's keep `setTimeout` and `setTimebounds` as before.

src/transaction_builder.js
paulbellamy: Yeah, was debating if we need one for the array. Probably wouldn't hurt

src/transaction_builder.js
paulbellamy: I don't really see why creating a txn without a timeout is invalid to begin with. Maybe you're not going to submit it right away?

Shaptic: We just discourage inf timeouts from being implicit behavior, so you must do `setTimeout(INF)`. I think the primary reason is because timebounds are the One True Way:tm: to get **definitive** success/failure state (see [Error Handling](https://developers.stellar.org/docs/tutorials/handling-errors/#retrying-until-success)), so we highly encourage people to set them no matter what.

src/transaction_builder.js
paulbellamy: `setTimebounds`?

src/transaction_builder.js
paulbellamy: afaik yes. but added a `.valueOf()` to be more explicit.

paulbellamy:
> expecting to get 5 seconds from now, they will in reality have a timeout that's far, **far** larger. 

Based on kibana logs, that is absolutely already happening to a lot of users.

src/transaction_builder.js
paulbellamy: You could never set the min time before, so.. It's breaking either way. Maybe keeping `setTimeout(max)` and adding `setTimebounds(min,max)` would be a cleaner api. 🤔

src/transaction_builder.js
paulbellamy: Why not always use a precondV2?

src/transaction_builder.js
Shaptic: :100: agreed, much cleaner, or just `setTimeout(max)` and setting `timebounds` via the builder constructor (I thought setTimebounds existed for some reason).

src/transaction_builder.js
Shaptic: Couple reasons (@leighmcculloch and I discussed this at length during the design phase): 
 - bwd compat: we want transactions built before this SDK release and after this SDK release to result in the same XDR
 - optimization: V2 w/ only timebounds forms a larger binary because of the other (empty) fields

src/transaction_builder.js
Shaptic: ah, conflating my SDKs... just `setTimeout`, then

src/transaction_builder.js
```diff
@@ -153,33 +172,284 @@ export class TransactionBuilder {
...
+   * For the transaction to be valid, there must be a signature corresponding
+   * to every Signer in this array, even if the signature is not otherwise
+   * required by the sourceAccount or operations.
+   * Internally this will set the <code>extraSigners</code> precondition.
+   *
+   * @param {string[]} extraSigners The required extra signers.
+   * @return {TransactionBuilder}
+   * @see TransactionPreconditions
+   */
+  setExtraSigners(extraSigners) {
```
piyalbasu, line 440: NIT: it may be worth checking that `extraSigners` is an array and throwing an error if not. If it's just a string, for example, it will throw an error on LN 540 instead of defaulting to `[]`

Shaptic:
@paulbellamy I pushed some changes ([diff here](https://github.com/stellar/js-stellar-base/compare/e2023de9de0a273cecd3bfb4fcee3071ffcda64b...5b8517893d0af18a0defb6e15c439d9f2020dc0f)) that get us closer to a complete implementation. 

TL;DR:
 - `.?` isn't compatible with IE :sob: 
 - `minAccountSequence` from `number` to `string` (since it's an int64)
 - building timebound vs v2 preconditions safer
 - minor xdr usage fixes

Also, I'm honestly thinking maybe we should drop the new `TransactionBuilder.set*` methods (besides `setTimeout` which already existed). It doesn't seem worth it to even have them, because (a) dealing with defaults gets wonky, e.g. minSeqAge is 0 by default yet we use `null` which complicates things and (b) it's already really straightforward to set them all via the `TransactionBuilder.constructor`

src/transaction_builder.js
```diff
@@ -234,9 +470,52 @@ export class TransactionBuilder {
...
+      const minSeqAge = UnsignedHyper.fromString(
+        this.minAccountSequenceAge !== null
+          ? this.minAccountSequenceAge.toString()
+          : '0'
+      );
+
+      const minSeqLedgerGap = this.minAccountSequenceLedgerGap || 0;
+
+      // TODO: Parse these somehow? or make them a richer type?
+      // Re: Yep, we might need a StrKey -> xdr.SignerKey abstraction :(
```
paulbellamy, line 501: Well.. `StrKey` wouldn't handle signed payloads, right? Maybe instead of a `string[]` input this needs to be a `Signer[]` input?

Shaptic: StrKey handles signed payloads after #511. 

In the Go SDK, we add extra methods on top of `xdr.SignerKey` to make the strkey <--> xdr.SignerKey abstraction possible (see [xdr/signer_key.go](https://github.com/stellar/go/blob/master/xdr/signer_key.go)). We can't do that in JS, really, but we can build a similar translation layer.

To keep this PR simple, though, we should just make this a `[]xdr.SignerKey` over a `[]string`. That way, we can work on the abstraction separately and squeeze it in before the final release.

Shaptic:
Agreed re: the builder pattern. Should we be so strict about "overwriting" properties? Again no strong preference, but it's misleading when you read the code that `this.minAccountSequence = null` but in reality it will be `0` in the final result.

Besides that, I think it LGTM. Can you do a final pass ensuring that all of the types align with the XDR? e.g. in 86bd783 I realized that we didn't need `string`/`UnsignedHyper` for ledgerbounds since

```c++
struct LedgerBounds {
    uint32 minLedger;
    uint32 maxLedger;
};
```

src/transaction_builder.js
```diff
@@ -183,6 +219,209 @@ export class TransactionBuilder {
...
+    }
+
+    if (minLedger < 0) {
+      throw new Error('min_ledger cannot be negative');
+    }
+    if (maxLedger < 0) {
+      throw new Error('max_ledger cannot be negative');
+    }
+
+    if (minLedger > maxLedger) {
```
overcat, line 298: If minLedger != 0 and maxLedger == 0, an exception will be thrown here.

paulbellamy:
> ensuring that all of the types align with the XDR?

The one mismatch remaining is `minAccountSequenceAge`, in the txn builder it's a `number` but in uint64 it's an `Uint64`. But given it is a number of seconds, I don't think we'd need the extra precision. As is now, it mirrors `setTimeout(durationInSeconds: number)`.

src/transaction_builder.js
paulbellamy: Nice catch, thanks.

paulbellamy:
> it's misleading when you read the code that `this.minAccountSequence = null` but in reality it will be `0` in the final result.

I disagree. It's the difference between unset and explicitly set to `0`.

test/unit/transaction_test.js
```diff
@@ -634,6 +635,85 @@ describe('Transaction', function() {
...
+        .setTimeout(5)
+        .setMinAccountSequenceLedgerGap(5)
+        .build();
+      expect(tx.minAccountSequenceLedgerGap).to.equal(5);
+    });
+
+    it('extraSigners', function() {
+      let tx = makeBuilder()
+        .setTimeout(5)
+        .setExtraSigners([address])
```
overcat, line 712: A quick question, maybe not related to this PR. Here, `address === 'GA7QYNF7SOWQ3GLR2BGMZEHXAVIRZA4KVWLTJJFC7MGXUA74P7UJVSGZ'`, but it should actually be an address starting with `P`, right? We currently have four types of SignerKey: ED25519, PRE_AUTH_TX, HASH_X and ED25519_SIGNED_PAYLOAD, can the SignerKey used here only be of type ED25519_SIGNED_PAYLOAD?

Shaptic: All signers can be used here, but yeah, as Paul said, we’re keeping it simple for now and using the XDR directly, but I’ll revisit this to create a simple abstraction layer to it from the strkey addresses.

src/transaction.js
```diff
@@ -120,11 +138,67 @@ export class Transaction extends TransactionBase {
...
+   */
+  get minAccountSequenceLedgerGap() {
+    return this._minAccountSequenceLedgerGap;
+  }
+  set minAccountSequenceLedgerGap(value) {
+    throw new Error('Transaction is immutable');
+  }
+
+  /**
+   * @type {string[]} array of extra signers
```
Shaptic, line 192:
```suggestion
   * @type {xdr.SignerKey[]} array of extra signers
```

src/transaction_builder.js
```diff
@@ -90,6 +91,20 @@ export const TimeoutInfinite = 0;
  *     timestamp or Date object
  * @param {number|string|Date}  [opts.timebounds.maxTime] - 64-bit UNIX
  *     timestamp or Date object
+ * @param {object}              [opts.ledgerbounds] - ledger bounds for the
+ *     validity of this transaction
+ * @param {number}       [opts.ledgerbounds.minLedger] - number of the minimum
+ *     ledger sequence
+ * @param {number}       [opts.ledgerbounds.maxLedger] - number of the maximum
+ *     ledger sequence
+ * @param {string}              [opts.minAccountSequence] - number for
```
Shaptic, line 100:
```suggestion
 * @param {number}              [opts.ledgerbounds.minLedger] - number of the minimum
 *     ledger sequence
 * @param {number}              [opts.ledgerbounds.maxLedger] - number of the maximum
 *     ledger sequence
 * @param {string}              [opts.minAccountSequence] - number for
```

Shaptic (approved):
Some nits I'll resolve, LGTM :+1: :clap:

leighmcculloch (commented):
A question (❓) inline about how we're storing some 64-bit values as `number`s.

PR #516, opened 2022-04-04
Properly do lexographic comparison of ascii strings

### LINKED ISSUE #515: Asset.compare treats asset codes wrong
Hi! This issue came up on the developer discord.

***Describe the bug**:
You currently can't create trustline for liquidity pools using lowercase and upper case assets of the same type.
If you for example want to create a LiquidityPoolTrustline on the assets `aBTC` and `USDA` from the same issuer, stellar base will force you to use `aBTC` as assetA `USDA` as assetB. 
Now if you build and sign that trasnaction and try to submit it, stellar core will return `op_malformed` as it expects the order to be reversed.

This is caused by the usage of `localeCompare` to compare the asset codes [here](https://github.com/stellar/js-stellar-base/blob/06f30c2df2e299b1c52ccb060511118241380a29/src/asset.js#L232) 

**What version are you on?**
Current master.

**To Reproduct**
Open up laboratory on testnet. 
Try to change trust to a liquidity pool share using the asset codes from above and some account as the issuer.
If you sign and submit it, you'll get back `op_malformed`.

Here the link to a working example  transaction on testnet with the currently (by stellar-base) disallowed order: https://horizon-testnet.stellar.org/transactions/798a488ca09b9463ccc40685b84b08aaf2038a29b3481fdb1f79dcf2cedf5287

**Expected behavior**
You should be able to create valid change trustline operations with lower- and uppercase assets using stellar-base.

Shaptic:
There are two changes here to the way that assets are compared:

1. In order to align better with how Stellar Core sorts assets (that is to say, in no special way besides [the default](https://github.com/xdrpp/xdrpp/blob/8756126cc734c11b75e05f174ca35a192860d0a6/xdrpp/types.h#L1019)), we rely on the actual, raw, XDR asset type (`xdr.AssetType`) value to distinguish based on type. The logic is still the same, but is much easier to understand since there are no convoluted branching string comparisons. 

2. JavaScript's [`String.prototype.localeCompare`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) function will sort strings with lowercase letters as having lower preference to those with uppercase ones (proof: [RunKit](https://runkit.com/61833274873f8c000b42e673/624b753823ad3c00081f0f39)). This is not how the ASCII table sorts them, nor is it how Stellar Core will sort them as it essentially looks at the raw byte values for ordering.

For (2), we introduce `asciiCompare`, which hard-codes the `en` locale rather than leaving it system specific and also forces uppercase letters to take precedence.

Closes #515.

### DIFF
--- a/src/asset.js
+++ b/src/asset.js
@@ -147,14 +147,15 @@ export class Asset {
    * * `credit_alphanum12`
    */
   getAssetType() {
-    if (this.isNative()) {
-      return 'native';
-    }
-    if (this.code.length >= 1 && this.code.length <= 4) {
-      return 'credit_alphanum4';
-    }
-    if (this.code.length >= 5 && this.code.length <= 12) {
-      return 'credit_alphanum12';
+    switch (this.getRawAssetType()) {
+      case xdr.AssetType.assetTypeNative():
+        return 'native';
+      case xdr.AssetType.assetTypeCreditAlphanum4():
+        return 'credit_alphanum4';
+      case xdr.AssetType.assetTypeCreditAlphanum12():
+        return 'credit_alphanum12';
+      default:
+        break;
     }
 
     return null;
--- a/src/asset.js
+++ b/src/asset.js
@@ -161,9 +161,7 @@ export class Asset {
   }
 
   /**
-   * Returns the XDR representation of the asset type.
-   *
-   * @return {xdr.AssetType} the raw asset type structure
+   * @returns {xdr.AssetType}  the raw XDR representation of the asset type
    */
   getRawAssetType() {
     if (this.isNative()) {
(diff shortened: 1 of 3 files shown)


src/asset.js
```diff
@@ -160,6 +160,21 @@ export class Asset {
     return null;
   }
 
+  /**
+   * @returns {xdr.AssetType}  the raw XDR representation of the asset type
+   */
+  getRawAssetType() {
```
acharb, line 166: 💭  I don't think it should block this PR, but thoughts on keeping `getAssetType` and `getRawAssetType` to use the same conditional logic? Imo this `getRawAssetType` logic makes more sense, so that null isn't a possible return val (my understanding is there shouldn't ever be an asset that has a code > 12 or 0). Not sure what the standard is for changing legacy code, but thought worth mentioning

acharb (approved):
lgtm! had 1 thought non-blocking

src/asset.js
Shaptic: Good idea! Done in 473595b.

PR #521, opened 2022-04-06
Merge Protocol 19 features into `master`.

Shaptic:


CHANGELOG.md
```diff
@@ -3,7 +3,15 @@
...
-**Note:** As of this release, the minimum supported version of NodeJS is 14.x.
+**Note:** As of this release, the minimum supported version of NodeJS is **14.x**.
+
+### Add
+
+- Support for converting signed payloads ([CAP-40](https://stellar.org/protocol/cap-40)) to and from their StrKey (`P...`) representation ([#511](https://github.com/stellar/js-stellar-base/pull/511)).
+
+- Support for creating transactions with the new preconditions ([CAP-21](https://stellar.org/protocol/cap-21)) via `TransactionBuilder` ([#513](https://github.com/stellar/js-stellar-base/pull/513)).
+
+- A way to convert between addresses (like `G...` and `P...`, i.e. the `StrKey` class) and their respective signer keys (i.e. `xdr.SignerKey`s), particularly for use in the new transaction preconditions ([#520](https://github.com/stellar/js-stellar-base/pull/520)).
```
paulbellamy, line 14: There's probably a few more things we've added here? like api support for protocol 19, etc.

paulbellamy (approved):

types/index.d.ts
```diff
@@ -846,22 +846,26 @@ export type Operation =
```
overcat, line 846: Is `signPayloadDecorated(data: Buffer): xdr.DecoratedSignature;` missing in the file?

src/strkey.js
```diff
@@ -136,31 +146,117 @@ export class StrKey {
...
+    case 'sha256Hash':
+      return decoded.length === 32;
+
+    case 'med25519PublicKey':
+      return decoded.length === 40; // +8 bytes for the ID
+
+    case 'signedPayload':
+      return (
+        // 32 for the signer, +4 for the payload size, +64 for the max payload
+        decoded.length >= 32 + 4 && decoded.length <= 32 + 4 + 64
```
overcat, line 254: According to the description in the latest CAP-40, the length of the payload is at least 1, maybe this should be modified to `decoded.length >= 32 + 4 + 4`?

See [Transaction Validity](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0040.md#transaction-validity)

CHANGELOG.md
Shaptic: API support is part of stellar/js-stellar-sdk; I don't want to conflate what changed where

types/index.d.ts
Shaptic: Yes it is - nice catch! Could've sworn I added it..

src/strkey.js
Shaptic: This condition was only just introduced. Thanks for the catch!

PR #522, opened 2022-04-06
Encodes the correct field name in the `RevokeLiquidityPoolSponsorship` type

### LINKED ISSUE #502: RevokeLiquidityPoolSponsorship interface returns balanceId instead of liquidityPoolId
**Describe the bug**
TypeScript interface Operation.RevokeLiquidityPoolSponsorship says the operation returns a string called balanceId, it doesn't. It returns a string called liquidityPoolId but TypeScript says no liquidityPoolId exists on that operation due to the bug in the interface.

**What version are you on?**
Stellar-base 7.0.0, stellar-sdk 10.0.0

**Expected behavior**
liquidityPoolId exists on the interface of Operation.RevokeLiquidityPoolSponsorship

**Additional context**
image

Shaptic:
Closes #502.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -746,7 +746,7 @@ export namespace Operation {
   ): xdr.Operation<RevokeClaimableBalanceSponsorship>;
 
   interface RevokeLiquidityPoolSponsorship extends BaseOperation<OperationType.RevokeSponsorship> {
-    balanceId: string;
+    liquidityPoolId: string;
   }
   function revokeLiquidityPoolSponsorship(
     options: OperationOptions.RevokeLiquidityPoolSponsorship
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 
 ### Fix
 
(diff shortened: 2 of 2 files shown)


quietbits (approved):

types/index.d.ts
```diff
@@ -747,7 +747,7 @@ export namespace Operation {
   ): xdr.Operation<RevokeClaimableBalanceSponsorship>;
 
   interface RevokeLiquidityPoolSponsorship extends BaseOperation<OperationType.RevokeSponsorship> {
-    balanceId: string;
+    liquidityPoolId: string;
```
leighmcculloch, line 750: No test?

leighmcculloch (approved):
Looks fine, no test though?

types/index.d.ts
Shaptic: Yeahh.. unfortunately the entire set of operations doesn't have tests. This is a one-off fix but I have a long list of how we can get better TypeScript testing in this library (and better API testing in general...). Unfortunately, it's a behemoth of a task.

PR #524, opened 2022-04-11
Add round-trip XDR tests for transaction preconditions

Shaptic:
Instead of just checking the returned `Transaction` abstraction, the tests now additionally check that the final, built XDR (i.e. the `xdr.TransactionEnvelope`) also contains the right preconditions within it.


leighmcculloch (approved):
:clap

PR #534, opened 2022-06-08
Add support for signed payloads to the `setOptions` operation.

Shaptic:
While many parts of the SDK have been updated to enable the new signer type from [CAP-40](https://stellar.org/protocol/cap-40), the `SetOptions` operation has not.

As part of this, new helpers exist:
```typescript
  function composeSignedPayload(address: string, payload: Buffer): string;
  function decomposeSignedPayload(address: string): { signer: string; payload: Buffer };
```

which should make it easy to create a P-address from a G-address and a payload and vice-versa.

### DIFF
--- a/src/signerkey.js
+++ b/src/signerkey.js
@@ -98,7 +98,7 @@ export class SignerKey {
    *
    * @returns {string}   the StrKey-encoded signed payload address (P...)
    */
-  static encodeSignedPayloadFromAddress(address, payload) {
+  static composeSignedPayload(address, payload) {
     return SignerKey.encodeSignerKey(
       xdr.SignerKey.signerKeyTypeEd25519SignedPayload(
         new xdr.SignerKeyEd25519SignedPayload({
@@ -107,5 +107,26 @@ export class SignerKey {
         })
       )
     );
+  }
+
+  /**
+   * Decomposes a signed payload (P...) into its signer (G...) and payload.
+   *
+   * @param {string} pAddress  the StrKey-encoded signed payload address (P...)
+   *
+   * @returns {object} an object with two keys:
+   *     `signer` indicating the signer of the signed payload as a StrKey (G...)
+   *     `payload` specifying the raw payload that was signed
+   */
+  static decomposeSignedPayload(pAddress) {
+    const sp = xdr.SignerKeyEd25519SignedPayload.fromXDR(
+      StrKey.decodeSignedPayload(pAddress),
+      'raw'
+    );
+
+    return {
+      signer: StrKey.encodeEd25519PublicKey(sp.ed25519()),
+      payload: sp.payload()
+    };
   }
 }
--- a/src/signerkey.js
+++ b/src/signerkey.js
@@ -89,4 +89,23 @@ export class SignerKey {
 
     return encodeCheck(strkeyType, raw);
   }
+
+  /**
+   * Creates a signed payload address (P...) from a signer (G...) and a payload.
+   *
+   * @param  {string} address  the signer of the signed payload, as a StrKey (G...)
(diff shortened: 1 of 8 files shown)


tyvdh (commented):
The ergonomics on this feel way better and a massive +1 for adding the `setOptions` support. I can't speak to if this should actually get merged as is as I'm not a core SDK contributor but it lgtm from an ergonomics improvement pov.

👏

src/signerkey.js
```diff
@@ -89,4 +89,23 @@ export class SignerKey {
...
+
+  /**
+   * Creates a signed payload address (P...) from a signer (G...) and a payload.
+   *
+   * @param  {string} address  the signer of the signed payload, as a StrKey (G...)
+   * @param  {Buffer} payload  the raw payload that was signed
+   *
+   * @returns {string}   the StrKey-encoded signed payload address (P...)
+   */
+  static encodeSignedPayloadFromAddress(address, payload) {
```
leighmcculloch, line 101: Any reason this can't have a shorter name?
```suggestion
  static encodeSignedPayload(address, payload) {
```

Shaptic: probably not :shrug: that clashes with `StrKey.encodeSignedPayload`, but separate namespaces should be enough of a hint that they're different

Shaptic: ooh how about `composeSignedPayload` :notes:

leighmcculloch: What is `StrKey.encodeSignedPayload` for? It seems like a very low-level thing to expose to developers. We don't expose it in other SDKs.

Shaptic: All of the `StrKey` functions in the SDK convert between raw buffers and strkeys. I'm not a fan and it deserves a refactor to line up more w/ the other SDK implementations, but that's historical.

src/signerkey.js
```diff
@@ -89,4 +89,44 @@ export class SignerKey {
...
+
+  /**
+   * Creates a signed payload address (P...) from a signer (G...) and a payload.
+   *
+   * @param  {string} address  the signer of the signed payload, as a StrKey (G...)
+   * @param  {Buffer} payload  the raw payload that was signed
+   *
+   * @returns {string}   the StrKey-encoded signed payload address (P...)
+   */
+  static composeSignedPayload(address, payload) {
```
leighmcculloch, line 101: Can we name this function `encodeSignedPayload`? It fits existing terms used in the SignerKey type and doesn't clash with any existing function. Using a new term `compose` seems unnecessary.

src/signerkey.js
leighmcculloch: I think the historic reason was fine because once-upon-a-time all strkeys were just a single 32byte buffer.

It's unintuitive to me that a user must use a SignerKey object to generate a strkey. Maybe we should do soemthing like what we did with `MuxedAccount`s, where you create a `SignedPayloadSigner` which is an object that sits apart, and then you can generate a `SignerKey` from it.

orbitlens:
Just realized that my PR #542 was intended to solve the same problem. Wish I'd done more comprehensive search before writing the code. There were no mentions of this PR in issues of this repo.

Is there any ETA? Quite a high-priority issue for us. Can we go at least with a simple MVP that allows to build a transaction plus correctly parse tx XDRs, and implement additional tooling later if there's no consensus on naming?

leighmcculloch:
Looks like this PR fell through the cracks, it was intended to go out when protocol 19 was released. cc @Shaptic

Shaptic:
I think we can close this in lieu of the merged-and-released #542. If usability and ergonomics become a concern, we can revisit this design.

PR #537, opened 2022-08-12
Replace Rakefile with Makefile

paulbellamy:
This does a few things:
- Replace Rakefile with Makefile (to match https://github.com/stellar/rs-stellar-xdr).
- Remove all ruby. Not needed anymore. xdrgen is dockerized
- Automate and dockerize the dts-xdr build step.
- Split `xdr` directory into `xdr/curr` and `xdr/next`. Which one it uses is hard-coded in `src/xdr.js` and `types/xdr.d.ts` for now.
- Cleaned up the readme for all of the above.

- Incidentally, this updated the curr xdr to latest core master. But could leave that out of this PR if desired.

The main potential issue with this are our hand-edit hacks in the .d.ts files. So like.. make can re-generate those files, but... They'll be broken until you copy-pasta over the changes again. oof.


leighmcculloch (commented):
This looks good to me. How will people choose current vs next in the SDK? We wouldn't want to release the next XDR, as it is very unstable. Or rather, we'd only want to release next on non-regular release numbers, like maybe a pre-release, or something else.

paulbellamy:
Yeah, my plan for now was to do a separate branch and switch it in there. But figured backporting this would minimize the diff later.

leighmcculloch (approved):
> Incidentally, this updated the curr xdr to latest core master. But could leave that out of this PR if desired.

If doing this is not inconvenient, could we leave that out of this PR so that this change is isolated to a refactor only, then follow up with another PR where we generate it?

Otherwise 👍🏻.

Makefile
```diff
@@ -0,0 +1,88 @@
+XDR_BASE_URL_CURR=https://github.com/stellar/stellar-core/raw/master/src/protocol-curr/xdr
+XDR_BASE_LOCAL_CURR=xdr/curr
+XDR_FILES_CURR= \
+	Stellar-SCP.x \
+	Stellar-ledger-entries.x \
+	Stellar-ledger.x \
+	Stellar-overlay.x \
+	Stellar-transaction.x \
+	Stellar-types.x
+XDR_FILES_LOCAL_CURR=$(addprefix xdr/curr/,$(XDR_FILES_CURR))
```
Shaptic, line 10: This whole `curr` and `next` thing broke the entire ecosystem, I bet. Why couldn't they keep `xdr/` and add `xdr-next/`??

README.md
```diff
@@ -141,39 +141,8 @@ versions.)
...
-# If src/generated/stellar-xdr_generated.js changed, then:
-git clone https://github.com/stellar/dts-xdr
-cd dts-xdr
-npm install
-OUT=stellar-xdr_generated.d.ts npx jscodeshift -t src/transform.js ../src/generated/stellar-xdr_generated.js
-cp stellar-xdr_generated.d.ts ../types/xdr.d.ts
-cd .. && rm -rf dts-xdr
-yarn run prettier --write types/xdr.d.ts
-```
+1. Make sure you have [Docker](https://www.docker.com/) installed and running.
```
Shaptic, line 144: Better link: https://docs.docker.com/engine/install/

package.json
```diff
@@ -7,7 +7,6 @@
   "scripts": {
     "test": "gulp test:node",
     "test:watch": "gulp test:watch",
-    "xdr": "bundle exec xdrgen -o src/generated -l javascript -n stellar-xdr xdr/Stellar-ledger-entries.x xdr/Stellar-ledger.x xdr/Stellar-overlay.x xdr/Stellar-SCP.x xdr/Stellar-transaction.x xdr/Stellar-types.x",
```
Shaptic, line 10: Why not add the `make reset-xdr` command here?

Shaptic (commented):
was out sick so late to the party but figured I'd add my thoughts anyway

PR #539, opened 2022-08-16
fix: fixed `TransactionBuilder.build()` to set correct `minAccountSequence`

overcat:
Please check the test case for more information.


Shaptic (approved):
Nice catch & fix, thank you! :clap:

PR #548, opened 2022-09-21
Fix github action to publish soroban tag

paulbellamy:


Shaptic (approved):
I'm mildly concerned about the separation between the "trigger" `on` and the two conditionals (i.e. there's now a weird chance of publishing to npm in a non-release scenario), but it does do what we need it to do :+1:

PR #566, opened 2022-12-08
New soroban xdr updates

paulbellamy:
- Missed an instance of the old `.parameters()` field
- Also, fixing a bug introduced in https://github.com/stellar/js-stellar-base/pull/557

### DIFF
--- a/src/contract.js
+++ b/src/contract.js
@@ -36,21 +36,22 @@ export class Contract {
    * @returns {xdr.Operation} Build a InvokeHostFunctionOp operation to call the contract.
    */
   call(method, ...params) {
+    let contractId = Buffer.from(this._id, 'hex');
     return Operation.invokeHostFunction({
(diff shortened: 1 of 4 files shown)


src/contract.js
```diff
@@ -36,21 +36,22 @@ export class Contract {
...
         ...params
       ]),
       // TODO: Figure out how to calculate this or get it from the user?
       footprint: new xdr.LedgerFootprint({
         readOnly: [
-          new xdr.LedgerKeyContractData({
-            contractId: this._id,
-            key: xdr.ScVal.scvStatic(xdr.ScStatic.scsLedgerKeyContractCode())
-          })
+          xdr.LedgerKey.contractData(
```
Shaptic, line 49: does this not need to be `new`'d?

Shaptic (approved):

src/contract.js
paulbellamy: I think it's a static method

PR #567, opened 2022-12-13
make sure sodium is not an empty object in service workers

piyalbasu:


src/signing.js
```diff
@@ -42,6 +42,10 @@ function checkFastSigningNode() {
     return checkFastSigningBrowser();
   }
 
+  if (!Object.keys(sodium).length) {
+    return checkFastSigningBrowser();
+  }
```
piyalbasu, line 47: this codebase determines if a user is in a node env by checking `window === "undefined"`. In a node env, the node lib `sodium-native` is installed

What currently happens: In a service worker, window is undefined. But, since it's not a node env, `sodium-native` is not installed. So, the var `sodium` is an empty object when it gets here. This cause the methods below to break.

What happens after the change: we do another check to make sure `sodium` has the appropriate keys. If it doesn't, use `checkFastSigningBrowser`

CHANGELOG.md
```diff
@@ -3,6 +3,11 @@
 
 ## Unreleased
 
+## [v8.2.2](https://github.com/stellar/js-stellar-base/compare/v8.2.1..v8.2.2)
+
+### Fix
+
+- Enable signing in service workers using FastSigning.
```
Shaptic, line 10: We usually include PR references:

```suggestion
- Enable signing in service workers using FastSigning ([#567](https://github.com/stellar/js-stellar-base/pull/567)).
```

src/signing.js
Shaptic: I'm following what you're saying, but does that mean that the `require('sodium-native')` line does **not** throw in the Service Worker case, and just silently returns `{}`?

Shaptic (approved):
LGTM! But I had a clarifying question as well.

src/signing.js
piyalbasu: Yup, that's right. So here is sort of another edge EDGE case - if you're compiling with Webpack and have not included `sodium-native` in your package, Webpack will recognize the missing dependency and silently replace `require('sodium-native')` with an empty `{}`.

Shaptic: lol ok awesome, thanks for clarifying! LGTM :+1:

PR #572, opened 2023-02-07
Add strkey parsing for contracts, and Address helper for building ScAddresses

paulbellamy:
Two things here:
- There's a new strkey type, for `C...` keys for contracts. This adds strkey parsing/serialization for those.
- When updating soroban-example-dapp, I found building ScAddresses to be a faff. Added an `Address` class to make that easier.
  - I considered having two classes (`AccountAddress`, and `ContractAddress`), but didn't see the point for now.
  - You would use this via: `new Address("C..34").toScVal().toXDR("base64")`, or `.toScAddress()` if that is what you wanted.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -12,6 +12,14 @@ export class Account {
   incrementSequenceNumber(): void;
 }
 
+export class Address {
+  constructor(address: string);
+  static fromString(address: string): Address;
+  toString(): string;
+  toScVal(): xdr.ScVal;
+  toScAddress(): xdr.ScAddress;
+}
+
 export class Contract {
   constructor(contractId: string);
   contractId(): string;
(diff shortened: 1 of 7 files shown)


types/index.d.ts
```diff
@@ -12,6 +12,16 @@ export class Account {
...
 
+export class Address {
+  constructor(address: string);
+  static fromString(address: string): Address;
+  static account(buffer: Buffer): Address;
+  static contract(buffer: Buffer): Address;
+  toString(): string;
+  toScVal(): xdr.ScVal;
+  toScAddress(): xdr.ScAddress;
+}
```
Shaptic, line 23: is there an `xdr.Address`, too?

Shaptic (approved):
Nice! Clean and well-tested.

types/index.d.ts
paulbellamy: lmao, no.

PR #584, opened 2023-04-04
SDK Modernization

Shaptic:
**Epic: stellar/js-stellar-sdk#792**

This is an overhaul to bring the `stellar-base` package to the modern JavaScript world.

## Summary
This involves a large set of changes, but I will do my best to summarize them to expedite review.

### Main Changes

 - [x] Webpack 5 + Babel is now the main build system
 - [x] GulpJS is no longer used for any build pipelines
 - [x] All dependencies have either been upgraded to their latest versions **OR** replaced with maintained equivalents.
   * This involved code changes to bring the library up to compatibility with `BigNumber.js`.

### Cascading Changes
These main changes motivate a cascade of other changes, broken down into a handful of categories:

 - [Building](#building): Packaging is done via Webpack for the browser and Babel for Node
 - [Linting](#linting): ESLint is updated, JS/TS linting is better
 - [Testing](#testing): Libraries are updated + modernized
 - [Running](#scripts): `yarn` scripts are different

#### Building
Webpack 5 support is here!

 - the node package is built using Babel via `yarn build:node`
 - the browser bundle is built using Webpack via `yarn build:browser`
   * this is configured in `cfg/webpack.config.browser.js`
   * its accessible after including the script via the `StellarBase` variable
   * it adds a handful of important polyfills
   * it does linting via ESLint

#### Linting
ESLint, Prettier, and DTSLint are all updated:

 - ESLint works with Babel via the `@babel/eslint-parser` in each `.eslintrc.js` file and is implicitly part of building via Webpack (see above)
 - DTSLint is part of the `@definitelytyped` project and has additional configuration implicitly as part of the `types/.eslintrc.js` file which uses `@typescript-eslint/parser`, instead
 - Prettier is configured in `cfg/prettier.config.js` (and `cfg/.prettierignore`) and runs on all files now. It's also (still) a pre-commit hook.

Just run `yarn pretty && yarn lint` to format your files and check that everything is up to par.

#### Testing

**Istanbul**: Istanbul is [unmaintained](https://www.npmjs.com/package/istanbul) and superceded by the `nyc` package.

 - the `istanbul` plugin in `.babelrc` tells Babel about it
 - `cfg/.nycrc` extends a Babel-compatible configuration file ([docs](https://github.com/istanbuljs/nyc#babel-projects))
 - the `nyc` key in `package.json` configures instrumentation behavior
 - the coverage is *actually* provided by `babel-plugin-istanbul` ([docs](https://github.com/istanbuljs/babel-plugin-istanbul)), which is why `"instrument": false` in the above

**Node Testing**: Mocha (and its helper friends Sinon and Chai) has been updated.

 - the `mocha` key in `package.json` configures Mocha (replacing `mocha.opts`)
 - this does Babel transpiling and hooks in the helper script
 - code coverage is provided by `nyc` (see above)

**Browser Testing**: Karma has been updated.

 - the SauceLabs Karma configuration has been removed entirely because it wasn't being used
 - Karma has been updated with a code coverage integration with `nyc` ([docs](https://github.com/istanbuljs/babel-plugin-istanbul#karma), see above)
 - It relies on the Webpack configuration but makes some minor changes

Run `yarn test` to run both suites.

#### Scripts
Development has a slightly different workflow now:

 - build+test locally with `yarn test:node`
 - this will only run node tests (which are faster/simpler)
 - you can `yarn build && yarn test:all` to get the both test suites (note that building first is *required*)
 - run `yarn preversion` when you want a thorough, clean test of the whole system in "production" mode


Shaptic:
@BenRacicot would appreciate your review as well

lib/account.js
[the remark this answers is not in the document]
paulbellamy: Also, if we are keeping these, we should probably add a CI check that these match the expected compiled version before allowing merge to master

package.json
```diff
@@ -1,22 +1,43 @@
...
   "scripts": {
-    "test": "gulp test:node",
-    "test:watch": "gulp test:watch",
-    "docs": "jsdoc -c .jsdoc.json --verbose",
-    "dtslint": "dtslint types --localTs node_modules/typescript/lib",
-    "preversion": "gulp test",
-    "version": "gulp build",
-    "postversion": "git push && git push --tags",
-    "prettier-all": "prettier --write **/*.js"
+    "build": "yarn build:node",
```
paulbellamy, line 9: Should be this, no?
```suggestion
    "build": "yarn build:all",
```

package.json
```diff
@@ -1,22 +1,43 @@
...
+    "build:node": "babel --out-dir lib/ src/",
```
paulbellamy, line 10: also `build:node:prod` for completeness?

package.json
```diff
@@ -1,22 +1,43 @@
...
+    "build:browser": "webpack -c ./config/webpack.config.browser.js",
+    "build:prod": "cross-env NODE_ENV=production yarn build",
+    "build:browser:prod": "cross-env NODE_ENV=production yarn build:browser",
+    "build:all": "yarn build:node && yarn build:browser",
+    "build:all:prod": "yarn build:prod && yarn build:browser:prod",
+    "test": "yarn build && yarn test:node",
```
paulbellamy, line 16: Also here?
```suggestion
    "test": "yarn build && yarn test:all",
```

lib/account.js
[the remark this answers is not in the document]
paulbellamy: Actually, no I see below `/lib` was excluded from the `files` list. So why are these here?

src/util/continued_fraction.js
```diff
@@ -11,6 +11,8 @@ const MAX_INT = ((1 << 31) >>> 0) - 1;
  * @returns {array} first element is n (numerator), second element is d (denominator)
  */
 export function best_r(rawNumber) {
+  BigNumber.DEBUG = true; // throws on bad number values
```
paulbellamy, line 14: Should probably do this globally, or unset it after this function.

paulbellamy (commented):
Outstanding work! 💯 🚀 

Couple questions...

lib/account.js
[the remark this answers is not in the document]
Shaptic: Ah yeah so in the latest commit I pushed it because I was trying to see if the changes were compatible with `js-stellar-sdk` via the git branch. I wasn't expecting it to be reviewed, but I'm glad you did, because it pointed me to lifecycle scripts (which are different for [yarn](https://yarnpkg.com/advanced/lifecycle-scripts) and [npm](https://docs.npmjs.com/cli/v9/using-npm/scripts#life-cycle-scripts), yay) which I think should accomplish what I was working around. I was confused about how installing from `npm` would render the `lib/` folder but it wouldn't from `git`, and I think that's exactly the reason. Thanks!

package.json
Shaptic: my intent was for `yarn build` to just do the node stuff for the local dev+test cycle because the browser stuff is slower, but I can see how that'd be misleading (under the mindset of "the browser stuff should 'just work' if I did the node stuff correctly, so I'd rather opt-in to build+testing it explicitly)

package.json
Shaptic: for completeness, true, but there is no prod setting for the node side (for the browser it does a minify and no source map)

src/util/continued_fraction.js
Shaptic: good call, fixed

package.json
Shaptic: ok, cleaned them up to be a little more consistent!

Makefile
silence48, line 1: I think this should be using the url for current:
https://github.com/stellar/stellar-core/tree/master/src/protocol-curr/xdr
for next:
https://github.com/stellar/stellar-core/tree/master/src/protocol-next

silence48 (commented):
I only came across one questionable thing which was the address to the xdr library...

Many of these changes are code linting, which is fine, but I'd prefer to see those broken into a separate commit to be able to quickly find the substantive changes.

Otherwise it looks good.

silence48 (approved):
that works, `/generated` after all.

lib/account.js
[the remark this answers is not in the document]
paulbellamy: There's a major-ish gotcha with `yarn prepare`. When you `npm i github.com/stellar/js-stellar-base`, it will run `yarn prepare` *unless* you are root (pretty common in docker).

paulbellamy (approved):
LGTM, but I'm not an expert here.

package.json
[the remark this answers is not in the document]
silence48: That will always be the node.js version. to get the browser version, you'd need to pull the repo and run `npm run build:browser`
You can generally find this type of info by looking at the package.json `scripts` section.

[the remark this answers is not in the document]
Shaptic: yep! you can either build it as @silence48 said, or use bower (this is in the [README](https://github.com/stellar/js-stellar-base#to-self-host-for-use-in-the-browser), and the repo is [auto-generated](https://github.com/stellar/bower-js-stellar-sdk))

package.json
```diff
@@ -1,22 +1,42 @@
...
+    "build:node:prod": "cross-env NODE_ENV=production yarn build",
+    "build:browser:prod": "cross-env NODE_ENV=production yarn build:browser",
+    "build:prod": "cross-env NODE_ENV=production yarn build",
+    "test": "yarn build && yarn test:node && yarn test:browser",
+    "test:node": "nyc --nycrc-path ./config/.nycrc mocha --recursive",
+    "test:browser": "karma start ./config/karma.conf.js",
+    "docs": "jsdoc -c ./config/.jsdoc.json --verbose",
+    "lint": "eslint -c ./config/.eslintrc.js src/ && dtslint --localTs node_modules/typescript/lib types/",
+    "preversion": "yarn clean && yarn pretty && yarn lint && yarn build:prod && yarn test",
+    "pretty": "prettier --config ./config/prettier.config.js --ignore-path ./config/.prettierignore --write './**/*.js'",
```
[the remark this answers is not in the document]
Shaptic: `fmt` sounds like the best balance (you can't do `prettier` because that conflicts w/ the command itself heh)

package.json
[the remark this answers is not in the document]
paulbellamy: Yeah, I was more asking if there is a standard or some way to have it auto-use the correct version. But it seems like no.

paulbellamy:
Also, if you could merge these changes into the `soroban` branch, after they are merged into main. 🙏

PR #587, opened 2023-04-20
Update generated XDR to latest version.

Shaptic:
aka I just ran `make reset-xdr` and added the `Operation` workaround


sreuland (approved):
just curious, do you know reason why the generated *.ts files couldn't be in a path under `src/generated` as well.

Shaptic:
@sreuland so because this library isn't "Typescript-native," you are supposed to place all type definitions under the `types/` folder (and then reference this in "types" field in the package.json).

PR #588, opened 2023-04-20
temp: share i128 <=> ScVal::i128 conversions

mootz12:
NOT INTENDED TO BE MERGED

Sharing a working ScVal::i128 <=> bigint converter after discussion with @paulbellamy , as it might be of use to the folks working on a robust js <=> ScVal conversion solution.

### DIFF
--- a/src/scvi128.js
+++ b/src/scvi128.js
@@ -1,0 +1,100 @@
+import xdr from "./xdr";
+
+/**
+ * @param {bigint} value
+ * @returns {xdr.ScVal}
+ */
+export function bigintToScvI128(value) {
+  const hex = value.toString(16).replace(/^-/, "");
+  if (hex.length > 32) {
+    throw new Error("value overflow i128");
+  }
+
+  const buf = Buffer.alloc(16);
+  buf.write(hex, 16 - Math.ceil(hex.length / 2), "hex"); // BE
+
+  // perform two's compliment if negative
+  if (value < 0) {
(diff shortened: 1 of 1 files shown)


src/scvi128.js
```diff
@@ -0,0 +1,100 @@
...
+    throw new Error("value overflow i128");
+  }
+
+  const buf = Buffer.alloc(16);
+  buf.write(hex, 16 - Math.ceil(hex.length / 2), "hex"); // BE
+
+  // perform two's compliment if negative
+  if (value < 0) {
+    // throw if MSB bit is 1 and is not i128:MIN
+    if ((buf[0] & 0x80) != 0 && hex != "80000000000000000000000000000000") {
```
Shaptic, line 19: there must be a better way than this string comparison :sweat_smile: 

what about some equivalent of `value != (1 << 64)`?

tomerweller:
cc @orbitlens who was also looking at i128 conversions

src/scvi128.js
mootz12: There likely is, yes! (and probably many more improvements that can be made)

The main goal here was to find when the input value was equivalent to `i128::MIN`, cause any other number with the 128th bit as 1 would be out of range. If `number` can support this with accuracy, I'd imagine `value  != BigInt(1 << 127)` would work.

Shaptic:
Closing in lieu of #620.

PR #591, opened 2023-04-20
Add fully-qualified `Operation` type to XDR type definitions.

Shaptic:
`Operation` refers to the high-level operation, when it should be `xdr.Operation` (or, alternatively, `Operation2`, but that's uglier).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,15 @@
 
 ## Unreleased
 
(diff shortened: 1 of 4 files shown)


paulbellamy (approved):

paulbellamy:
Actually, looking at the types more. I'm not sure this is right. For example, `Operation.toXDR`, seems like it should take an `Operation`, not an `xdr.Operation`.

Shaptic:
nah @paulbellamy, cuz `xdr.Operation` is aliased to it so they're the same thing

PR #599, opened 2023-05-05
update 'next' xdr to latest preview 9 version

sreuland:
includes xdr updates for fees and multi-function invokehostfunction xdr changes.

closes #596

PR #601, opened 2023-05-08
Update XDR and contract invocations to conform to the latest version (Preview 9).

Shaptic:
Major changes:

 - [x] raw XDR updates (via `make reset-xdr` + manual changes like in https://github.com/stellar/js-stellar-base/commit/d3a6b9887dd4ebffb5e0cec77f6512391d6e8747)
 - [x] code updates to conform to the above, mostly changing any interactions with `xdr.InvokeHostFunctionOp` to allow multiple function calls
 - [x] a new `Operation.invokeHostFunctions` (note the plural `s`) to do multiples, in order to preserve usage compatibility with the previous, single-invoke case

The diff for `src/generated/*.js` (too large for GitHub) is in [this gist](https://gist.github.com/Shaptic/e9c5caa207df82752d91d5e41029241f).

This supersedes #599.


sreuland (approved):
thanks for taking this to the finish line, there are details captured here for 'next' protocol of soroban of which I didn't notice and missed, as had just done the xdr gen step only. One question wanted to confirm is on the hash used from  https://github.com/stellar/stellar-xdr#next for codegen, do you want that to be the latest from branch which had a change just today, or to be [2f16687fdf6f4b](https://github.com/stellar/stellar-xdr/commit/2f16687fdf6f4bcfb56805e2035f69997f4b34c4), which had been ref'd in association with last core build for upcoming release

Shaptic:
fuuuuuuuuuuuuuuuuuuuck, great catch, thanks :disappointed: it looks like the changes break the meta/event stuff pretty significantly, so this will probably break `js-soroban-client` downstream

I've rectified that in 2d0df39, I think :raised_hands:

PR #605, opened 2023-05-10
Update asset sorting method to match other sdk's

### LINKED ISSUE #593: Lexicographical sorting is not in pair with what the network expects
## **Describe the bug**
Back in January I got contacted by one of the users of xBull and he told me that he was not able to create a pool with two assets he's holding so I started trying it myself with the same assets and I realized that there is a mismatch between the Horizon/validators code and the JS SDK. And yesterday in the Stellar Developers discord we saw another user with the same issue but with different assets

In the JS SDK the validation of the LP assets order is done like this:
```js
    if (_asset.Asset.compare(assetA, assetB) !== -1) {
      throw new Error('Assets are not in lexicographic order');
    }
```

Which is done in the file `liquidity_pool_asset.js` with the final validation being done with: 

```js
function asciiCompare(a, b) {
  return a.localeCompare(b, 'en', { caseFirst: 'upper' });
}
```

The issue here is that with some assets combinations, the Javascript library gives a different result than the ones from the network or even with other SDKs (python for example)

## **What version are you on?**
v8.2.2

## **To Reproduce**
Here is an example written in both JS and Python (@silence48 helped me with the python part) to show the difference:

**JS**
```js
const fn = async () => {
  const loadedAccount = await server.loadAccount('GCOQDNWHUCTE5Y2II6UBJ2CBOZVU23I7YIZZHTOUMOGEPZEY3WN2FDSB');
  const sourceAccount = new SDK.Account(loadedAccount.account_id, loadedAccount.sequence);
  const assetA = new SDK.Asset('USDC', 'GA5ZSEJYB37JRC5AVCIA5MOP4RHTM335X2KGX3IHOJAPP5RE34K4KZVN');
  const assetB = new SDK.Asset('iDRA', 'GA7NLOF4EHWMJF6DBXXV2H6AYI7IHYWNFZR6R52BYBLY7TE5Q74AIDRA');
  const asset = new SDK.LiquidityPoolAsset(
    assetA,
    assetB,
    30,
  );

  const tx = new SDK.TransactionBuilder(loadedAccount, {
      fee: 10000000,
      networkPassphrase: 'Public Global Stellar Network ; September 2015'
    })
    .setTimeout(0)
    .addOperation(SDK.Operation.changeTrust({ asset }));
}
fn();
```

At this point the library will throw an error saying the lexicografical order is wrong, but this is not the case if we take in count that the network do expects this order (we can confirm it here: `https://horizon.stellar.org/liquidity_pools/ba986a54a2623ca07596e25aedcf688b640a6646e024b33e3fa47d1c8dceb926`)

Now unlikely to the JS version, if we do the same with python we don't face that issue:
```python
from stellar_sdk import Keypair, Network, Server, TransactionBuilder, Asset, LiquidityPoolAsset
import requests

server = Server(horizon_url="https://horizon.stellar.org")
source_account = server.load_account(account_id="GCOQDNWHUCTE5Y2II6UBJ2CBOZVU23I7YIZZHTOUMOGEPZEY3WN2FDSB")

native = Asset.native

asset_1 = Asset("USDC", "GA5ZSEJYB37JRC5AVCIA5MOP4RHTM335X2KGX3IHOJAPP5RE34K4KZVN")
asset_2 = Asset("iDRA", "GA7NLOF4EHWMJF6DBXXV2H6AYI7IHYWNFZR6R52BYBLY7TE5Q74AIDRA")

lp = LiquidityPoolAsset(asset_1, asset_2, fee=30)

transaction = (
    TransactionBuilder(
        source_account=source_account,
        network_passphrase=Network.PUBLIC_NETWORK_PASSPHRASE,
        base_fee=100,
    )
    .set_timeout(0)
    .append_change_trust_op(lp)
)


transaction.build()
```

We confirmed this with the two pairs some users had issues: USDC/iDRA and DRIFT/DicInu

**We also want to say to possible readers:** We do not endorse any of the assets mentioned in this ticket, they were used only because users reported issues with them

## **Expected behavior**
The expected behavior is that the Stellar Base should behave in the same way the network and other stellar libraries behave.

## **Additional context**
Initial message in January: https://discord.com/channels/761985725453303838/761985800132493354/1100230956738154526

Most recent case a user faced: https://discord.com/channels/897514728459468821/905516645408182343/1100066080719900744

silence48:
In order to fix #593 
I update the asciicompare function to properly sort upper case letters before lower case regardless of their position of the alphabet. 

We should probably add some better fuzz tests also.  Also someone else should review + test this before merging it as I only spent a short time on it.  

Thanks!

### DIFF
--- a/src/asset.js
+++ b/src/asset.js
@@ -261,5 +261,29 @@ export class Asset {
  * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare
  */
 function asciiCompare(a, b) {
-  return a.localeCompare(b, 'en', { caseFirst: 'upper' });
+  if (a==b) return 0;
+  for (each in a) {
+    //example if the letters position being compared is in the 4th position, and it hasn't been sorted yet,
+    //and the b string doesnot have a letter in that position, well then... return -1 because a goes first.
+    if((each + 1) > b.length){
+        return -1
+    }
+    //if all the characters are the same and string b is longer, b goes first.
+    if( ( (each +1) == a.length && b.length > a.length ) ){
+        return 1
+        }
+    //if they're both the same letter go to the next letter.
+    if(a[each] == b[each]){continue};
+    //if a is upper case and b is not, return -1 to indicate 'a' goes first.
+    if( (a[each].toUpperCase()  == a[each] && b[each].toUpperCase() != b[each]) ){
+        return -1 
+        }
+    if( (b[each].toUpperCase()  == b[each] && a[each].toUpperCase() != a[each]) ){
+        return 1
+        }    
+    if(( a[each].toUpperCase()==a[each] && b[each].toUpperCase() == b[each] ) ||
+        ( a[each].toLowerCase()==a[each] && b[each].toLowerCase() == b[each] ) ) {
+        return a[each].localeCompare(b[each], 'en', { caseFirst: 'upper' });
+      }
+    }
 }
--- a/src/asset.js
+++ b/src/asset.js
@@ -263,27 +263,28 @@ export class Asset {
 function asciiCompare(a, b) {
   if (a==b) return 0;
   for (each in a) {
+    let letterpos = parseInt(each)
     //example if the letters position being compared is in the 4th position, and it hasn't been sorted yet,
     //and the b string doesnot have a letter in that position, well then... return -1 because a goes first.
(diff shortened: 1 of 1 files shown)


paulbellamy:
Thanks so much for the PR, this is great. I pulled it into https://github.com/stellar/js-stellar-base/pull/606, and did a bit of refactoring.

PR #606, opened 2023-05-11
Fix asset code sorting with mixed cases

paulbellamy:
Fixes https://github.com/stellar/js-stellar-base/issues/593

Based on https://github.com/stellar/js-stellar-base/pull/605, but because what we want is ascii byte comparison. We can use `Buffer.compare` to do that for us.

### DIFF
--- a/src/asset.js
+++ b/src/asset.js
@@ -261,30 +261,47 @@ export class Asset {
  * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare
  */
 function asciiCompare(a, b) {
-  if (a==b) return 0;
-  for (each in a) {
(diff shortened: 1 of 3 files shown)


leighmcculloch (approved):

src/asset.js
```diff
@@ -261,5 +261,5 @@ export class Asset {
  * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare
```
sreuland, line 261: just nit, could remove this comment as it's no longer

sreuland (approved):

Shaptic:
Add a changelog entry plox

silence48:
a more elegant solution, didn't think of this.

PR #607, opened 2023-05-15
Migrate code to use the native `BigInt` type wherever possible.

Shaptic:
Per @orbitlens' in https://github.com/stellar/js-stellar-sdk/issues/792#issuecomment-1518413229, this is an attempt to minimize usage of the `bignumber.js` external dependency.

Unfortunately, a full replacement of the library is not possible for a key reason: we need big rationals for price management. Anything to do with offers supports passing real values as strings (that may exceed `number`) and we need to convert those to fractions for the XDR. This necessitates big rationals, which `BigInt` does not support, obviously. Until there's also a native `BigRat`, we will need `bignumber.js` in these (not really) niche cases.

### Known Limitations
Honestly, I'm not certain about the utility of this change given its potential bug surface. Since we still do need `bignumber.js` for big rational values, and since this is imported by `operation.js`, this will still need to be used by a giant portion of downstream consumers. (The number of people using the SDK but **not** using `Operation` is probably minuscule.)


orbitlens:
@Shaptic Ok, here is the solution:

```js
/**
* Divide two numbers with 20 digits precision (result is floored)
* @param {String|Number|BigInt} a
* @param {String|Number|BigInt} b
* @return {String}
**/
function div(a, b) {
    const val = (BigInt(a) * (10n**20n) / BigInt(b)).toString().padStart(20, '0')
    const int = val.substring(0, val.length - 20) || '0'
    const fract =  val.substring(val.length - 20)
    return int + '.' + fract
}
```
It behaves the same as `BigNumber.js` with the only difference that the result is always `floor`ed in contrast to `BigNumber`, where the result is rounded by default. But I guess this doesn't matter with 20-digits fractional part. The precision can be easily doubled if needed.

```js
new BigNumber(2).div(3).toString()      // '0.66666666666666666667'
div(2, 300000)                          // '0.66666666666666666666'
new BigNumber(2).div(300000).toString() // '0.00000666666666666667'
div(2, 300000)                          // '0.00000666666666666666'
new BigNumber(200000).div(3).toString() // '66666.66666666666666666667'
div(200000, 3)                          // '66666.66666666666666666666'

('200000'/'3').toString()               // '66666.66666666667' - regular Numbers division with limited precision
```

PR #612, opened 2023-05-30
Add Contract support for strkey-style contract IDs

paulbellamy:
Will fix the system-tests breakage at: https://github.com/stellar/soroban-tools/actions/runs/5121655313/jobs/9209699526?pr=668

Now you can do:
```
// constructor will auto-detect c-strkey, or hex string
let c = new Contract("C...1234");

// default is changed to strkey output
console.log(c.contractId());
// C...1234

// old style hex output is via the new format parameter
console.log(c.contractId('hex'));
// abcd1234
```

### DIFF
--- a/src/contract.js
+++ b/src/contract.js
@@ -26,7 +26,7 @@ export class Contract {
     } catch (_) {
       // If that fails, try it as a hex string
       // TODO: Add methods based on the contractSpec (or do that elsewhere?)
-      let b = Buffer.from(contractId, 'hex');
+      const b = Buffer.from(contractId, 'hex');
       if (b.length !== 32) {
         throw new Error('Invalid contract ID');
       }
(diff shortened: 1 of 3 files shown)


src/contract.js
```diff
@@ -12,23 +13,42 @@ import xdr from './xdr';
...
   constructor(contractId) {
-    // TODO: Add methods based on the contractSpec (or do that elsewhere?)
-    this._id = Buffer.from(contractId, 'hex');
+    try {
+      // First, try it as a strkey
+      this._id = StrKey.decodeContract(contractId);
+    } catch (_) {
+      // If that fails, try it as a hex string
+      // TODO: Add methods based on the contractSpec (or do that elsewhere?)
+      let b = Buffer.from(contractId, 'hex');
```
tsachiherman, line 29: 'b' is never reassigned. Use 'const' instead

tsachiherman (commented):
just a small request..

tsachiherman (approved):

Issue #618, opened 2023-06-05
js-stellar-base: xdr: update xdr to support Auth & single host invoke function


tsachiherman:
### What ? 

xdr: update xdr to support Auth & single host invoke function

### Epic

https://github.com/stellar/soroban-tools/issues/657

PR #619, opened 2023-06-05
update xdr to support single function invocation

tsachiherman:
closes https://github.com/stellar/js-stellar-base/issues/618


Shaptic:
Updating XDR is, unfortunately, much more involved than just running `make reset-xdr` with the new hash. Thank you for the commit reference to stellar-xdr, though!

Issue #610, opened 2023-05-19
Make dealing with large `ScVal` integers easier.


Shaptic:
**Is your feature request related to a problem? Please describe.**
Dealing with `xdr.Int128Parts`, `xdr.Hyper`, etc. is a pain in the ass. Nobody outside of this library should have to deal with it.

**Describe the solution you'd like**
Abstractions around these values. I'd like something like:

```ts
type int64_t = number | bigint | string | xdr.Int64;
class I128 {
  constructor(lo: int128_t, hi?: int128_t);
  toNumber(): number; // throws if out of range
  toBigInt(): bigint;
  toHyper(): xdr.Hyper;
  toUnsignedHyper(): xdr.UnsignedHyper;
  toI128(): ScVal; // with ScValType = I128
  toU128(): ScVal; // with ScValType = U128
}
```

and support for this value when building transactions / operations / contract invocations.

**Describe alternatives you've considered**
Pain & suffering.

**Additional context**
A ton of work has already been done by @orbitlens here: stellar/js-xdr#96 as a starting point.

Shaptic:
@piyalbasu @aristidesstaffieri @paulbellamy @sreuland, I've drafted a proposal in the [`bignumber-wrapper`](https://github.com/stellar/js-stellar-base/tree/bignumber-wrapper) branch that combines @orbitlens' work in https://github.com/stellar/js-xdr/pull/96 with a high-level wrapper oriented towards `ScVal`s. The details are in [`src/numbers/generic.js`](https://github.com/stellar/js-stellar-base/blob/bignumber-wrapper/src/numbers/generic.js) (and you can see more usage in [`test/unit/i256_test.js`](https://github.com/stellar/js-stellar-base/blob/bignumber-wrapper/test/unit/i256_test.js)), but I've also replicated the jsdoc example below for ease of reading on how this can be used by downstream systems:

```js
import * as sdk from "stellar-base";

// You have an ScVal from a contract and want to parse it into JS native.
const value = sdk.xdr.ScVal.fromXDR(someXdr, "base64");
let bigi = sdk.BigInteger.fromScVal(value);

// You have a number and want to shove it into a contract.
bigi = sdk.BigInteger([1, 2, 3, 4]);
bigi.toBigInt() // returns BigInt((4 << 96) | (3 << 64) | (2 << 32) | 1)
bigi.toNumber() // will throw: too large

// Pass any to e.g. a Contract.call(), conversion happens automatically.
[...]

PR #620, opened 2023-06-05
Add wrappers to easily deal with the myriad of `ScVal` integer types.

Shaptic:
There's some irrelevant `yarn fmt` changes in here that can be ignored. The main thing to look at is `src/numbers/*.js` and the corresponding tests. Context: #610. This adds a new API and a helper function:

```ts
export type ScIntType =
  | 'i64'
  | 'u64'
  | 'i128'
  | 'u128'
  | 'i256'
  | 'u256';

export class XdrLargeInt {
  constructor(
    type: ScIntType,
    values: number|bigint|string|Array<number|bigint|string>
  );

  toNumber(): number;
  toBigInt(): bigint;

  toI64(): xdr.ScVal;
  toU64(): xdr.ScVal;
  toI128(): xdr.ScVal;
  toU128(): xdr.ScVal;
  toI256(): xdr.ScVal;
  toU256(): xdr.ScVal;

  valueOf(): any; // FIXME
  toString(): string;
  toJSON(): {
    value: string;
    type: ScIntType;
  };
}

export class ScInt extends XdrLargeInt {
  constructor(value: number|bigint|string|ScInt, opts?: { type: ScIntType });
}

export function scValToBigInt(scv: xdr.ScVal): bigint;
```

Their functionality should be relatively self-explanatory, but you can also refer to the docstrings for details.


types/index.d.ts
```diff
@@ -1053,3 +1053,40 @@ export function decodeAddressToMuxedAccount(address: string, supportMuxing: bool
...
+  | 'u64'
+  | 'i128'
+  | 'u128'
+  | 'i256'
+  | 'u256';
+
+export class XdrInt {
+  constructor(
+    type: ScIntType,
+    values: number|bigint|string|ScInt|Array<number|bigint|string|ScInt>
```
Shaptic, line 1068: Oops, you can't pass a `ScInt` to an `XdrInt` :facepalm: 

```suggestion
    values: number|bigint|string|Array<number|bigint|string>
```

src/numbers/scint.js
```diff
@@ -0,0 +1,118 @@
...
+ * // Or reinterpret it as a different type (size permitting):
+ * const scv = i.toI64();
+ * ```
+ *
+ * @param {number|bigint|string|ScInt} value - a single, integer-like value
+ *    which will be interpreted in the smallest appropriate XDR type supported
+ *    by Stellar (64, 128, or 256 bit integer values). signed values are
+ *    supported, though they are sanity-checked against `opts.type`. if you need
+ *    32-bit values, you can construct them directly without needing this
+ *    wrapper, e.g. `xdr.ScVal.scvU32(1234)`.
```
paulbellamy, line 65: Suggestion: Would it be convenient to also support 32-bit ints here as well? You can make them directly, but it might be more convenient to just have one way to do all ints. Fewer exceptions to remember.

src/numbers/xdrint.js
```diff
@@ -0,0 +1,246 @@
...
+
+  /**
+   * Transforms an opaque {@link xdr.ScVal} into a native BigInt, if possible.
+   *
+   * You can then give this back to create an {@link ScInt} instance, but the
+   * rationale here is that the native type is more likely to be immediately
+   * useful.
+   *
+   * @param {xdr.ScVal} scv - the raw XDR value to parse into an integer
+   * @returns {bigint} the integer value, regardless of size (even 32-bit)
```
paulbellamy, line 23: Shouldn't `XdrInt.fromScVal(scv)` return an `XdrInt`?

src/numbers/xdrint.js
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  /**
+   * @returns {number}
+   * @throws {RangeError} if the value can't fit into a Number
+   */
+  toNumber() {
+    const bi = this.int.toBigInt();
+    if (bi > Number.MAX_SAFE_INTEGER || bi < Number.MIN_SAFE_INTEGER) {
+      throw RangeError(`value ${bi} too large for Number`);
```
paulbellamy, line 113: Maybe two errors, one for too large, one for too small?

src/numbers/xdrint.js
paulbellamy: Seems like this could just do the switch to determine the `type`, then call into the constructor...

paulbellamy (approved):

package.json
```diff
@@ -120,9 +120,10 @@
   "dependencies": {
     "base32.js": "^0.1.0",
     "bignumber.js": "^9.1.1",
+    "buffer": "^6.0.3",
     "crc": "^4.3.2",
     "crypto-browserify": "^3.12.0",
-    "js-xdr": "^2.0.0",
+    "js-xdr": "git+https://github.com/stellar/js-xdr#ad2e43e",
```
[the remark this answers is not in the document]
Shaptic: Mind peeking at stellar/js-xdr#100? :pray:

src/numbers/uint128.js
```diff
@@ -0,0 +1,17 @@
+import { LargeInt } from 'js-xdr';
+
+export class Uint128 extends LargeInt {
+  constructor(...args) {
```
sreuland, line 4: this will inherit the same constructor from base class, and skip re-declaring here in the child classes..I think.

src/numbers/scint.js
Shaptic: It's a little obnoxious, unfortunately, because then the underlying `.int` is not a `jsXdr.LargeInt` type, which provides the essential `.size`, `.toBigInt()`, etc. methods. We could upscale it to an `Int64`, but that feels worse than ignoring the parameter to me. Wdyt?

src/numbers/xdrint.js
Shaptic: You're right that this can be refactored now that `XdrInt` exists! I did that, and also moved it outside of the class and renamed it to `convertScVal` to be more explicit about its purpose.

The reason why it doesn't return `XdrInt` is to support 32-bit ints. It's not perfect, I know, but because of https://github.com/stellar/js-stellar-base/pull/620#discussion_r1223408413, we can't wrap it all up into one interface.

src/numbers/xdrint.js
Shaptic: Fixed it by including the range itself in the error! ee42e30

src/numbers/scint.js
```diff
@@ -0,0 +1,361 @@
...
+/**
+ * Provides an easier way to manipulate large numbers for Stellar operations.
+ *
+ * You can instantiate this value either from bigints, strings, or numbers
+ * (whole numbers, or this will throw).
+ *
+ * If you need to create a native BigInt from a list of integer "parts" (for
+ * example, you have a series of encoded 32-bit integers that represent a larger
+ * value), you should use the specific XDR type, like {@link U128}. For example,
+ * you could do `new U128(values...).toBigInt()`.
```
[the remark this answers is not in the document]
Shaptic: Amended in 6f902efeff7e6513010c80fde429c6fe2003ff7c

src/numbers/xdr_int.js
```diff
@@ -0,0 +1,251 @@
...
+ * new ScInt(bigi);           // if you don't care about types, and
+ * new XdrInt('i128', bigi);  // if you do
+ * ```
+ *
+ * @param {xdr.ScVal} scv - the raw XDR value to parse into an integer
+ * @returns {bigint} the native value of this input value
+ *
+ * @throws {TypeError} if the `scv` input value doesn't represent an integer
+ */
+export function scValToInt(scv) {
```
sreuland, line 221: should this be refactored to `XdrInt.fromScVal()`, to be symmetric with the `XdrInt.toXYZ()` methods that return ScVal. Noticed usage is referenced in ScInt docs - https://github.com/stellar/js-stellar-base/pull/620/files#diff-b8cf857162926efd3588586c0e321ff3a60ef479b2ab7e9482d7e3c9a3ea82b5R22 and in the index.d.ts

sreuland (approved):
nice work, left some minor comments for consideration, my approval lighter weight in this area, would look to others from discussion on https://github.com/stellar/js-stellar-base/issues/610.

src/numbers/uint128.js
piyalbasu: Yup, this should inherit the base class constructor automatically.

But doesn't hurt to double check that this still works if you remove it 😅

src/numbers/xdr_int.js
Shaptic: Hah, this is the exact discussion from above, here: https://github.com/stellar/js-stellar-base/pull/620#discussion_r1223413318. I decided to move it out because it returns a `bigint` not an `XdrInt`. The thought was that it should be separate because 32-bit support is wonky and users are likelier to want the native value. Wdyt?

paulbellamy (approved):

src/numbers/xdr_int.js
sreuland: ah, ok, see that aspect of 64 or larger now, similar to the suggested rename for `scValToBigInt` function, a few adjustments for consideration:

* `XdrLargeInt` as the class name? I think it may help sdk users quickly get the intent, otherwise, 'int' types with no other qualifier in the type name usually hold anything up to a limit. Or add some docs at the class level to drive the point of only 64 bit and up.

* relocate `scValToBigInt` as it's not related to the class, like up to numbers/index.js or something?

Shaptic: good suggestions, done :+1:

src/numbers/uint128.js
Shaptic: yep I like that it's explicit bc it will document, added jsdocs to make this even more obvious to users

PR #621, opened 2023-06-06
remove the dependency on crc by calculating checksum in strkey

silence48:
remove the dependency on crc by calculating checksum in strkey
I didn't love that it used a different version of buffer that wasn't 6.0.3 plus we only needed the one checksum function so why include it all.
I returned a `Uint8Array` but it could use a `Buffer` instead, but that will go into a buffer and the tests passed so i think it's fine.


Shaptic (approved):
This is so awesome, every effort towards stellar/js-stellar-sdk#836 helps!

One request: can you `rm yarn.lock && yarn` so that we drop `crc` and get fresh dependencies in the lockfile?

silence48:
> This is so awesome, every effort towards [stellar/js-stellar-sdk#836](https://github.com/stellar/js-stellar-sdk/issues/836) helps!
> 
> One request: can you `rm yarn.lock && yarn` so that we drop `crc` and get fresh dependencies in the lockfile?

Yep i'll do it now.  You might wanna checkout the bufferless branch on my fork, while not all tests are passing yet, it eliminates buffer polyfill altogether.

silence48:
There's the updated yarn.lock thank you!

PR #624, opened 2023-06-13
remove lodash

silence48:
Remove the lodash dependency.


silence48:
this is still WIP because my tests are failing sorry.

silence48:
Should be ready now, well all the tests pass at least.

src/operations/create_claimable_balance.js
silence48, line 1: is claimants supposed to be an object? it was one here.

Shaptic (commented):
Ugh this is so awesome, thank you for investing into this effort! I think it just needs a `yarn fmt` pass and it'll be good to go!

silence48:
> Ugh this is so awesome, thank you for investing into this effort! I think it just needs a `yarn fmt` pass and it'll be good to go!

Notta problem happy to help. I did notice yarn run fmt fails for me. strange because when i run the script manually it works perfectly fine.  Added the format.

Shaptic (approved):

PR #632, opened 2023-06-27
Prototype for Soroban auth-next signing

paulbellamy:
Inspired by @aristidesstaffieri big mega-thread asking how to sign ContractAuth in a transaction... As well as @overcat's nice python sdk signing (https://github.com/StellarCN/py-stellar-base/blob/soroban/stellar_sdk/soroban/contract_auth.py)

This is a rough proposal for how to add support for soroban auth-next contract-auth signing to the sdk.

Usage:
```ts
// The "figure it out" mode, where you pass all the keys, and it figures out which one
// should sign which contract auths. Side effect here is that all keys will also sign the
// txn (which may or may not be desired).
txn.sign(key1, key2)

// Alternatively, if you have access to the ContractAuth components you could sign them
// individually. We would probably do this in js-soroban-client's `prepareTransaction` method:
function prepareTransaction(txn, key) {
  // simulate it to get the auths
  op.auths.forEach(auth => {
    if (auth.canSign(key)) {
      auth.sign(key)
    } else {
      // We need another key to sign this. Figure out what to do here. Panic?
    }
  })
  // sign the whole txn
  tx.sign(key)
  // submit & send
}
```

Obviously, it needs updating for preview 10, as well as some tests written, and types updated.

Thoughts?

### DIFF
--- a/src/contract_auth.js
+++ b/src/contract_auth.js
@@ -1,0 +1,22 @@
+import { ContractAuth, HashIdPreimageContractAuth } from './xdr';
+
+ContractAuth.prototype.canSign = (keypair) => {
+  const address = this.addressWithNonce()?.address();
+  if (!address || address.switch() !== 'scAddressTypeAccount') return false;
+  return (
+    keypair.publicKey() ===
+    StellarSdk.StrKey.encodeEd25519PublicKey(address.accountId().ed25519())
+  );
+};
+
+ContractAuth.prototype.sign = (keypair, networkId) => {
+  const a = this.addressWithNonce();
+  const sig = keypair.sign(
+    new HashIdPreimageContractAuth({
+      networkId,
+      nonce: a.nonce(),
+      invocation: this.rootInvocation()
+    }).toXDR('raw')
+  );
+  this.signatureArgs.push(sig);
+};
--- a/src/index.js
+++ b/src/index.js
@@ -4,6 +4,7 @@ import xdr from './xdr';
 
 BigNumber.DEBUG = true; // gives us exceptions on bad constructor values
 
+import './contract_auth';
 export { xdr };
 export { hash } from './hashing';
 export { sign, verify, FastSigning } from './signing';
--- a/src/transaction_base.js
+++ b/src/transaction_base.js
@@ -63,23 +63,33 @@ export class TransactionBase {
     this._networkPassphrase = networkPassphrase;
   }
 
+  get networkId() {
+    return hash(this._networkPassphrase);
+  }
+
   /**
    * Signs the transaction with the given {@link Keypair}.
    * @param {...Keypair} keypairs Keypairs of signers
    * @returns {void}
    */
   sign(...keypairs) {
-    // Temporary warning for contract auth-next signatures not being supported.
-    const requiresContractSignatures = (this.operations || []).some(
-      (op) =>
-        op.type === 'invokeHostFunction' &&
-        op.functions.some((fn) => fn.auth().some((a) => a.addressWithNonce()))
-    );
-    if (requiresContractSignatures) {
-      throw new Error(
-        'Soroban contract signatures are not supported in this version of the SDK.'
-      );
-    }
+    // Sign all the invokeHostFunction auths where we have a matching keypair.
+    // TODO: Before sending, we should check that we have signed all the required auths.
+    const networkId = this.networkId;
+    (this.operations || []).forEach((op) => {
+      if (op.type !== 'invokeHostFunction') return;
+      op.auth().forEach((auth) => {
+        // Check if this needs signing
+        const expectedSigner = auth.addressWithNonce()?.address()?.accountId();
+        if (!expectedSigner) return;
+        // Check if we have the matching keypair to sign with
+        const signer = keypairs.find((kp) => auth.canSign(kp));
+        if (!signer) return;
+        // Sign it
+        auth.sign(signer, networkId);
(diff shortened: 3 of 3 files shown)


lmorgan824:
In option #1, would this produce a lot of "bad_auth" failures for too many signatures on the transaction?
In option #2, What if multiple signatures are required for the transaction?  What if the SC needs multiple signers/auths?

I like a more manual style of signing since I know who has to sign where.  Since prepareTransaction gives us the auths object, I would just use that to sign auths, then sign tx, and submit as/when needed.

We use remote signing in many cases, and signing may happen over time.  (Channel accounts)  I wouldn't want to necessarily have to have all the signing keys available when preparing the transaction.

aristidesstaffieri:
Nice, this is great. My thoughts -

For the "figure it out" mode, if I understand correctly, this would only work when all signers need to sign both the tx/op and the contract auth?

In my example case(atomic swap), we have the need to "pass a transaction around" to different signers that want to participate in the swap, so I think it would be nice to have 2 helpers -

`getContractAuthPreImageEnvelope` - to get an xdr representation of the `HashIDPreimange::ENVELOPE_TYPE_CONTRACT_AUTH`

...then user would sign that hash, and use the following....

`addAuthNextSignature` - to take signed `HashIDPreimange::ENVELOPE_TYPE_CONTRACT_AUTH` , find the correct contract auth from the tx, and adds the correct xdr structure to the `signatureArgs` of that `ContractAuth`.

I'm still working on an example script, but one way I matched `ContractAuth` to signature was by [matching account IDs in `buildContractAuth`](https://github.com/stellar/js-soroban-client/blob/f5ed95d36e73f96f7f1f9f0e2cedc31c97b25cad/src/transaction.ts#L100), like this - 

```
const contractAuth = SorobanClient.xdr.ContractAuth.fromXDR(authStr, 'base64')
// First signer does not need an address or sig, can default to the tx source account
// If auth address matches any in key to sig map, add it to auth
if (contractAuth.addressWithNonce()) {
  const authAccount = contractAuth.addressWithNonce().address().accountId().toXDR('hex')

  if (pubKeyToSig[authAccount]) {
    contractAuth.signatureArgs(pubKeyToSig[authAccount])
  }
}
```

overcat:
I also agree with the idea of users manually signing ContractAuth. In my opinion, signing ContractAuth and signing a transaction should be two completely independent steps. When a user signs ContractAuth, they do not need to be concerned with the contents of the transaction.

paulbellamy:
Sounds good. Ok, so the "figure-it-out" mode of `txn.sign(keypairs)` is out then. It seems like the `ContractAuth.sign(keypair)` and `ContractAuth.canSign(keypair)` might still be useful, but needs a better way to fit into the workflow.

overcat:
> actually, @overcat do you have an example workflow of how a user would do this in the python SDK? Maybe we can copy that pattern here for consistency.

Please check the following exampes: 
https://github.com/StellarCN/py-stellar-base/blob/soroban/examples/soroban_auth_atomic_swap.py#L100-L105
https://github.com/StellarCN/py-stellar-base/blob/soroban/examples/soroban_auth_with_stellar_account.py
https://github.com/StellarCN/py-stellar-base/blob/soroban/examples/soroban_auth_with_transaction_invoke.py

lmorgan824:
I have this code adopted from @overcat's py

https://github.com/blocktimefinancial/option/blob/main/samples/contractAuth.js

Not tested!!!

Shaptic (commented):
I like the idea of attaching this to the auth object (though I dunno if you can `.prototype` it since the xdr objects aren't "real" classes [thanks js-xdr]), but here's an update for how I think this works in Preview 10:

```ts
import {
  Address,
  Keypair,
  hash,
  nativeToScVal,
  xdr,
} from "stellar-base";

import { randomInt } from "crypto";

function authorizeInvocation(
  signer: Keypair,
  networkPassphrase: string,
  invocation: xdr.SorobanAuthorizedInvocation,
  ledgerValidityCount: number,
): xdr.SorobanAuthorizationEntry {
  const networkId = hash(Buffer.from(networkPassphrase));
  const nonce = randomInt(1, Number.MAX_SAFE_INTEGER); // don't actually do this lmao

  const envelope = new xdr.HashIdPreimageSorobanAuthorization({
    networkId,
    invocation,
    nonce: new xdr.Uint64(nonce),
    signatureExpirationLedger: ledgerValidityCount,
  });

  const sig = signer.sign(hash(envelope.toXDR('raw')));

  return new xdr.SorobanAuthorizationEntry({
    credentials: xdr.SorobanCredentials.sorobanCredentialsAddress(
      new xdr.SorobanAddressCredentials({
        address: new Address(signer.publicKey()).toScAddress(),
        nonce: envelope.nonce(),
        signatureExpirationLedger: envelope.signatureExpirationLedger(),
        signatureArgs: [nativeToScVal({
          "public_key": signer.rawPublicKey(),
          "signature": sig,
        })],
      })
    ),
    rootInvocation: invocation,
  })
}
```

overcat:
I have written a prototype for the Python version, which may not be perfect, but can serve as a reference.
https://github.com/StellarCN/py-stellar-base/blob/48b55f32fa2eb8a5418afae48a51f3234936287e/stellar_sdk/soroban/authorization_entry.py

Shaptic:
Closing, see [src/auth.js](https://github.com/stellar/js-stellar-base/blob/master/src/auth.js).

Issue #637, opened 2023-06-30
Improve abstractions between `ScVal`s and native types


Shaptic:
While #628 makes progress on this for basic conversions (see #630 and #620), there is more to be done.

For example, 

```js
nativeToScVal("a string")
```

will always return an `ScVal.scvString`. However, it should be possible to have it return an `scvSymbol` (or even `scvBytes`, arguably). The inverse is also true, e.g. an `ScVal.scvSymbol` could be a buffer or a string. There are other irregularities as well. 

-----------

Here are some notes from @willemneal about allow type specifications for the conversions:

https://github.com/stellar/js-stellar-base/issues/628#issuecomment-1608088245:

> One thing to note. Scvmap could be a rust struct, which would be better represented as a js object (TS interface). I had initially gone with Map, but it's a much worse DevX experience than just accessing a field. So perhaps the conversion interface should also allow passing the corresponding ScTypeDef.

https://github.com/stellar/js-stellar-base/issues/628#issuecomment-1609769231:

> It would be similar to the CLI: [`soroban-spec-tools/src/lib.rs`](https://github.com/stellar/soroban-tools/blob/d16a84c2bbceccec67c3fa06a6d8f6f294a055bc/cmd/crates/soroban-spec-tools/src/lib.rs#L486)
> 
> The tricky part is figuring out the corresponding the ScTypeDef. This could be done by the bindings code. For example adding it here: [`soroban-spec-typescript/fixtures/ts/src/index.ts`](https://github.com/stellar/soroban-tools/blob/d16a84c2bbceccec67c3fa06a6d8f6f294a055bc/cmd/crates/soroban-spec-typescript/fixtures/ts/src/index.ts#L235)

-----------------

#### Acceptance Criteria

- [x] enable converting `string` to Symbol
- [x] enable converting `Buffer|Uint8Array` to String or Symbol
- [ ] attach the conversions to the `xdr.ScVal` type
- [ ] design (not necessarily implement) a way to construct objects/interfaces from a `scvMap`

Shaptic:
@willemneal so I'm realizing that enforcing an interface type is basically impossible in this library because `stellar-base` is pure JavaScript, not TypeScript. So, really, the best we can do is return a key-value mapping as we already do here, then you can `as YourInterface` (or whatever) in your higher-level library.

The current implementation will give you a very reasonable structure, I think, as a starting point.

willemneal:
The as casting will fail if you try to cast a map as an interface.  Eg

[...]

PR #638, opened 2023-06-30
Improve type flexibility when building `ScVal`s from native types

Shaptic:
This enhanced the `nativeToScVal` helper, allowing more flexibility in the output types.

Specifically, this lets you:

 * build `scvU32` and `scvI32` types
 * build `scvSymbol`s from strings and bytes
 * build `scvString`s from bytes
 * build `scvVec`s with specific types (i.e. recursive type specification)
 * build `scvMap`s with specific types for both keys and values (via a simple type specification format (e.g. `type: { field: [ keyType?, valueType? ] }` for each entry in the map)

Related: #637


src/numbers/index.js
```diff
@@ -31,7 +28,7 @@ export { XdrLargeInt };
  * @throws {TypeError} if the `scv` input value doesn't represent an integer
  */
 export function scValToBigInt(scv) {
-  const type = scv.switch().name.slice(3).toLowerCase();
+  const type = scv.switch().name.slice(3).toLowerCase(); // if it ain't broke...
```
sreuland, line 31: can scval.js encapsulate how to obtain scval type as `getType(xdr.ScVal)`, the `slice(3)` part seems really low level aspect which this conversion routine probably doesn't need to know about?

Shaptic: I disagree that this needs to be pulled out since it isn't being reused anywhere, but I do see what you're getting at from a self-documenting code perspective. Done in 06b9adec, take a look!

src/scval.js
```diff
@@ -82,9 +107,35 @@ import { ScInt, scValToBigInt } from './numbers/index';
  *    types, custom classes)
  *  - the type of the input object (or some inner value of said object) cannot
  *    be determined (via `typeof`)
+ *  - the type you specified (via `opts.type`) is incompatible with the value
```
sreuland, line 110: it only throws for vals that have sub-types that actually check on 'type' , non-sub-types like address, contract, object, will just ignore, not sure if this inconsistency is important.

Shaptic: Great point! Done in 2b1b010b.

src/scval.js
```diff
@@ -67,12 +71,33 @@ import { ScInt, scValToBigInt } from './numbers/index';
...
+ * `toScVal("a string", {type: "i256"})` will throw.
  *
- * @param {any} val         a native (or convertible) input value to wrap
- * @param {object} [opts]   an optional set of options to pass which allows you
- *      to specify a type when the native `val` is an integer-like (i.e.
- *      number|bigint) type
+ * @param {any} val -       a native (or convertible) input value to wrap
+ * @param {object} [opts] - an optional set of hints around the type of
+ *    conversion you'd like to see
+ * @param {string} [opts.type] - there is different behavior for different input
```
sreuland, line 79: this may get to be an uber-method to maintain all type permutations for native types in one method. As of now, should docs explicitly show a matrix of valid native type to opt.types?  or callers need to look at code.

just for consideration later, ways to cordon options directly to their native types, like a TypeConverter class with `toScVal/fromScVal` or functional like passing converter function instead like `nativeToScVal(val, converterFn)`. I think recall seeing in past reviews that baking this type of conversions into generated ScVal from xdr was not reasonable.

sreuland (approved):

src/numbers/index.js
sreuland: looks good!

src/scval.js
Shaptic: I agree that there's a risk here of permutation blow-up, BUT I'm hoping that this can stay relatively lightweight. The matrix is sparse, so to speak, in that many types have one or two possibilities at most. So hopefully it doesn't get ugly. (In fact, you could even argue that some of the permutations are the fault of design mistakes on the XDR level, like `ScVal.scvString`s taking a `Buffer`, but I digress...)

With the discussion on custom, user-defined types in #637, you very well may be right that this can end up exploding into an uber method... so let's cross that bridge if we get to it? I don't want to over-invest a lot of architectural thought into this until we get community feedback on whether or not this is even a useful abstraction to provide in the first place. I think of this as a "minimum high-level implementation" to try to help downstream folks not deal with XDR so much, more-so than a perfect v0 solution.

PR #658, opened 2023-07-27
feat!: drop support for hex contract IDs

### LINKED ISSUE #650: Drop support for hex-encoded contract addresses
Hex contract addresses were a "v0" way to express a unique contract identifier. We have moved on to using Stellar's strkey format (see [SEP-23](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0023.md)) and have deprecated support for the hex format.

This format should be removed entirely as part of the testnet upgrade, because anything deprecated prior to a "v1.0.0 release" should disappear. People need time and indication to upgrade their code to use the strkey format.

This may have corresponding work in stellar/js-soroban-client.

chadoh:
Fixes #650

### DIFF
--- a/src/contract.js
+++ b/src/contract.js
@@ -14,9 +14,7 @@ import { StrKey } from './strkey';
  * @constructor
  *
  * @param {string} contractId - ID of the contract (ex.
- *     `CA3D5KRYM6CB7OWQ6TWYRR3Z4T7GNZLKERYNZGGA5SOAOPIFY6YQGAXE`, or as a
- *     32-byte hex string
- *     `000000000000000000000000000000000000000000000000000000000000000001`).
+ *     `CA3D5KRYM6CB7OWQ6TWYRR3Z4T7GNZLKERYNZGGA5SOAOPIFY6YQGAXE`).
  */
 // TODO: Support contract deployment, maybe?
 export class Contract {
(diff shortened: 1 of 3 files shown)


test/unit/contract_test.js
```diff
@@ -1,49 +1,34 @@
...
-      let contract = new StellarBase.Contract(contractId);
-      expect(contract.contractId('strkey')).to.equal(contractId);
-    });
-
-    it('parses hex addresses', function () {
-      let contractId = '0'.repeat(63) + '1';
-      let contract = new StellarBase.Contract(contractId);
-      expect(contract.contractId('hex')).to.equal(contractId);
-    });
+const NULL_ADDRESS = 'CAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAD2KM'
```
chadoh, line 1: I got this by using the current version of stellar-base and:

```
let contractId = '0'.repeat(63) + '1';
let contract = new StellarBase.Contract(contractId);
contract.address('strkey')
```

src/contract.js
```diff
@@ -25,30 +23,20 @@ export class Contract {
...
-    switch (format) {
-      case 'strkey':
-        return StrKey.encodeContract(this._id);
-      case 'hex':
-        return this._id.toString('hex');
-      default:
-        throw new Error(`Invalid format: ${format}`);
+  contractId() {
+    try {
+      return StrKey.encodeContract(this._id);
```
Shaptic, line 37: there's no reason why this should fail after construction succeeds, I think...?

test/unit/contract_test.js
```diff
@@ -7,21 +9,6 @@ describe('Contract', function () {
       expect(contract.contractId('strkey')).to.equal(contractId);
     });
 
-    it('parses hex addresses', function () {
```
Shaptic, line 10: let's make this `throws on deprecated hex addresses`, instead?

Shaptic (commented):
You also need to update `types/index.d.ts` signatures :pray: otherwise this is almost ready

test/unit/contract_test.js
chadoh: Sure. I'll make it say "obsolete", since "deprecated" means "still technically supported, but will be removed soon."

chadoh: See below

src/contract.js
chadoh: Yeah, I wasn't sure. Better to start without it and only add it later if we know it's useful, since it seems probably-not-useful.

chadoh:
Needing to manually update `*.d.ts` files 😩 

All feedback addressed.

Shaptic (approved):
awesome, thank you! :clap: mind adding a breaking change entry to #661 so we have it documented?

chadoh:
I don't have the ability to edit the description there, if that's what you mean

Shaptic:
That is what I meant - no worries! I'll do it

PR #660, opened 2023-07-27
Add `TransactionBuilder.setFootprint()` method to prepare footprint easily

Shaptic:
This is much better than creating an entire, mostly-empty `xdr.SorobanTransactionData` and using `setSorobanData()` when you want to create a bump or restore operation.


src/resources.js
```diff
@@ -0,0 +1,23 @@
+import xdr from './xdr';
+
+export function newSorobanData() {
+  return sorobanDataFromFootprint([], []);
+}
+
+export function sorobanDataFromFootprint(readOnly, readWrite) {
```
aristidesstaffieri, line 7: so for my use case I think I needed to keep the rest of the resources, so I set it like this 
```
const sorobanTxData = xdr.SorobanTransactionData.fromXDR(
    simulation.transactionData,
    "base64",
  );
  sorobanTxData.resources().footprint(footprint);
  txnBuilder.setSorobanData(sorobanTxData);
```

so basically I merged a new simulation's resources with an old footprint, would I still be able to do that if `sorobanDataFromFootprint` 0s out the other keys of the SorobanResources?

Shaptic: yeah so you wouldn't want to use these. instead, you'd want this:

```js
txnBuilder
  .setSorobanData(sorobanTxData)
  .setFootprint(footprint.readOnly(), footprint.readWrite());
```
for this: the latter call will leave the existing fields alone.

src/transaction_builder.js
```diff
@@ -459,6 +460,48 @@ export class TransactionBuilder {
...
+   *    passed, the field is left untouched (so if you want to clear it, pass an
+   *    empty array)
+   * @param {xdr.LedgerKey[] | null} readWrite  the set of ledger keys to set in
+   *    the read-write portion of the transaction's `sorobanData`. if null is
+   *    passed, the field is left untouched (so if you want to clear it, pass an
+   *    empty array)
+   *
+   * @returns {TransactionBuilder}
+   *
+   * @warning If the `sorobanData` has NOT been set (e.g. via
```
sreuland, line 483: not suggesting change, but seeing two different methods on builder that overwrite same state is not ideal, another option may be to go with builder pattern like   `SorobanTransactionDataBuilder` in resources.js, and then no changes in transaction_builder, external client code looks something like:
```
// tx data builder creates from nothing or an existing xdr instance as initial state
SorobanTransactionDataBuilder dataBuilder = new SorobanTransactionDataBuilder(null | xdr.SorobanTransactionData);
dataBuilder.setReadOnlyFootprint(roFootPrint);
dataBuilder.setReadWriteFootprint(rwFootPrint);
sorobanTxData = dataBuilder.build()

TransactionBuilder tb = new TransactionBuilder();
...
tb.setSorobanData(sorobanTxData)
...
// something triggers change on footprints, want to rebuild the tx
dataBuilder = new SorobanTransactionDataBuilder(tb.sorobanData);
dataBuilder.setReadOnlyFootprint(newRoFootPrint);
tb.setSorobanData(dataBuilder.build())
newerTx = tb.build()
```

Shaptic: >two different methods on builder that overwrite same state is not ideal 

There is precedent with `.setTimebounds()` and `.setTimeout()`, heh.

Your design is quite elegant, though! I like it. My only concern is that there are probably not very many other things you'd want to set with this pattern besides the footprint (hence why the `.setFootprint()` subset isn't so bad), since e.g. the fees nor extension should almost certainly never be set directly. I like it, though! So it's done in 46ab05be9d225d5c3b3e8cd815f667362f08b52b.

src/resources.js
Shaptic: After the refactor based on @sreuland's comment, this would now be:

```js
txnBuilder.setSorobanData(
  new SorobanDataBuilder(sorobanTxData)
    .setReadOnly(footprint.readOnly(), 
    .setReadWrite(footprint.readWrite())
    .build()
);

sreuland (approved):

PR #662, opened 2023-07-28
Expand contract footprint with contract code ledger key.

Shaptic:
Add the correct `xdr.LedgerKeyContractCode` instance to the contract's default access footprint.

### DIFF
--- a/src/contract.js
+++ b/src/contract.js
@@ -36,6 +36,10 @@ export class Contract {
     return StrKey.encodeContract(this._id);
   }
 
+  toString() {
+    return this.contractId();
+  }
+
   /**
    * Returns the address of this contract as an Address type.
    * @returns {Address}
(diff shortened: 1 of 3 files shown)


src/contract.js
```diff
@@ -36,6 +36,10 @@ export class Contract {
     return StrKey.encodeContract(this._id);
   }
 
+  toString() {
```
sreuland, line 39: what prompted the need to add this, should it have js-doc and be included in [Contract typedef](https://github.com/stellar/js-stellar-base/blob/better-contract-footprint/types/index.d.ts#L28) ?

src/contract.js
```diff
@@ -64,20 +68,29 @@ export class Contract {
...
   /**
-   * Returns the read-only footprint entry necessary for any invocations to this
-   * contract, for convenience when adding it to your transaction's overall
-   * footprint.
+   * Returns the read-only footprint entries necessary for any invocations to
+   * this contract, for convenience when adding it to your transaction's overall
+   * footprint or doing bump/restore operations.
    *
-   * @returns {xdr.LedgerKey} the contract's executable data ledger key
+   * @returns {xdr.LedgerKey[]} the ledger keys containing the contract's code
```
sreuland, line 75: the typedef for [Contract.getFootPrint()](https://github.com/stellar/js-stellar-base/blob/better-contract-footprint/types/index.d.ts#L33) should reflect this different return type ?

sreuland (approved):

src/contract.js
Shaptic: ugh yes thank you so much

Issue #575, opened 2023-02-14
Soroban: Support multi-party transaction signing


paulbellamy:
With auth-next in soroban, transactions can require signatures for the `invokeHostFunction` arguments from multiple addresses. (In addition to the normal submission signatures).

We should figure out the workflow to support this, and add support.

One option would be to take the multiple keypairs as part of `txn.sign(keypair)`, parse through the operations, finding any where this address is the relevant signer, and signing those as well. However, this conflates the txn signer with the argument signer, which may not be true in all cases. We could add a new `txn.signContractAuth(keypair)`? Needs some design thinking.

Shaptic:
Related: #632

Shaptic:
Closed, at least in part, by #663

PR #663, opened 2023-07-28
Add helpers method to build authorization entries.

Shaptic:
Replaces #632 in an attempt to tackle #575. This adds three helpers (see the jsdocs for details):

```typescript
function authorizeInvocation(
  signer: Keypair,
  networkPassphrase: string,
  validUntil: number,
  invocation: xdr.SorobanAuthorizedInvocation
): xdr.SorobanAuthorizationEntry;

function buildAuthEnvelope(
  networkPassphrase: string,
  validUntil: number,
  invocation: xdr.SorobanAuthorizedInvocation
): xdr.HashIdPreimage;

function buildAuthEntry(
  envelope: xdr.HashIdPreimage,
  signature: Buffer | Uint8Array,
  publicKey: string
): xdr.SorobanAuthorizationEntry;
```

### DIFF
--- a/src/auth.js
+++ b/src/auth.js
@@ -48,15 +48,6 @@ export function authorizeInvocation(
 }
 
 /**
- * An asynchronous callback to sign an input buffer.
- *
- * @async
- * @callback signingCallback
- * @param {Uint8Array} input  the raw buffer to sign
- * @returns {Uint8Array} the signature on the input buffer
- */
-
-/**
  * This works like {@link authorizeInvocation}, but allows passing an
  * asynchronous callback as a "signing method" (e.g. {@link Keypair.sign}) and a
  * public key instead of a specific {@link Keypair}.
--- a/src/auth.js
+++ b/src/auth.js
@@ -6,17 +6,25 @@ import { nativeToScVal } from './scval';
 import xdr from './xdr';
 
 /**
- * Builds an {@link xdr.SorobanAuthorizationEntry} that indicates to
+ * This builds an authorization entry that indicates to
  * {@link Operation.invokeHostFunction} that a particular identity (i.e. signing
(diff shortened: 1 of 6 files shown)


src/auth.js
```diff
@@ -0,0 +1,159 @@
...
+ */
+export function authorizeInvocation(
+  signer,
+  networkPassphrase,
+  validUntil,
+  invocation
+) {
+  const preimage = buildAuthEnvelope(networkPassphrase, validUntil, invocation);
+  const input = hash(preimage.toXDR());
+  const signature = signer.sign(input);
```
aristidesstaffieri, line 46: We talked about this a bit but I wanted to ask - is there any reason to not support the async case here? I believe if we used an async signature and changed from taking a `signer` to taking a `async(input) => ...sign input and return...` then both sync/async cases would be supported. Something like this -

```
export async function authorizeInvocation(
  signHash,
  networkPassphrase,
  validUntil,
  invocation
) {
  const preimage = buildAuthEnvelope(networkPassphrase, validUntil, invocation);
  const input = hash(preimage.toXDR());
  const signature = await signHash(input);
  ... rest of it ...
```

aristidesstaffieri (approved):

src/auth.js
aristidesstaffieri: and simple keypair sync signing case can just pass in a `(input) => signer.sign(input)`

Shaptic: This method is cleaner because with the callback you need to pass a public key, heh, but I added `authorizeInvocationCallback` for this purpose!

```typescript

export function authorizeInvocationCallback(
  publicKey: string,
  signingMethod: (input: Buffer) => Buffer,
  networkPassphrase: string,
  validUntil: number,
  invocation: xdr.SorobanAuthorizedInvocation
): xdr.SorobanAuthorizationEntry;
```

PR #664, opened 2023-08-04
Upgrades the Soroban XDR to the latest version.

### LINKED ISSUE #655: Upgrade the Soroban XDR to the latest version

Shaptic:
Closes #655.

### DIFF
--- a/src/auth.js
+++ b/src/auth.js
@@ -1,6 +1,5 @@
 import xdr from './xdr';
 
-import { StrKey } from './strkey';
 import { Keypair } from './keypair';
 import { hash } from './hashing';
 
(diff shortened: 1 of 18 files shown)


src/auth.js
```diff
@@ -170,22 +169,7 @@ export function buildAuthEntry(envelope, signature, publicKey) {
...
-              // force the keys to be interpreted as symbols (expected for
-              // Soroban [contracttype]s)
-              type: {
-                public_key: ['symbol', null],
-                signature: ['symbol', null]
-              }
-            }
-          )
-        ]
+        signature: nativeToScVal(signature)
```
Shaptic, line 172: TODO: Confirm that this is the right thing to do ([Slack thread](https://stellarfoundation.slack.com/archives/C030Z9EHVQE/p1691174074687949)).

Shaptic: narrator: "it wasn't"

test/unit/auth_test.js
```diff
@@ -28,16 +28,41 @@ describe('building authorization entries', function () {
     );
 
     let cred = entry.credentials().address();
-    let args = cred.signatureArgs().map((v) => StellarBase.scValToNative(v));
-
     expect(cred.signatureExpirationLedger()).to.equal(123);
-    expect(args.length).to.equal(1);
+    expect(entry.rootInvocation()).to.eql(invocation);
+
+    // sanity check raw xdr signature types: should be ScVal{
```
sreuland, line 34: not suggesting any change, since mostly defined the expected outcome as xdr js object here, would using that as `expect(sig).to.equal(expectedXdrSig)` be a substitute instead of multiple lower level verification checks, and then the code becomes the docs rather than maintaining over time, which is great that it's here in any case for those less in tune like myself.

sreuland (approved):
nice work!

PR #667, opened 2023-08-08
Add utilities for formatting token amounts (`formatTokenAmount` and `parseTokenAmount`)

jeesunikim:
**Summary:** 
- adding soroban token display formatter and parser from 
- moving the commonly used functions [parseTokenAmount](https://github.com/stellar/soroban-react-payment/blob/904ecebbccb0ec05587999117e28f46bf0b9c0db/src/helpers/soroban.ts#L58) and [formatTokenAmount](https://github.com/stellar/soroban-react-payment/blob/904ecebbccb0ec05587999117e28f46bf0b9c0db/src/helpers/format.ts#L27) from `soroban-react-payment`

### DIFF
--- a/config/.eslintrc.js
+++ b/config/.eslintrc.js
@@ -6,9 +6,6 @@ module.exports = {
   extends: ['airbnb-base', 'prettier'],
   plugins: ['@babel', 'prettier', 'prefer-import'],
   parser: '@babel/eslint-parser',
-  globals: {
-    BigInt: true
-  },
   rules: {
     // OFF
     'import/prefer-default-export': 0,
--- a/config/.eslintrc.js
+++ b/config/.eslintrc.js
@@ -6,6 +6,9 @@ module.exports = {
   extends: ['airbnb-base', 'prettier'],
   plugins: ['@babel', 'prettier', 'prefer-import'],
   parser: '@babel/eslint-parser',
+  globals: {
+    BigInt: true
+  },
   rules: {
     // OFF
     'import/prefer-default-export': 0,
--- a/src/soroban.js
+++ b/src/soroban.js
@@ -52,9 +52,7 @@ export class Soroban {
     }
 
     // Trim trailing zeros
-    while (fraction[fraction.length - 1] === '0') {
-      fraction = fraction.substring(0, fraction.length - 1);
-    }
+    fraction = fraction.replace(/0+$/, '');
 
     // If decimals is 0, we have an empty string for fraction
     if (fraction === '') {
--- a/src/soroban.js
+++ b/src/soroban.js
@@ -43,7 +43,8 @@ export class Soroban {
    *
    * @example
    * const displayValueAmount = "123.4560"
-   * const parsedAmountForSmartContract = StellarBase.Soroban.parseTokenAmount(displayValueAmount, 5); // "12345600"
+   * const parsedAmountForSmartContract = parseTokenAmount("123.4560", 5);
+   * parsedAmountForSmartContract === "12345600"
    */
   static parseTokenAmount(value, decimals) {
     let [whole, fraction, ...rest] = value.split('.').slice();
--- a/src/soroban.js
+++ b/src/soroban.js
@@ -1,5 +1,3 @@
-import BigNumber from 'bignumber.js';
-
 /**
  * Soroban helper class
  * formatting, parsing, and etc
@@ -7,25 +5,27 @@ import BigNumber from 'bignumber.js';
  */
 export class Soroban {
   /**
-   * format token amount to its ideal specific number of decimals
+   * All arithmetic inside the contract is performed on integers to
+   * avoid potential precision and consistency issues of floating-point
    *
-   * @param {BigNumber} amount - the token amount you want to display
-   * @param {number} decimals
-   * @returns {string}
+   * This function takes the smart contract value and its decimals (if the token has any) and returns a display value
+   * @param {bigint | number | string} amount - the token amount you want to display
+   * @param {number} decimals - specify how many decimal places a token has
+   * @returns {string} - display value
    */
   static formatTokenAmount(amount, decimals) {
     let formatted = amount.toString();
 
     if (decimals > 0) {
-      formatted = amount.shiftedBy(-decimals).toFixed(decimals).toString();
+      formatted = (amount / 10 ** decimals).toFixed(decimals).toString();
 
       // Trim trailing zeros
       while (formatted[formatted.length - 1] === '0') {
-        formatted = formatted.substring(0, formatted.length - 1);
+        formatted = formatted.replace(/0+$/, '');
       }
 
       if (formatted.endsWith('.')) {
-        formatted = formatted.substring(0, formatted.length - 1);
+        formatted = formatted.replace(/0+$/, '');
       }
     }
 
@@ -35,15 +35,23 @@ export class Soroban {
   /**
    * parse token amount to use it on smart contract
    *
+   * * This function takes the display value and its decimals (if the token has any) and returns an integer that'll be used it on smart contract
    * @param {string} value - the token amount you want to use it on smart contract
-   * @param {number} decimals
-   * @returns {string}
+   * @param {number} decimals - specify how many decimal places a token has
+   * @returns {string} - smart contract value
+   *
+   *
+   * @example
+   * const displayValueAmount = "123.4560"
+   * const parsedAmountForSmartContract = StellarBase.Soroban.parseTokenAmount(displayValueAmount, 5); // "12345600"
    */
   static parseTokenAmount(value, decimals) {
-    const comps = value.split('.');
+    let [whole, fraction, ...rest] = value.split('.').slice();
 
-    let whole = comps[0];
-    let fraction = comps[1];
+    if (rest.length) {
+      throw new Error('Invalid decimal');
+    }
+
     if (!whole) {
       whole = '0';
     }
@@ -64,9 +72,10 @@ export class Soroban {
       fraction += '0';
     }
 
-    const wholeValue = new BigNumber(whole);
-    const fractionValue = new BigNumber(fraction);
+    const shifted = BigInt(
+      whole + (fraction?.padEnd(decimals, '0') ?? '0'.repeat(decimals))
+    );
 
-    return wholeValue.shiftedBy(decimals).plus(fractionValue);
+    return shifted.toString();
   }
 }
--- a/src/soroban.js
+++ b/src/soroban.js
@@ -22,11 +22,9 @@ export class Soroban {
 
     if (decimals > 0) {
       if (decimals > formatted.length) {
-        const targetShiftedLength = decimals - formatted.length;
-        const targetTotalLength = targetShiftedLength + formatted.length;
         formatted = [
           '0',
-          formatted.toString().padStart(targetTotalLength, '0')
+          formatted.toString().padStart(decimals, '0')
         ].join('.');
       } else {
         formatted = [
--- a/src/soroban.js
+++ b/src/soroban.js
@@ -1,4 +1,4 @@
-import BigNumber from "bignumber.js";
+import BigNumber from 'bignumber.js';
 
 /**
  * Soroban helper class
@@ -20,11 +20,11 @@ export class Soroban {
       formatted = amount.shiftedBy(-decimals).toFixed(decimals).toString();
 
       // Trim trailing zeros
-      while (formatted[formatted.length - 1] === "0") {
+      while (formatted[formatted.length - 1] === '0') {
         formatted = formatted.substring(0, formatted.length - 1);
       }
 
-      if (formatted.endsWith(".")) {
+      if (formatted.endsWith('.')) {
         formatted = formatted.substring(0, formatted.length - 1);
       }
     }
@@ -40,30 +40,30 @@ export class Soroban {
    * @returns {string}
    */
   static parseTokenAmount(value, decimals) {
-    const comps = value.split(".");
+    const comps = value.split('.');
 
     let whole = comps[0];
     let fraction = comps[1];
     if (!whole) {
-      whole = "0";
+      whole = '0';
     }
     if (!fraction) {
-      fraction = "0";
+      fraction = '0';
     }
 
     // Trim trailing zeros
-    while (fraction[fraction.length - 1] === "0") {
+    while (fraction[fraction.length - 1] === '0') {
       fraction = fraction.substring(0, fraction.length - 1);
     }
 
     // If decimals is 0, we have an empty string for fraction
-    if (fraction === "") {
-      fraction = "0";
+    if (fraction === '') {
+      fraction = '0';
     }
 
     // Fully pad the string with zeros to get to value
     while (fraction.length < decimals) {
-      fraction += "0";
+      fraction += '0';
     }
 
     const wholeValue = new BigNumber(whole);
--- a/src/soroban.js
+++ b/src/soroban.js
@@ -21,10 +21,19 @@ export class Soroban {
     }
 
     if (decimals > 0) {
-      formatted = [
-        formatted.slice(0, -decimals),
-        formatted.slice(-decimals)
-      ].join('.');
+      if (decimals > formatted.length) {
+        const targetShiftedLength = decimals - formatted.length;
+        const targetTotalLength = targetShiftedLength + formatted.length;
+        formatted = [
+          '0',
+          formatted.toString().padStart(targetTotalLength, '0')
+        ].join('.');
+      } else {
+        formatted = [
+          formatted.slice(0, -decimals),
+          formatted.slice(-decimals)
+        ].join('.');
+      }
     }
 
     // remove trailing zero if any
@@ -34,7 +43,7 @@ export class Soroban {
   /**
    * parse token amount to use it on smart contract
    *
-   * This function takes the display value and its decimals (if the token has 
+   * This function takes the display value and its decimals (if the token has
    * any) and returns a string that'll be used within the smart contract.
    * @param {string} value - the token amount you want to use it on smart contract
    * @param {number} decimals - specify how many decimal places a token has
--- a/src/soroban.js
+++ b/src/soroban.js
@@ -1,0 +1,74 @@
+import BigNumber from "bignumber.js";
+
+/**
+ * Soroban helper class
+ * formatting, parsing, and etc
+ * @class Soroban
+ */
+export class Soroban {
+  /**
+   * format token amount to its ideal specific number of decimals
+   *
+   * @param {BigNumber} amount - the token amount you want to display
+   * @param {number} decimals
+   * @returns {string}
+   */
+  static formatTokenAmount(amount, decimals) {
+    let formatted = amount.toString();
+
+    if (decimals > 0) {
+      formatted = amount.shiftedBy(-decimals).toFixed(decimals).toString();
+
+      // Trim trailing zeros
+      while (formatted[formatted.length - 1] === "0") {
+        formatted = formatted.substring(0, formatted.length - 1);
+      }
+
+      if (formatted.endsWith(".")) {
+        formatted = formatted.substring(0, formatted.length - 1);
+      }
+    }
+
+    return formatted;
+  }
+
+  /**
+   * parse token amount to use it on smart contract
+   *
+   * @param {string} value - the token amount you want to use it on smart contract
+   * @param {number} decimals
+   * @returns {string}
+   */
+  static parseTokenAmount(value, decimals) {
+    const comps = value.split(".");
+
+    let whole = comps[0];
+    let fraction = comps[1];
+    if (!whole) {
+      whole = "0";
+    }
+    if (!fraction) {
+      fraction = "0";
+    }
+
+    // Trim trailing zeros
+    while (fraction[fraction.length - 1] === "0") {
+      fraction = fraction.substring(0, fraction.length - 1);
+    }
+
+    // If decimals is 0, we have an empty string for fraction
+    if (fraction === "") {
+      fraction = "0";
+    }
+
+    // Fully pad the string with zeros to get to value
+    while (fraction.length < decimals) {
+      fraction += "0";
+    }
+
+    const wholeValue = new BigNumber(whole);
+    const fractionValue = new BigNumber(fraction);
+
+    return wholeValue.shiftedBy(decimals).plus(fractionValue);
+  }
+}
--- a/src/soroban.js
+++ b/src/soroban.js
@@ -9,19 +9,26 @@ export class Soroban {
    * avoid potential precision and consistency issues of floating-point
    *
    * This function takes the smart contract value and its decimals (if the token has any) and returns a display value
-   * @param {bigint | number | string} amount - the token amount you want to display
+   * @param {string} amount - the token amount you want to display
    * @param {number} decimals - specify how many decimal places a token has
    * @returns {string} - display value
    */
   static formatTokenAmount(amount, decimals) {
-    let formatted = amount.toString();
+    let formatted = amount;
 
-    if (decimals > 0) {
-      formatted = (amount / 10 ** decimals).toFixed(decimals);
-      formatted = formatted.replace(/\.?0+$/, '');
+    if (amount.includes('.')) {
+      throw new Error('No decimal is allowed');
     }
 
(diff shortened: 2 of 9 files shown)


src/soroban.js
```diff
@@ -0,0 +1,74 @@
...
+  static formatTokenAmount(amount, decimals) {
+    let formatted = amount.toString();
+
+    if (decimals > 0) {
+      formatted = amount.shiftedBy(-decimals).toFixed(decimals).toString();
+
+      // Trim trailing zeros
+      while (formatted[formatted.length - 1] === '0') {
+        formatted = formatted.substring(0, formatted.length - 1);
+      }
```
Shaptic, line 25:
```js
formatted = formatted.replace(/0+$/, '');
```
would be simpler

src/soroban.js
```diff
@@ -0,0 +1,74 @@
...
+ * Soroban helper class
+ * formatting, parsing, and etc
+ * @class Soroban
+ */
+export class Soroban {
+  /**
+   * format token amount to its ideal specific number of decimals
+   *
+   * @param {BigNumber} amount - the token amount you want to display
+   * @param {number} decimals
```
Shaptic, line 13: needs a description

Shaptic: should it default to 7?

src/soroban.js
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    return formatted;
+  }
+
+  /**
+   * parse token amount to use it on smart contract
+   *
+   * @param {string} value - the token amount you want to use it on smart contract
+   * @param {number} decimals
```
Shaptic, line 39: again, needs same description

src/soroban.js
```diff
@@ -0,0 +1,74 @@
...
+   * @returns {string}
+   */
+  static parseTokenAmount(value, decimals) {
+    const comps = value.split('.');
+
+    let whole = comps[0];
+    let fraction = comps[1];
```
Shaptic, line 46:
```js
const [ whole, fraction ] = value.split('.').slice(2);
```
I think this should work?

src/soroban.js
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * Soroban helper class
+ * formatting, parsing, and etc
+ * @class Soroban
+ */
+export class Soroban {
+  /**
+   * format token amount to its ideal specific number of decimals
+   *
+   * @param {BigNumber} amount - the token amount you want to display
```
Shaptic, line 12: can we make this `bigint | number | string`, instead?

src/soroban.js
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  /**
+   * parse token amount to use it on smart contract
+   *
+   * @param {string} value - the token amount you want to use it on smart contract
+   * @param {number} decimals
+   * @returns {string}
+   */
+  static parseTokenAmount(value, decimals) {
```
Shaptic, line 42: at least one `@example` would be fantastic

src/soroban.js
```diff
@@ -0,0 +1,74 @@
...
+
+    // Fully pad the string with zeros to get to value
+    while (fraction.length < decimals) {
+      fraction += '0';
+    }
+
+    const wholeValue = new BigNumber(whole);
+    const fractionValue = new BigNumber(fraction);
+
+    return wholeValue.shiftedBy(decimals).plus(fractionValue);
```
Shaptic, line 72: Can we do this with `bigint` somehow, instead? Like `(wholeValue / decimals) + fractionValue`? Again, trying to lower dependence on BigNumber.js.

src/soroban.js
aristidesstaffieri: IMHO we should not provide a default here. Even though we might be able to guess that folks will need 7 the most often I think having a default makes it easier to overlook the arg and make a mistake.

src/soroban.js
aristidesstaffieri: that works but without the slice

Shaptic: yep that was a safety check in case it was > 2 items

src/soroban.js
Shaptic: yep, fair!

src/soroban.js
Shaptic: we should have an assertion that `.` is in the string, then: this code :boom:s if you just pass a whole number, right?

test/unit/soroban_test.js
```diff
@@ -0,0 +1,38 @@
...
+
+    it("returns '100000.0001' for '10000000010' of a token set to 5 decimals", function () {
+      const amount = new BigNumber(10000000010);
+
+      expect(StellarBase.Soroban.formatTokenAmount(amount, 5)).to.equal(
+        '100000.0001'
+      );
+    });
+
+    it("returns '10000.00001' for '1000000001.'.' of a token set to 5 decimals", function () {
```
Shaptic, line 21: nit: typo?
```suggestion
    it("returns '10000.00001' for '1000000001.' of a token set to 5 decimals", function () {
```

test/unit/soroban_test.js
```diff
@@ -0,0 +1,38 @@
...
+      expect(StellarBase.Soroban.formatTokenAmount(amount, 5)).to.equal(
+        '10000.00001'
+      );
+    });
+  });
+
+  describe('parseTokenAmount', function () {
+    it("returns '10000' for '100' of a token set to 2 decimals", function () {
+      const amount = '100';
+      const parsedAmount = StellarBase.Soroban.parseTokenAmount(amount, 2);
```
Shaptic, line 33: wait, why doesn't this :boom: if `amount` doesn't have a `.` to `split('.')` on?

Shaptic: oh because it's `comps[1] === undefined` in the current code..? then `fraction = '0'`. gotcha

Shaptic: okay so we need a test mixing a fractional number and decimals value, e.g. `parseTokenAmount("100.123", 5)`

src/soroban.js
aristidesstaffieri: it doesn't work though
```
value = '10.01'
'10.01'
const [ whole, fraction ] = value.split('.').slice(2);
undefined
whole
undefined
```

It doesn't actually blow up on whole numbers, the fractional piece that handles the whole number case is below this block.

aristidesstaffieri:
```
if (fraction === "") {
    fraction = "0";
  }
```

aristidesstaffieri: could do `const [ whole, fraction ] = value.split('.').slice(0, 2);` but this doesn't need any extra checks as is.

Shaptic: oh wtf :facepalm: that's what I meant re: `slice(0, 2)`

aristidesstaffieri: figured as much 😄

src/soroban.js
```diff
@@ -0,0 +1,72 @@
...
+      if (formatted.endsWith('.')) {
+        formatted = formatted.substring(0, formatted.length - 1);
+      }
+    }
+
+    return formatted;
+  }
+
+  /**
+   * parse token amount to use it on smart contract
```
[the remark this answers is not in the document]
Shaptic: oh derp the return value is a whole number not a fraction, so it would return `10012300`, right?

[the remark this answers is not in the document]
aristidesstaffieri: yup yup

Shaptic:
I'm confused about why CI isn't running :raised_eyebrow:

src/soroban.js
Shaptic: even if we need to use BigNumber _internally_, let's make them pass in a native value

aristidesstaffieri: doesn't it need to be a BigInt or BigNumber since the number can be bigger than MAX_INT?

src/soroban.js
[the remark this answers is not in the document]
Shaptic: hmm, I think that means we can do `bigint(whole) * decimal` + `bigint(frac_padded_with_decimal_zeroes)`?? :thinking:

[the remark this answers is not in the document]
aristidesstaffieri: oh do you mean to do that instead of this?
`wholeValue.shiftedBy(decimals).plus(fractionValue);`

Hmm does `bigint(whole) * decimal` == `wholeValue.shiftedBy(decimals).`? I think it would be something like `bigint(whole) * (decimal * 10)` in order to move the decimal by the number of places defined by the decimal?

src/soroban.js
Shaptic: I'm saying we should have these be the signatures:

```typescript
function parseTokenAmount(value: string | number | bigint, decimals: number): string;
function formatTokenAmount(value: string, decimals: number): bigint;
```

Then, you can `let bigNum = BigNumber(value)` internally to work with it.

If they want to parse a floating point value that exceeds the limits of `number`, they can use a `string` (even if that means `BigNumber(value).toString()`, but then it's _their_ dependency, not ours :laughing:)

src/soroban.js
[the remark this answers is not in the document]
Shaptic: oh yeah that's what I meant hehe nice catch - I'm doing off-the-cuff napkin math

src/soroban.js
aristidesstaffieri: ahhhh ya I see I see. Thanks for explaining, yeah that sounds like an easy way to avoid the dep.

src/soroban.js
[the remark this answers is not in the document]
aristidesstaffieri: 👍 I think that checks out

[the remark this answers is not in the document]
Shaptic: wait LOL isn't this `whole + part.padEnd(decimals, '0')`??? I just tried it:

```javascript
// inputs 
let value = "123.4560";
let decimals = 5;

// algo
let [ whole, part ] = value.split('.').slice(0, 2);
let shifted = BigInt(whole + part.padEnd(decimals, '0'));
console.log(shifted.toString())

// "12345600"
```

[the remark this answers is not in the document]
aristidesstaffieri: hmm yeah I think that does work, but still squinting at that nullish coalesce.
It looks to me like `shifted` could == `undefined00` for the whole number case because the coalesce is happening on `part?.padEnd(decimals, '0')` and not just on `part`, right?

aristidesstaffieri:
btw [the original inspiration for these is here](https://github.com/ethers-io/ethers.js/blob/master/packages/bignumber/src.ts/fixednumber.ts#L70) and they do some extra stuff that we don't like work with negative values so maybe worth discussing? Ours have worked for all of our use cases so far but they may get used differently as part of the sdk.

src/soroban.js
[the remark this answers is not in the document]
Shaptic: but `part?.padEnd` does not call `padEnd` if `part === null | undefined`, so it hits the right side of the `??` case, right?

Shaptic:
@aristidesstaffieri thanks for the ref! we should steal their (non-negative) test cases, then. I think the usage aligns (besides `multiplier`, which I don't understand).

Btw, a lot of the code comments in this review I'd make there, too, but I bet theirs was written before e.g. `.padEnd` or `??` support was common in JS.

src/soroban.js
Shaptic: closing since we're discussing below

src/soroban.js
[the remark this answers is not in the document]
aristidesstaffieri: ah yup, I see. Thanks!

src/soroban.js
aristidesstaffieri: sorry to bring this one up again but wondering if we should be throwing if 
```
const [ whole, fraction, ..rest ] = value.split('.').slice();
if (rest.length) {.. throw?...}
```

src/soroban.js
jeesunikim: hmm but we use `BigNumber` for calculation in `parseTokenAmount`
`const wholeValue = new BigNumber(whole);
const fractionValue = new BigNumber(fraction);

return wholeValue.shiftedBy(decimals).plus(fractionValue);`

jeesunikim: deleted my comment because I see the other comment mentioning it

aristidesstaffieri:
> I'm confused about why CI isn't running 🤨

Github is having issues, probably related https://www.githubstatus.com/incidents/nwb8m3ppfxw6

src/soroban.js
Shaptic: Yeah for sure, that's just flat-out an invalid decimal

config/.eslintrc.js
```diff
@@ -5,6 +5,9 @@ module.exports = {
   extends: ['airbnb-base', 'prettier'],
   plugins: ['@babel', 'prettier', 'prefer-import'],
   parser: '@babel/eslint-parser',
+  globals: {
+    BigInt: true
```
jeesunikim, line 9: added this @Shaptic

src/soroban.js
```diff
@@ -0,0 +1,81 @@
...
+   * This function takes the smart contract value and its decimals (if the token has any) and returns a display value
+   * @param {bigint | number | string} amount - the token amount you want to display
+   * @param {number} decimals - specify how many decimal places a token has
+   * @returns {string} - display value
+   */
+  static formatTokenAmount(amount, decimals) {
+    let formatted = amount.toString();
+
+    if (decimals > 0) {
+      formatted = (amount / 10 ** decimals).toFixed(decimals).toString();
```
jeesunikim, line 20: thoughts? @Shaptic

src/soroban.js
```diff
@@ -0,0 +1,81 @@
...
+   *
+   * * This function takes the display value and its decimals (if the token has any) and returns an integer that'll be used it on smart contract
+   * @param {string} value - the token amount you want to use it on smart contract
+   * @param {number} decimals - specify how many decimal places a token has
+   * @returns {string} - smart contract value
+   *
+   *
+   * @example
+   * const displayValueAmount = "123.4560"
+   * const parsedAmountForSmartContract = StellarBase.Soroban.parseTokenAmount(displayValueAmount, 5); // "12345600"
```
jeesunikim, line 46: is this example okay? @Shaptic

src/soroban.js
```diff
@@ -0,0 +1,81 @@
...
+      formatted = (amount / 10 ** decimals).toFixed(decimals).toString();
+
+      // Trim trailing zeros
+      while (formatted[formatted.length - 1] === '0') {
+        formatted = formatted.replace(/0+$/, '');
+      }
+
+      if (formatted.endsWith('.')) {
+        formatted = formatted.replace(/0+$/, '');
+      }
```
Shaptic, line 29: You don't need the loop, and I think the second clause doesn't cut off the period? I **think** this whole thing can be a simple one-liner:

```js
formatted = formatted.replace(/\.?0+$/, '');
```

test/unit/soroban_test.js
```diff
@@ -0,0 +1,43 @@
...
+      );
+    });
+
+    it("returns '10000.00001' for '1000000001.' of a token set to 5 decimals", function () {
+      const amount = '1000000001.';
+
+      expect(StellarBase.Soroban.formatTokenAmount(amount, 5)).to.equal(
+        '10000.00001'
+      );
+    });
```
Shaptic, line 25: these tests would be more readable and extensible as a loop over an array of test cases, like

```js
[
  [ 1000000001, 7, '100.0000000' ],
  // etc.
].forEach(([ amount, decimals, expected ]) => {
});
```

src/soroban.js
```diff
@@ -0,0 +1,81 @@
...
+   *
+   * @example
+   * const displayValueAmount = "123.4560"
+   * const parsedAmountForSmartContract = StellarBase.Soroban.parseTokenAmount(displayValueAmount, 5); // "12345600"
+   */
+  static parseTokenAmount(value, decimals) {
+    let [whole, fraction, ...rest] = value.split('.').slice();
+
+    if (rest.length) {
+      throw new Error('Invalid decimal');
```
Shaptic, line 52: suggestion: more error details
```suggestion
      throw new Error(`Invalid decimal value: ${value}`);
```

src/soroban.js
```diff
@@ -0,0 +1,81 @@
...
+
+    // If decimals is 0, we have an empty string for fraction
+    if (fraction === '') {
+      fraction = '0';
+    }
+
+    // Fully pad the string with zeros to get to value
+    while (fraction.length < decimals) {
+      fraction += '0';
+    }
```
Shaptic, line 73: trim zeroes then replace them? :confused: :raised_eyebrow:

I don't think we need this if we have L76 with the `.padEnd` call

src/soroban.js
```diff
@@ -0,0 +1,81 @@
...
+    if (rest.length) {
+      throw new Error('Invalid decimal');
+    }
+
+    if (!whole) {
+      whole = '0';
+    }
+    if (!fraction) {
+      fraction = '0';
+    }
```
Shaptic, line 60: I think these are handled implicitly by the shift calculation

src/soroban.js
Shaptic: yeah, great! minor suggestion for readability

```suggestion
   * const parsedAmountForSmartContract = parseTokenAmount("123.4560", 5);
   * parsedAmountForSmartContract === "12345600"
```

src/soroban.js
Shaptic: This is an interesting approach, but I'm worried it won't preserve enough precision in the `number` case?

Separate question: is `amount` expected to be a whole number? 

Also, `.toFixed()` returns a string already.

config/.eslintrc.js
jeesunikim: I'll add `es2020: true` instead

src/soroban.js
jeesunikim: `amount` is expected to be a whole number since it's the value that's used on smart contract. Removed `.toString();`

Shaptic: should we validate that, then? e.g. `BigInt(amount)` will throw if amount is a decimal or checking `formatted.indexOf('.') === -1`

src/soroban.js
```diff
@@ -0,0 +1,54 @@
...
+   * This function takes the smart contract value and its decimals (if the token has any) and returns a display value
+   * @param {bigint | number | string} amount - the token amount you want to display
+   * @param {number} decimals - specify how many decimal places a token has
+   * @returns {string} - display value
+   */
+  static formatTokenAmount(amount, decimals) {
+    let formatted = amount.toString();
+
+    if (decimals > 0) {
+      formatted = (amount / 10 ** decimals).toFixed(decimals);
```
Shaptic, line 20: Thinking about this some more, I'm a bit worried about precision. @aristidesstaffieri, @jeesunikim what do you two think? Is it possible that we'll lose bits by dividing by such large numbers (e.g. `1 / 1e-10`)? Maybe we should string-manipulate here, too, like we do in the inverse function. I tried it out [with a runkit here](https://runkit.com/61833274873f8c000b42e673/64e3c3b97c07390008e92d45), basically relying on this:

```js
return [formatted.slice(0, -decimals), formatted.slice(-decimals)].join('.');
```

to figure out where to insert the `.`.

aristidesstaffieri: yeah 100% I think that should be the approach, if you look back at the original one we did start by shifting decimal places which I believe is just string manipulation on big numbers.

Shaptic: The thing that's a bit uncertain to me with this approach is what to do if (a) there are already `.`s in the input `amount` and (b) if there are zeroes at the end of the `amount` (I elaborate in the FIXME in the runkit), but maybe you guys have already considered these cases?

aristidesstaffieri: ah well I don't see why zeros at the end would be different than any number at the end? I think the number of decimals only tells you how many chars to move when placing the `.`. As far already being a decimal, before this we basically assume some sanity and don't check for malformed strings that can't become a formatted value, but I think we could check for existing dots, non numeric chars, and others things to throw in case it can't be formatted.

Shaptic: Yeah that makes sense to me. Maybe the ambiguity is clearer with an existing decimal:

```js
formatTokenAmount("1230.000", 5) === 12.3 or 0.0123
```

You see what I mean? But it sounds like you're saying just treat the input as-is, irrespective of no-impact trailing zeroes.

aristidesstaffieri: well I think actually a string with an existing decimal is not valid input for this function, because that would imply that it is already formatted. We never checked for this before but I think this should throw in that case. 

Otherwise I think for trailing zeros we could drop them but also not sure we want to assume that it will always be desired. I wonder if it's better to have an option that lets you toggle wether or not to trim zeros?

Shaptic: Yeah, I think we should throw on decimals. And imo we don't need an option until requested; trimming off trailing zeroes is a simple regex that can (probably should) be done by the caller.

aristidesstaffieri: sweet, works for me. Thanks for being diligent on that one.

test/unit/soroban_test.js
jeesunikim: i'm so noob here LOL

src/soroban.js
jeesunikim: thank you both. updated.

src/soroban.js
```diff
@@ -0,0 +1,61 @@
...
+
+    if (decimals > 0) {
+      formatted = [
+        formatted.slice(0, -decimals),
+        formatted.slice(-decimals)
+      ].join('.');
+    }
+
+    // remove trailing zero if any
+    return formatted.replace(/\.?0+$/, '');
```
Shaptic, line 31: I think this breaks with non-decimal zeroes due to the `?`, e.g. you should test

```js
formatTokenAmount("1230", 0)
```

I think this will give `123` :fearful: The fix is to require the decimal in the regex.

src/soroban.js
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  /**
+   * parse token amount to use it on smart contract
+   *
+   * * This function takes the display value and its decimals (if the token has any) and returns an integer that'll be used it on smart contract
```
Shaptic, line 37: nit: wrapping, typo

```suggestion
   * This function takes the display value and its decimals (if the token has 
   * any) and returns a string that'll be used within the smart contract.
```

Shaptic (commented):
One bug, I think, and I'd love to see a few more test cases so we can validate the behavior better! But I think this is almost ready besides that. (Also you need to run `yarn fmt` heh.)

test/unit/soroban_test.js
```diff
@@ -0,0 +1,82 @@
+describe("Soroban", function () {
+  describe("formatTokenAmount", function () {
+    const SUCCESS_TEST_CASES = [
+      { amount: "1000000001", decimals: 7, expected: "100.0000001" },
+      { amount: "10000000010", decimals: 5, expected: "100000.0001" },
+      { amount: "10000000010", decimals: 0, expected: "1000000001" },
```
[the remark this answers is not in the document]
jeesunikim: ooo good point. this failed, working on it.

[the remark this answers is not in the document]
jeesunikim: ahhh b/c of my trailing zero regex, kill me. thanks for catching it

src/soroban.js
```diff
@@ -0,0 +1,71 @@
...
+      throw new Error('No decimal is allowed');
+    }
+
+    if (decimals > 0) {
+      if (decimals > formatted.length) {
+        const targetShiftedLength = decimals - formatted.length;
+        const targetTotalLength = targetShiftedLength + formatted.length;
+        formatted = [
+          '0',
+          formatted.toString().padStart(targetTotalLength, '0')
```
jeesunikim, line 29: what ya think @Shaptic

src/soroban.js
jeesunikim: updated

src/soroban.js
Shaptic: hmm... I'm expanding this math and it sounds like: 

```js
targetTotalLength = (decimals - formatted.length) + formatted.length;
```

so that's just `decimals`? Am I missing something?

jeesunikim: lol... :homerhide:

jeesunikim: fixed

Shaptic (approved):
Woohoo, we made it!! Thanks for this :clap:

PR #668, opened 2023-08-08
Fixes missing / incorrect TypeScript descriptions.

Shaptic:
* Adds a missing method to `XdrLargeInt`s TypeScript definition
* Removes `ScInt` from constructor (it isn't handled at all)

### DIFF
--- a/src/numbers/sc_int.js
+++ b/src/numbers/sc_int.js
@@ -53,12 +53,12 @@ import { XdrLargeInt } from './xdr_large_int';
  * // Or reinterpret it as a different type (size permitting):
  * const scv = i.toI64();
  *
- * @param {number|bigint|string|ScInt} value - a single, integer-like value
- *    which will be interpreted in the smallest appropriate XDR type supported
- *    by Stellar (64, 128, or 256 bit integer values). signed values are
- *    supported, though they are sanity-checked against `opts.type`. if you need
- *    32-bit values, you can construct them directly without needing this
- *    wrapper, e.g. `xdr.ScVal.scvU32(1234)`.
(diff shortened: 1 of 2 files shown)


src/numbers/sc_int.js
```diff
@@ -53,12 +53,12 @@ import { XdrLargeInt } from './xdr_large_int';
  * // Or reinterpret it as a different type (size permitting):
  * const scv = i.toI64();
  *
- * @param {number|bigint|string|ScInt} value - a single, integer-like value
- *    which will be interpreted in the smallest appropriate XDR type supported
- *    by Stellar (64, 128, or 256 bit integer values). signed values are
- *    supported, though they are sanity-checked against `opts.type`. if you need
- *    32-bit values, you can construct them directly without needing this
- *    wrapper, e.g. `xdr.ScVal.scvU32(1234)`.
+ * @param {number|bigint|string} value - a single, integer-like value which will
```
sreuland, line 56: just for consideration, to assert these value types only are allowed, another unit test like 'it throws when invalid value' in [scint_test.js](https://github.com/stellar/js-stellar-base/blob/another-ts-bugfix/test/unit/scint_test.js#L254) with `StellarBase.ScInt({})`, would sweep up coverage for not allowing ScInt anymore.

sreuland (approved):

tsachiherman (approved):

src/numbers/sc_int.js
Shaptic: it's more like the jsdoc was wrong - `ScInt` was never actually supported

PR #670, opened 2023-08-10
Add ability to clear out operations from the transaction builder

Shaptic:
This suggestion comes from @sreuland after #656 to clear out operations from a builder to allow them to be built separately. This is particularly key for Soroban operations in configuring auth.

### DIFF
--- a/test/unit/transaction_builder_test.js
+++ b/test/unit/transaction_builder_test.js
@@ -947,6 +947,11 @@ describe('TransactionBuilder', function () {
         fee: '10000'
       }).build();
       expect(cloneTx.fee).to.equal('20000'); // double because two ops
+
+      cloneTx = StellarBase.TransactionBuilder.cloneFrom(tx)
+        .clearOperations()
+        .build();
+      expect(cloneTx.operations).to.be.empty;
     });
   });
 });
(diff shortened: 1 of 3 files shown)


test/unit/transaction_builder_test.js
```diff
@@ -80,23 +82,32 @@ describe('TransactionBuilder', function () {
...
+      return StellarBase.xdr.HostFunction.hostFunctionTypeInvokeContract(
+        new StellarBase.xdr.InvokeContractArgs({
+          contractAddress: new StellarBase.Address(contractId).toScAddress(),
+          functionName: 'hello',
+          args: []
+        })
+      );
+    };
+
     it('should set the soroban data from object', function (done) {
```
sreuland, line 103: should we add a test to verify `builder.clearOperations()` ?

sreuland (approved):
looks good! most of the code appears unrelated to clearing the builder state, mostly addressing a bump on xdr schema.

Shaptic:
omg thank you for catching that @sreuland, I was trying to keep xdr changes separate (see soroban-next branch) so that we could add bug-fixes/features to current futurenet. fixed!

PR #675, opened 2023-08-10
Auth Entry helpers - Adds arg for passing credential address nonce

aristidesstaffieri:
After trying to use the new helpers in [this branch of the swap dapp](https://github.com/stellar/soroban-react-atomic-swap/compare/feature/soroban-sdk-v0.10.0?expand=1#). I found that the nonce required in both `buildAuthEntry` and `buildAuthEnvelope` should be the nonce from the original auth entries credential address,[ like you see here](https://github.com/stellar/soroban-react-atomic-swap/blob/main/src/helpers/soroban.ts#L217).

This adds `credentialAddressNonce` to all of the helpers, updates tests/types.
Fixes the return type for `authorizeInvocationCallback` signing method arg, Buffer -> Promise<<Buffer>>.

### DIFF
--- a/src/auth.js
+++ b/src/auth.js
@@ -46,7 +46,7 @@ export function authorizeInvocation(
   const preimage = buildAuthEnvelope(networkPassphrase, validUntil, invocation, credentialAddressNonce);
   const input = hash(preimage.toXDR());
   const signature = signer.sign(input);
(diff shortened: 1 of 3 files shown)


Shaptic (commented):
I still want to prove that this is necessary :triumph:

aristidesstaffieri:
closing in favor of the changes in https://github.com/stellar/js-stellar-base/pull/678/files

PR #676, opened 2023-08-11
fixed lint warn on jsdoc

sreuland:
noticed the npm publish step was failing on prior commits to `soroban` due to a lint warning:

[build: src/transaction_builder.js#L231](https://github.com/stellar/js-stellar-base/commit/6d3a5d5958830673fea6ba8a007bcc9a8857be3a#annotation_13300108768)
Missing JSDoc @returns for function

and increased base's version to `10.0.0-soroban.7` as the `soroban` branch has been updated with latest interim xdr updates, and this will enable getting npm published?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "stellar-base",
-  "version": "10.0.0-soroban.6",
+  "version": "10.0.0-soroban.7",
   "description": "Low-level support library for the Stellar network.",
   "main": "./lib/index.js",
   "browser": {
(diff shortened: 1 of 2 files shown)


Shaptic:
The publish is failing because it automatically tries publishing whenever `soroban` gets a commit, and the version already exists because I'm not bumping the `soroban.<N>` on every PR so that we can have discrete, meaningful version bumps.

This is still a valid issue (it's just a warning though), but I want to get a few more PRs in before we bump to `soroban.7` (see https://github.com/stellar/js-stellar-base/pull/674).

sreuland:
ah ok, was looking to get js-soroban-client working in e2e tests, and that is pending on it getting latest xdr from base, ok will defer to https://github.com/stellar/js-stellar-base/pull/674, thanks!

PR #678, opened 2023-08-11
Provide a better authorization abstraction, signing entries in-place 

Shaptic:
**Note for reviewers:** I recommend reading [the code directly](https://github.com/stellar/js-stellar-base/blob/better-auth-helpers/src/auth.js) rather than looking through the diff because it is quite messy after removing the existing helpers in 873d0fe. It also might help to read the new test cases.

### What
This replaces the existing authorization helpers with a single, unified method:

```typescript
export async function authorizeEntry(
  entry: xdr.SorobanAuthorizationEntry,
  signer: 
    // or anything with a .sign(buf) => buf method
    | Keypair 
    // given a buffer, return its signature
    | (inputToSign: Buffer) => Promise<Buffer>,
  validUntilLedgerSeq: number,
  networkPassphrase: string = Networks.FUTURENET
);
```

You can reference the jsdoc in the PR itself for more details.

### Why
This redesign simplifies the way we can build entries, motivated by a quirk of Soroban: When simulating an invocation via Soroban RPC, the returned operation will **indiscriminately** have its `auth[]` array replaced with _fresh_ (i.e. new nonces) and _unsigned_ authorization entries. Originally, our intention was:
   - you can replace these entries entirely by just passing the root invocation to `authorizeInvocation`, i.e. something like `op.auth[i] = authorizeInvocation(op.auth[i].rootInvocation())`, in pseudocode
   - this let us control the entire entry from the invocation alone, and made it easier to build arbitrary entries w/o dependence on preflight

However, this is not ergonomic with preflight because:
 - when referencing a nonce, you need to ensure that nonce is also in the transaction footprint
 - we had no way to notify the user on how to adjust the footprint (drop entry for simulated nonce, add for new one)
 - so if you just replaced the entries, your tx would fail because the nonce wasn't in the footprint

We _could_ do something to adjust the footprint on submission in soroban-client (e.g. more magic [here](https://github.com/stellar/js-soroban-client/blob/main/src/transaction.ts#L88-L93), but this feels like a better, direct approach.

### Known Limitations
This does mean that it's much more difficult to create an entry "from scratch" (i.e. from an invocation alone), but that seems like an edge case at this time. Perhaps we should still keep the old helpers (e.g. `authorizeInvocation`) rather than replacing them, but it's unclear whether or not those are useful to anybody? Maybe the introduction of #669 would help this along.


paulbellamy:
You may find https://github.com/stellar/soroban-tools/blob/main/cmd/soroban-rpc/lib/preflight/src/preflight.rs#L50-L64 illuminating

src/auth.js
```diff
@@ -1,197 +1,161 @@
...
+ *        preppedTx.sign(source);
+ *        return server.sendTransaction(preppedTx);
+ *      });
+ * }
  */
-export function buildAuthEntry(envelope, signature, publicKey) {
-  // ensure this identity signed this envelope correctly
+export async function authorizeEntry(
+  entry,
+  signingMethod,
```
paulbellamy, line 95:
```suggestion
  signer,
```
Nit: naming suggestion. If this were a go interface we would probably call it:
```go
type Signer interface {
  Sign(blob []byte) Signature
}
```

src/auth.js
```diff
@@ -1,197 +1,161 @@
...
+  validUntil,
```
paulbellamy, line 96:
```suggestion
  validBeforeLedgerSeq,
```
Nit: naming suggestion. Or `validThroughLedgerSeq`, I never remember which it is.

src/auth.js
```diff
@@ -1,197 +1,161 @@
...
+  networkPassphrase = Networks.FUTURENET
+) {
+  // no-op
   if (
-    !Keypair.fromPublicKey(publicKey).verify(hash(envelope.toXDR()), signature)
+    entry.credentials().switch() !==
+    xdr.SorobanCredentialsType.sorobanCredentialsAddress()
   ) {
-    throw new Error(`signature does not match envelope or identity`);
+    return entry;
```
paulbellamy, line 104: It might be more useful for this to fail if you try to sign something that can't be signed. idk tho

src/auth.js
```diff
@@ -1,197 +1,161 @@
...
+  if (signingMethod instanceof Keypair) {
+    signature = signingMethod.sign(payload);
+    publicKey = signingMethod.publicKey();
+  } else {
+    const [sig, pk] = await signingMethod(preimage);
+    signature = Buffer.from(sig);
+    publicKey = pk;
+  }
+
+  const expectedPubkey = Address.fromScAddress(addrAuth.address()).toString();
```
paulbellamy, line 133: If we already know the pubkey, why require the `signingMethod` to return it? Any mismatch would be caught by the `verify()` below afaict.

The advantage of not requiring the `signingMethod` to return it, is that it lowers the barrier for implementing a custom `signingMethod`. e.g. fewer round trips to a hw wallet, potentially.

src/auth.js
```diff
@@ -1,197 +1,161 @@
...
+      nonce: addrAuth.nonce(),
+      invocation: entry.rootInvocation(),
+      signatureExpirationLedger: addrAuth.signatureExpirationLedger()
+    })
+  );
+  const payload = hash(preimage.toXDR());
+
+  let signature;
+  let publicKey;
+  if (signingMethod instanceof Keypair) {
```
paulbellamy, line 124: Side note: you could actually make this fn more general by switching the if/else blocks, like:
```js
let sign = signingMethod;
if (typeof signingMethod !== 'function') {
  // Would handle Keypair, or any object with a `.sign`
  // method, like: `{sign: (payload) => bytes}`
  sign = signingMethod.sign;
}
const signature = sign(payload);
// etc...

File: src/sorobandata_builder.js [hunk 6]
```diff
@@ -154,6 +154,21 @@ export class SorobanDataBuilder {
...
 
+  /**
+   * Adds a set of ledger keys to the existing access footprint.
+   *
+   * @param {xdr.LedgerKey[]} readOnly  read-only keys added to access footprint
+   * @param {xdr.LedgerKey[]} readWrite read-write keys added to access footprint
+   *
+   * @returns {SorobanDataBuilder}
+   */
+  appendFootprint(readOnly, readWrite) {
```
paulbellamy, line 165: Alternative would be to add a method to the `xdr.LedgerFootprint.append(other: {readOnly?: []xdr.LedgerKey, readWrite?: xdr.LedgerKey[]}|xdr.LedgerFootprint)`, which would merge two footprints.
`concat`? `merge`?

Not sure which is better.

paulbellamy (commented):
Seems like a useful handler. Some design suggestions below, and comment on how to use this to sign txns above.

src/auth.js
Shaptic: ooh great call, done!

src/auth.js
Shaptic: great point, idk why I didn't realize this

src/auth.js
Shaptic: I would say rtfm but who does that? Amended!

src/auth.js
Shaptic: sgtm!

src/auth.js
Shaptic: That would make it such that you can't do a `auth.map((entry) => authorizeEntry(entry, complexAsyncSigningMethod, ...)` kinda thing, unfortunately. You could just add a `.filter` first, I guess, but I don't have a strong opinion either way either

Shaptic:
@paulbellamy thank you for the CLI references! The difference in the SDK is two-fold:

 * signing is likely to be multi-machine (e.g. sending to another user or server), which isn't often true in the CLI case, so we can't just pass a `signers: Keypair[]` as that centralized everyone's private keys
 * this helper is distinct from the "assembly" step that would happen in the soroban-client portion (i.e. `Server.prepareTransaction`, where we already do misc. fee / auth / footprint adjustments). There, I entirely agree that we could do something like:
  1. `simulateTransaction` to get the footprint, fees, and auth
  2. kick that back to the user for them to figure out how they want to do auth, probably via `authorizeEntry`
  3. `simulateTransaction` again to get updated fees and leave auth untouched
  4. replace the footprint with the footprint from (1), because the second simulation would give us bs/irrelevant nonce footprints
  5. finally, sign & send

(Does that sound right, @aristidesstaffieri?)

src/sorobandata_builder.js
Shaptic: I would really love to stick these into the object prototypes but it doesn't work? Or I'm doing something very wrong... Either way, `SorobanDataBuilder` already has this heh, so closing for now.

Shaptic:
@paulbellamy in the sense that it will leave auth entries untouched if present in the `invokeHostFunction` being simulated? :pray:

paulbellamy:
If they are present on the txn it uses enforcing auth mode and doesn't alter the auth entries, yes.

Shaptic:
That's promising! Then we can utilize the other abstractions.

aristidesstaffieri:
> @paulbellamy thank you for the CLI references! The difference in the SDK is two-fold:
> 
> * signing is likely to be multi-machine (e.g. sending to another user or server), which isn't often true in the CLI case, so we can't just pass a `signers: Keypair[]` as that centralized everyone's private keys
> * this helper is distinct from the "assembly" step that would happen in the soroban-client portion (i.e. `Server.prepareTransaction`, where we already do misc. fee / auth / footprint adjustments). There, I entirely agree that we could do something like:
> 
> 1. `simulateTransaction` to get the footprint, fees, and auth
> 2. kick that back to the user for them to figure out how they want to do auth, probably via `authorizeEntry`
> 3. `simulateTransaction` again to get updated fees and leave auth untouched
> 4. replace the footprint with the footprint from (1), because the second simulation would give us bs/irrelevant nonce footprints (one might call it "noncense" ayy lmao, see PR description for details)
> 5. finally, sign & send
> 
> (Does that sound right, @aristidesstaffieri?)

Yeah that does sound right and is basically what current examples do.

aristidesstaffieri (approved):
Sorry I lost track of this one but this lgtm

aristidesstaffieri:
@Shaptic I'm curious what the motivation is for providing `authorizeInvocation`? Thinking of this through the lends of an of an app workflow, I can't imagine a scenario that would be better suited for `authorizeInvocation` than `authorizeEntry`. When you need to sign an auth entry in practice, I would imagine it would almost always be pre-built for you.

Shaptic:
@aristidesstaffieri that's a good point; I guess I'm thinking about advanced or optimized scenarios. For example, if any of the following isn't true, you'd want a `authorizeInvocation` to exist:

 - you don't want to (or need to) do simulation for auth (or don't trust it)
 - you want to send around just the call tree to misc. parties (easier, smaller than the whole entry)
 - you received a JSONified call tree (e.g. the Freighter popup? or via `buildInvocationTree`) and want to build an entry from it

Maybe these are far fetched? But it still seems nice for flexibility at the stellar-base level to not necessarily explicitly rely on preflight to build the data structures you may need.

aristidesstaffieri:
> @aristidesstaffieri that's a good point; I guess I'm thinking about advanced or optimized scenarios. For example, if any of the following isn't true, you'd want a `authorizeInvocation` to exist:
> 
> * you don't want to (or need to) do simulation for auth (or don't trust it)
> * you want to send around just the call tree to misc. parties (easier, smaller than the whole entry)
> * you received a JSONified call tree (e.g. the Freighter popup? or via `buildInvocationTree`) and want to build an entry from it
> 
> Maybe these are far fetched? But it still seems nice for flexibility at the stellar-base level to not necessarily explicitly rely on preflight to build the data structures you may need.

Not far fetched, those make a ton of sense. I just wanted to make sure I understood the use case.

PR #680, opened 2023-08-15
Add ability to modify and retrieve the resource footprint

Shaptic:
This makes it easier to build the data structure from an existing footprint.

### DIFF
--- a/src/sorobandata_builder.js
+++ b/src/sorobandata_builder.js
@@ -35,10 +35,8 @@ export class SorobanDataBuilder {
   constructor(sorobanData) {
     let data;
 
-    if (typeof sorobanData === 'string') {
-      data = SorobanDataBuilder.fromXDR(sorobanData, 'base64');
-    } else if (ArrayBuffer.isView(sorobanData)) {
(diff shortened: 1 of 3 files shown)


src/sorobandata_builder.js
```diff
@@ -102,22 +100,37 @@ export class SorobanDataBuilder {
     return this;
   }
 
+  /**
+   * Appends the given ledger keys to the existing storage access footprint.
+   * @param {xdr.LedgerKey[]} readOnly   read-only keys to add
+   * @param {xdr.LedgerKey[]} readWrite  read-write keys to add
+   * @returns {SorobanDataBuilder} this builder instance
+   */
+  appendFootprint(readOnly, readWrite) {
```
aristidesstaffieri, line 109: is it worth here to check if the `readOnly`/`readWrite` are actually `LedgerKey`s?

aristidesstaffieri (approved):

src/sorobandata_builder.js
Shaptic: nah, we pretty much assume sanity in many, **many** places in this codebase. the TypeScript should do a good enough job enforcing this

PR #685, opened 2023-08-31
Upgrade dependencies to their latest minor/patch version

Shaptic:
This also runs `yarn fmt` which wasn't done, for some reason.


sreuland (approved):
just for context, was the upgrade triggered with just 'yarn upgrade'?

Shaptic:
@sreuland I do

```bash
yarn upgrade-interactive --latest
```

PR #688, opened 2023-09-11
Merges the final Protocol 20 XDR (i.e. for testnet) into `soroban`

Shaptic:
You can see that the XDR now points to the following:
 * curr: https://github.com/stellar/stellar-xdr/commit/9ac02641139e6717924fdad716f6e958d0168491
 * next: https://github.com/stellar/stellar-xdr/commit/440dc9512b6e72cf84965641c5eb495d6043ed73

As indicated [here](https://stellarfoundation.slack.com/archives/C030Z9EHVQE/p1694460508382989) by @sisuresh. The following migrations needed to be made:
 * src/xdr.js has been updated to import from `curr_generated`, flipping the switch from using `next` to `curr`
 * other changes are described in 3b1eb163bcad70bd3246e0a7f2b357a5c9b3e8a4, replicated here:
 * function invocation now needs an `InvokeContractArgs` structure
 * `signatureArgs` was a 0-elem vec, now an `ScVal` w/ 0-elem vec
 * `bodyType` and `metaDataSizeBytes` are gone
 * some minor test stuff has been tweaked for readability

Note for reviewers: the diffs of src/next_generated.js and types/next.d.ts should actually be rather readable and a good way to see the underlying XDR schema changes in this bump.


Makefile
```diff
@@ -25,7 +30,7 @@ XDR_FILES_NEXT= \
 	Stellar-contract-config-setting.x
 XDR_FILES_LOCAL_NEXT=$(addprefix xdr/next/,$(XDR_FILES_NEXT))
 
-XDRGEN_COMMIT=8d303b1
+XDRGEN_COMMIT=master
```
sreuland, line 33: is there a reason not to pin xdrgen runtime to a specific hash/tag?

test/unit/auth_test.js
```diff
@@ -28,12 +28,14 @@ describe('building authorization entries', function () {
     );
 
     let cred = entry.credentials().address();
-    let args = cred.signatureArgs().map((v) => StellarBase.scValToNative(v));
+    let sig = cred.signature();
+
+    expect(sig.switch().name).to.be.equal('scvVec');
+    const args = StellarBase.scValToNative(sig.vec()[0]);
```
sreuland, line 34: minor nit, would this throw on `sig.vec()` to avoid  need for `expect` on type first

test/unit/auth_test.js
```diff
@@ -28,12 +28,14 @@ describe('building authorization entries', function () {
...
 
     expect(cred.signatureExpirationLedger()).to.equal(123);
-    expect(args.length).to.equal(1);
     expect(
-      StellarBase.StrKey.encodeEd25519PublicKey(args[0]['public_key'])
+      StellarBase.StrKey.encodeEd25519PublicKey(args['public_key'])
```
sreuland, line 38: are sig arg keys like `public_key` doc'd elsewhere such as js/xdr/public, or maybe this test is the doc.

Makefile
Shaptic: its usually always been master, just got changed when we had the weird `set_` bug so I could do xdrgen w/o needing that PR merged

test/unit/auth_test.js
Shaptic: it's documented as part of the signature docs, here: https://soroban.stellar.org/docs/fundamentals-and-concepts/invoking-contracts-with-transactions#stellar-account-signatures (that should probably be referenced here! added in 70ad49e

test/unit/auth_test.js
Shaptic: good point! amended this to a length check, instead: 19fdbf7

test/unit/auth_test.js
sreuland: i did find ref to `public_key` in the js auth conversion - https://github.com/stellar/js-stellar-base/blob/soroban-testnet-xdr/src/auth.js#L176, that key name is a constant which core defines correct?

Shaptic: yeah, `public_key` and `signature` are the field names of the struct that Core expects as the signature. that's what I added in the cross-reference!

test/unit/transaction_builder_test.js
```diff
@@ -71,31 +71,27 @@ describe('TransactionBuilder', function () {
...
-          }),
-          instructions: 0,
-          readBytes: 5,
-          writeBytes: 0,
-          extendedMetaDataSizeBytes: 0
-        }),
-        refundableFee: StellarBase.xdr.Int64.fromString('1'),
-        ext: new StellarBase.xdr.ExtensionPoint(0)
-      });
+      sorobanTransactionData = new StellarBase.SorobanDataBuilder()
```
sreuland, line 74: nice reduction!

Makefile
```diff
@@ -1,15 +1,20 @@
...
+	Stellar-types.x \
+	Stellar-contract.x \
+	Stellar-contract-env-meta.x \
+	Stellar-contract-meta.x \
+	Stellar-contract-spec.x \
+	Stellar-contract-config-setting.x
 XDR_FILES_LOCAL_CURR=$(addprefix xdr/curr/,$(XDR_FILES_CURR))
 
-XDR_BASE_URL_NEXT=https://github.com/stellar/stellar-xdr/raw/e372df9f677961aac04c5a4cc80a3667f310b29f
+XDR_BASE_URL_NEXT=https://github.com/stellar/stellar-xdr/raw/440dc9512b6e72cf84965641c5eb495d6043ed73
```
sreuland, line 17: the convention is to generate next and curr even though respectively one is never referenced, just curious. does the dormant un-used one still get packaged into dist?

sreuland (approved):

Makefile
Shaptic: ooh that's a great question. I think it does not get packaged because it never gets used nor exported by `src/xdr.js`, but I will double-check this!

PR #690, opened 2023-09-11
Add support to convert strings to large integer and address `ScVal`s

### LINKED ISSUE #683: nativeToScVal did not handle numbers correctly.
**Describe the bug**

The following code should output `false`, but we got `true`, `2 ** 127 - 1` was not converted correctly.
```js
const sorobanClient = require("soroban-client")

const a = sorobanClient.nativeToScVal(-(2 ** 127), {
    type: "i128",
}).toXDR("base64")

const b = sorobanClient.nativeToScVal(2 ** 127 - 1, {
    type: "i128",
}).toXDR("base64")

console.log(a === b)
```

**What version are you on?**
Check `yarn.lock` or `package-lock.json` to find out precisely what version of `stellar-base` you're running.
```
soroban-client: 0.11.0
stellar-base 10.0.0-soroban.7
```
**To Reproduce**
Check the code above.

**Expected behavior**
Check the code above.

**Additional context**
i256, u128, and u256 also have this issue. I speculate that it is due to the incorrect handling of the sign bit?

Shaptic:
In summary, this adds the following conversions:

 * JS string -> `ScVal` with an `ScAddress`
 * JS string -> `ScVal` with large integers such as `Int128Parts` (accomplished via `XdrLargeInt`)

For details, you can refer to the changes in test/unit/scval_test.js for new usage examples.

Closes #683.

--------

An unrelated change is to tests.yml which actually does `prettier` formatting before checking the diff :facepalm:. This is the cause of a handful of formatting changes (`yarn fmt` wasn't run in a previous PR).


src/numbers/xdr_large_int.js
```diff
@@ -238,16 +226,29 @@ export class XdrLargeInt {
...
+  }
+
+  /**
+   * Convert the raw `ScValType` string (e.g. 'scvI128', generated by the XDR)
+   * to a type description for {@link XdrLargeInt} construction (e.g. 'i128')
+   *
+   * @param {string} scvType  the `xdr.ScValType` as a string
+   * @returns {string} a suitable equivalent type to construct this object
+   */
+  static getType(scvType) {
```
sreuland, line 251: I know this is unchanged, but would it make sense to call `isType(scvType)` here and jsdoc that it throws if `scvType` invalid?

test/unit/scval_test.js
```diff
@@ -206,4 +210,31 @@ describe('parsing and building ScVals', function () {
   it('throws on arrays with mixed types', function () {
     expect(() => nativeToScVal([1, 'a', false])).to.throw(/same type/i);
   });
+
+  it('lets strings be large integer ScVals', function () {
+    ['i64', 'i128', 'i256', 'u64', 'u128', 'u256'].forEach((type) => {
+      const scv = nativeToScVal('12345', { type });
+      expect(XdrLargeInt.getType(scv.switch().name)).to.equal(type);
```
sreuland, line 217: is it worth adding a line or separate test case, to demonstrate expectation when invalid
```
XdrLargeInt.getType('ab');
```

sreuland (approved):

src/numbers/xdr_large_int.js
Shaptic: I thought this was a good idea so I gave it a shot, but then I realized two things:

 - this shouldn't ever be called by clients, honestly, so I made it private in 7de0557 (it was only exported for tests)
 - the non-int cases are already covered by `scValToBigInt`, here: https://github.com/stellar/js-stellar-base/blob/scval-better-ints/src/numbers/index.js#L57

So this check would just add unnecessary internal overhead. If we ever need to make this public, we should definitely do the check!

test/unit/scval_test.js
Shaptic: good call! but this is implicitly already handled by an existing test case, here:

https://github.com/stellar/js-stellar-base/blob/25a2b68a7dd99c182edf74f46bd75a998d59e7bd/test/unit/scint_test.js#L247-L251

PR #692, opened 2023-09-13
Correctly build operations via `Contract.call()`

Shaptic:
The type of `InvokeContractArgs.functionName` is supposed to be a `string` now, not an `ScSymbol` like before.

This **should** have been caught by testing but they were being skipped with `xit` by mistake :facepalm:

### DIFF
--- a/test/unit/contract_test.js
+++ b/test/unit/contract_test.js
@@ -83,7 +83,7 @@ describe('Contract', function () {
     });
 
     it('passes the method name as the second arg', function () {
-      expect(args.functionName()).to.deep.equal(xdr.ScVal.scvSymbol('method'));
+      expect(args.functionName()).to.deep.equal('method');
     });
 
     it('passes all params after that', function () {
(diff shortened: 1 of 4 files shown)


sreuland (approved):

test/unit/contract_test.js
```diff
@@ -59,7 +62,11 @@ describe('Contract', function () {
       StellarBase.nativeToScVal(2, { type: 'i32' })
     );
 
-    xit('builds valid XDR', function () {
+    it('works with no parameters', function () {
+      contract.call('empty').toXDR();
```
sreuland, line 66: not now, but would the `toXDR()` output need to be compared with expected output? or if it doesn't throw at that point is good enough.

Shaptic: yeah the structure itself is checked in other tests, this is just a "did you build working xdr" test (helps check types, etc. when it isn't marked to skip :facepalm:)

PR #693, opened 2023-09-13
Prepare v10.0.0-beta.1 for release

Shaptic:
Includes a critical bugfix, see changelog.


Shaptic:
Note that package.json is already set

PR #694, opened 2023-09-14
Prepare v10.0.0-beta.2 for release

Shaptic:
This only contains #678 and a bugfix for `xdr.d.ts`.


package.json
```diff
@@ -1,10 +1,10 @@
 {
   "name": "stellar-base",
-  "version": "10.0.0-beta.1",
+  "version": "10.0.0-beta.2",
   "description": "Low-level support library for the Stellar network.",
   "main": "./lib/index.js",
   "browser": {
-    "main": "./dist/stellar-base.min.js",
+    "main": "./dist/stellar-base.js",
```
Shaptic, line 7: this change is to make debugging easier for downstream systems until we release a stable version

aristidesstaffieri (approved):

Issue #695, opened 2023-09-18
memo only accepts string in typescript, but code and comments allow for array


acharb:
**Describe the bug**
in the [memo code and comment](https://github.com/stellar/js-stellar-base/blob/f17ef334b7b26438a57d5d1f516fae291d9a7346/src/memo.js#L181-L188) it says can give an array to Memo.hash , but the type definition only allows for string

not a blocker because can just do buffer.toString('hex'), but it was a little confusing to me for a bit

**What version are you on?**

10.0.0-beta.1

**To Reproduce**
// in typescript code

const buffer = Buffer.from("AAAAAAAAAAAAAAAAAAAAAMAP+8deo0TViBD09TfOBY0=", "base64");
const memo = Memo.hash(buffer); // throws error

const memo = Memo.hash(buffer as any as string); // no error thrown, so code handles it fine

**Expected behavior**

no error

**Additional context**

PR #698, opened 2023-10-02
Fix type definition for `Memo.hash` to allow buffers

Shaptic:
Closes https://github.com/stellar/js-stellar-base/issues/695.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -211,7 +211,7 @@ export type MemoValue = null | string | Buffer;
 
 export class Memo<T extends MemoType = MemoType> {
   static fromXDRObject(memo: xdr.Memo): Memo;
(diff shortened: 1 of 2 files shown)


types/index.d.ts
```diff
@@ -211,7 +211,7 @@ export type MemoValue = null | string | Buffer;
 
 export class Memo<T extends MemoType = MemoType> {
   static fromXDRObject(memo: xdr.Memo): Memo;
-  static hash(hash: string): Memo<MemoType.Hash>;
+  static hash(hash: string | Buffer): Memo<MemoType.Hash>;
```
sreuland, line 214: for consideration, would be nice to add a test demonstrating the fixed outcome.

sreuland (approved):

PR #701, opened 2023-10-09
Fixes a type comparison bug in authorization helpers.

Shaptic:
For some reason, !== on the object does not work in some envs. 
Note that it works fine in the unit tests.
This is a faster check, anyway.

Reported by @tyvdh on Discord [here](https://discord.com/channels/897514728459468821/1138959805218836510/1161061470289789038).


tyvdh (approved):
lgtm

PR #704, opened 2023-10-16
Upgrade XDR to the latest (hopefully) stable version

Shaptic:
TODO: List of changes.

Note: The diffs to `src/generated/*.js` and `types/*.d.ts` should be very readable.


Makefile
```diff
@@ -14,7 +14,7 @@ XDR_FILES_CURR= \
 	Stellar-contract-config-setting.x
 XDR_FILES_LOCAL_CURR=$(addprefix xdr/curr/,$(XDR_FILES_CURR))
 
-XDR_BASE_URL_NEXT=https://github.com/stellar/stellar-xdr/raw/440dc9512b6e72cf84965641c5eb495d6043ed73
+XDR_BASE_URL_NEXT=https://github.com/stellar/stellar-xdr/raw/45fbdc24a3afaffdaa47bef4656f9945d1a63581
```
sreuland, line 17: just curious, what's in `next` that isn't in `curr` at this point? since both are getting new target refs, assuming the `curr` xdr changes were ported into the `next` as well, and are just being updated here for completeness, but the `next` generated code is not actively imported by prod build/dist, correct?

sreuland (approved):
great changelog notes

Makefile
Shaptic: You are correct that the `next` xdr is not in the final build! The two are identical right now, but I figured it makes sense to still keep generating `next` since we will eventually need it to separate Protocol 21 work from mainnet.

PR #712, opened 2023-11-10
correct the return type to 'string'

jeesunikim:
It's supposed to be `string` not `Address`


Shaptic (approved):
dunno where that came from heh, thank you!

PR #714, opened 2023-11-20
Republish package under @stellar/ scope

Shaptic:
Notice that this is targeting the `stable` branch: we will only include this rename as part of the stable package release.

See stellar/js-xdr#111, stellar/js-xdr#112, and stellar/js-xdr#113 for what was needed in that package to accomplish this.

This also moves the code to use `@stellar/js-xdr` instead of `js-xdr`.

Related: stellar/js-stellar-sdk#851.

### DIFF
--- a/.github/workflows/npm_publish.yml
+++ b/.github/workflows/npm_publish.yml
@@ -32,6 +32,6 @@ jobs:
 
       - name: Deprecate the old package
         run: |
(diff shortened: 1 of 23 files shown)


stellarsaur (approved):
LGTM aside from the very minor GitHub Actions warnings

PR #721, opened 2023-12-22
Actually omit the 'contractId' field from humanized events when it's empty

Shaptic:
Both the documentation and the TypeScript definitions outline it as being omitted when the field is not present in the event (in case of system events), so this is a bugfix rather than a breaking change.

### DIFF
--- a/test/unit/events_test.js
+++ b/test/unit/events_test.js
@@ -14,9 +14,10 @@ describe('humanizing raw events', function () {
   // workaround for xdr.ContractEventBody.0(...) being invalid lol
   const cloneAndSet = (newBody) => {
     const clone = new xdr.ContractEventBody(
-      0, new xdr.ContractEventV0({
+      0,
+      new xdr.ContractEventV0({
         topics: [],
-        data: xdr.ScVal.scvVoid(),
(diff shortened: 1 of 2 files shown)


test/unit/events_test.js
```diff
@@ -12,9 +12,14 @@ describe('humanizing raw events', function () {
...
-  const bodyModel = xdr.ContractEventBody.fromXDR('AAAAAAAAAAAAAAAB', 'base64');
   const cloneAndSet = (newBody) => {
-    const clone = xdr.ContractEventBody.fromXDR(bodyModel.toXDR());
+    const clone = new xdr.ContractEventBody(
+      0,
+      new xdr.ContractEventV0({
+        topics: [],
+        data: xdr.ScVal.scvVoid()
+      })
+    );
```
sreuland, line 22: it doesn't look like test is exercising omitted contractId behavior, can another test be added to assert if this is important?

sreuland (approved):

test/unit/events_test.js
Shaptic: ooh great point, thanks :pray:

PR #730, opened 2024-02-12
Release v11.0.0

Shaptic:
**Note:** This version is (still) compatible with Protocol 20. Most people should be unaffected by the technically-breaking changes and can treat this more like a v10.0.3 patch release (see changelog).


Shaptic:
@SocketSecurity ignore npm/sodium-native@4.0.8

tamirms (approved):

PR #738, opened 2024-04-25
Upgrade XDR to Protocol 21

Shaptic:
This updates the XDR to stellar/stellar-xdr@1a04392432dacc0092caaeae22a600ea1af3c6a5.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-XDR_BASE_URL_CURR=https://github.com/stellar/stellar-xdr/raw/bb54e505f814386a3f45172e0b7e95b7badbe969
+XDR_BASE_URL_CURR=https://github.com/stellar/stellar-xdr/raw/1a04392432dacc0092caaeae22a600ea1af3c6a5
 XDR_BASE_LOCAL_CURR=xdr/curr
 XDR_FILES_CURR= \
 	Stellar-SCP.x \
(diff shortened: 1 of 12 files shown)


Makefile
```diff
@@ -14,7 +14,7 @@ XDR_FILES_CURR= \
 	Stellar-contract-config-setting.x
 XDR_FILES_LOCAL_CURR=$(addprefix xdr/curr/,$(XDR_FILES_CURR))
 
-XDR_BASE_URL_NEXT=https://github.com/stellar/stellar-xdr/raw/bb54e505f814386a3f45172e0b7e95b7badbe969
+XDR_BASE_URL_NEXT=https://github.com/stellar/stellar-xdr/raw/4bd4827b3361f266d295cbc83c5d7f7396d782d9
```
sreuland, line 17: is it necessary to bump the 'next' codeline in this pr scope, seems the intent is to just change the xdr version for client usage which is 'curr' ?

sreuland (approved):
all this gen'd code looks great!

Makefile
Shaptic: hmm I guess technically not but there is no harm either in case someone wanted to use the SDK to dev on `next` for whatever reason :shrug: imo it's nice to keep them updated at the same time

PR #744, opened 2024-05-08
Expose certain functions from `@stellar/js-xdr` so downstream don't need to mix dependencies.

Shaptic:


chadoh (approved):
Thanks!

BlaineHeffron:
When I use the cereal namespace instead of just importing directly from js-xdr this code gets a TypeError:
```js
export function processSpecEntryStream(buffer: Buffer) {
  const reader = new cereal.XdrReader(buffer);
  const res: xdr.ScSpecEntry[] = [];
  while (!reader.eof){
    res.push(xdr.ScSpecEntry.read(reader));
  }
  return res;
}
```

```
Argument of type 'XdrReader' is not assignable to parameter of type 'Buffer'.
  Type 'XdrReader' is missing the following properties from type 'Buffer': write, toJSON, equals, compare, and 91 more.

93     res.push(xdr.ScSpecEntry.read(reader));
```

aditya1702 (approved):

PR #752, opened 2024-06-10
Updated the docs for extend/restore footprint ops.

dmkozh:
Extend TTL doc was incorrect, and restore doc lacked some important details.


Shaptic (approved):
Nice, I added some clarity in https://github.com/stellar/js-stellar-base/pull/743 but that would've been a breaking change so this is a good stop-gap inbetween :+1: thank you!

PR #753, opened 2024-06-11
Fix handling `ScError` types in `scValToNative`

Shaptic:
Errors were updated, but converting them to native objects wasn't. Errors are now converted to the following format:

```typescript
interface Error {
  type: "contract" | "system";
  code: number;
  value?: string; // only present for type === 'system'
}
```

### DIFF
--- a/src/scval.js
+++ b/src/scval.js
@@ -354,12 +354,12 @@ export function scValToNative(scv) {
     case xdr.ScValType.scvError().value:
       // Distinguish errors from the user contract.
       if (scv.error().switch() === xdr.ScErrorType.sceContract().value) {
-        return { "type": "contract", "code": scv.error().contractCode() };
+        return { type: 'contract', code: scv.error().contractCode() };
       }
       return {
-        "type": "soroban",
-        "code": scv.error().code().value,
-        "value": scv.error().code().name,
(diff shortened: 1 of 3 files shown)


Shaptic:
cc @chadoh and @willemneal in case this needs to be done in `ContractSpec`, too

src/scval.js
```diff
@@ -351,22 +351,16 @@ export function scValToNative(scv) {
...
-        case xdr.ScStatusType.sstHostAuthError().value:
-        default:
-          break;
+    case xdr.ScValType.scvError().value:
+      // Distinguish errors from the user contract.
+      if (scv.error().switch() === xdr.ScErrorType.sceContract().value) {
+        return { type: 'contract', code: scv.error().contractCode() };
       }
+      return {
+        type: 'soroban',
```
sreuland, line 360: would 'system' be a more generalized type name, rather than 'soroban' which may change due to branding, etc?

sreuland (approved):
can a unit test be added to verify the new output?

src/scval.js
Shaptic: yesss, great idea

Shaptic:
@sreuland yessir! done :saluting_face:

PR #757, opened 2024-06-22
added clearing and adding operations at a specified index

JuanRdBO:
# Summary of changes

- Added two methods: `addOperationAtIndex` and `clearOperationAtIndex` in the `TransactionBuilder` class 

# Use Case

I wanted to add them because I was trying to change the amount of one operations after cloning and realized it would that much easier if I could just clear an operation at a specified index and then add it add the updated operation at the same index again.

### DIFF
--- a/src/transaction_builder.js
+++ b/src/transaction_builder.js
@@ -226,14 +226,39 @@ export class TransactionBuilder {
   addOperation(operation) {
     this.operations.push(operation);
     return this;
+  }	
+  
+  /**
+  * Adds an operation to the transaction at a specific index.
+  *
+  * @param {xdr.Operation} operation - The xdr operation object to add, use {@link Operation} static methods.
+  * @param {number} index - The index at which to insert the operation.
+  *
+  * @returns {TransactionBuilder} - The TransactionBuilder instance for method chaining.
+  */
+  addOperationAtIndex(operation, index) {
+    this.operations.splice(index, 0, operation);
+    return this;
   }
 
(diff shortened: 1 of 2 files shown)


src/transaction_builder.js
```diff
@@ -226,6 +226,19 @@ export class TransactionBuilder {
...
+  
+  /**
+   * Adds an operation to the transaction at a specific index.
+   *
+   * @param {xdr.Operation} operation - The xdr operation object to add, use {@link Operation} static methods.
+   * @param {number} index - The index at which to insert the operation.
+   *
+   * @returns {TransactionBuilder} - The TransactionBuilder instance for method chaining.
+   */
+  addOperationAtIndex(operation, index) {
```
Shaptic, line 239: This is great! Thanks for making it so thorough. I'd like to suggest we name these `addOperationAt` and `insertOperationAt` (no `Index` suffix) to better align with JavaScript's [`Array.at`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at).

Shaptic (commented):
I love this idea! Thank you for the PR. Could you add some unit tests that demonstrate that they function? You can reference `/test/unit/transaction_builder_test.js` for existing cases and run them with `yarn test`.

Shaptic (approved):

Shaptic:
Can you run (and commit) `yarn fmt`? Linter is complaining.

PR #758, opened 2024-06-27
Compare assets by integer rather than by object

Shaptic:
This increases cross-compatibility across different instances of `stellar-base` by comparing integers rather than objects (which will be `!=` given two different `base`s).


psheth9 (approved):
Looks good !!

PR #759, opened 2024-06-27
Automatically sort maps to match Soroban environment expectations

Shaptic:
Closes stellar/js-stellar-sdk#996 and stellar/js-stellar-sdk#966.

Note that test changes just involve changing the order of the map fields to be sorted accordingly in the "expected output" portion of the test.

Also, the diff is simpler if you [turn whitespace off](https://github.com/stellar/js-stellar-base/pull/759/files?diff=unified&w=1).

### DIFF
--- a/src/scval.js
+++ b/src/scval.js
@@ -138,7 +138,7 @@ import { ScInt, XdrLargeInt, scValToBigInt } from './numbers/index';
  */
 export function nativeToScVal(val, opts = {}) {
   switch (typeof val) {
-    case 'object':
+    case 'object': {
       if (val === null) {
         return xdr.ScVal.scvVoid();
       }
@@ -190,8 +190,14 @@ export function nativeToScVal(val, opts = {}) {
         );
       }
 
+      // The Soroban runtime expects maps to have their keys in sorted order, so
+      // let's do that here as part of the conversion to prevent confusing error
+      // messages on execution.
+      const entries = Object.entries(val);
+      entries.sort((entry1, entry2) => entry1[0].localeCompare(entry2[0]));
+
       return xdr.ScVal.scvMap(
-        Object.entries(val).map(([k, v]) => {
+        entries.map(([k, v]) => {
           // the type can be specified with an entry for the key and the value,
           // e.g. val = { 'hello': 1 } and opts.type = { hello: [ 'symbol',
           // 'u128' ]} or you can use `null` for the default interpretation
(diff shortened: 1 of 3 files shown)


src/scval.js
```diff
@@ -191,20 +191,26 @@ export function nativeToScVal(val, opts = {}) {
...
-          return new xdr.ScMapEntry({
-            key: nativeToScVal(k, keyOpts),
-            val: nativeToScVal(v, valOpts)
-          });
-        })
+        Object.entries(val)
+          // The Soroban runtime expects maps to have their keys in sorted
+          // order, so let's do that here as part of the conversion to prevent
+          // confusing error messages on execution.
+          .sort(([key1], [key2]) => key1.localeCompare(key2))
```
sreuland, line 198: js maps allow null key, for the paranoid, is a null check on `key1` worthwhile?

Shaptic: Good point! But that would be an invalid map for Soroban anyway, so... user error?

test/unit/scval_test.js
```diff
@@ -182,12 +182,12 @@ describe('parsing and building ScVals', function () {
     scv = nativeToScVal(
       {
         hello: 'world',
-        goodbye: [1, 2, 3]
+        there: [1, 2, 3]
```
sreuland, line 185: it's tough to grasp how expectations are validating the new sorting of keys from this test.

sreuland (approved):
left a minor comment for consideration.

src/scval.js
sreuland: sounds fine, it seemed`val` param is created by user, and not much convention over what gets passed in that case, if they send a map with a null key they'll get the message from the stack trace!

re1ro (approved):

re1ro:
Thank you for the expedited solution! 😄

src/scval.js
Shaptic: Yep, exactly! This is all just conversion to smart contract values which users define anyway - if they pass null it isn't valid, so we can avoid the extra check!

test/unit/scval_test.js
Shaptic: So the main test for order was the earlier one. This test was evaluating that `type` was being used correctly, and there are some assertions about `scv[0]` and `scv[1]` that become out of order since `goodbye` comes before `hello` lexicographically, so instead of modifying the assertion checks I just changed the key name so it'd stay in the same order as before.

PR #763, opened 2024-07-18
Add support for `nativeToScVal` to convert strings to all number sizes.

Shaptic:
Previously, `i32` and `u32` could not automatically come from string values when using `nativeToScVal`. Namely, `nativeToScVal("4", { type: "i32" })` was explicitly forbidden. This was my design: the rationale was that if it's a 32-bit integer, it should be passed as a `Number`. 

However, this logic does not hold true for maps, because in JavaScript, all map keys are strings. Therefore, creating an `ScVal` representing an integer-to-integer map is impossible via `nativeToScVal`, e.g., a function signature like this one:

```rust
fn spin(env: Env, pairs: Map<u32, u64>) -> boolean;
```

cannot be executed by 
```js
nativeToScVal(
  { 1234: 5678 }, 
  { 
    type: { 
      "1234": [ "u32, "u64" ],
      "5678": [ "u32, "u64" ],
    }
);
```

because `1234` would be interpreted as a string, and thus could not be automatically converted to a `number` for a `u32` inference.

### DIFF
--- a/test/unit/transaction_builder_test.js
+++ b/test/unit/transaction_builder_test.js
@@ -956,51 +956,60 @@ describe('TransactionBuilder', function () {
     });
 
     it('adds operations at a specific index', function () {
-        const builder = new StellarBase.TransactionBuilder(source, {
-          fee: '100',
-          timebounds: { minTime: 0, maxTime: 0 },
-          memo: new StellarBase.Memo(StellarBase.MemoText, 'Testing adding op at index'),
-          networkPassphrase
-        });
+      const builder = new StellarBase.TransactionBuilder(source, {
+        fee: '100',
+        timebounds: { minTime: 0, maxTime: 0 },
+        memo: new StellarBase.Memo(
+          StellarBase.MemoText,
+          'Testing adding op at index'
+        ),
+        networkPassphrase
+      });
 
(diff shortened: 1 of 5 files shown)


test/unit/transaction_builder_test.js
```diff
@@ -956,51 +956,60 @@ describe('TransactionBuilder', function () {
...
-        const newTx = clonedTx.build()
-        // Assert that the operations are the same, but the first one is updated
-        expect(newTx.operations.length).to.equal(2);
-        expect(newTx.operations[0].source).to.equal(source.accountId());
-        expect(parseInt(newTx.operations[0].amount)).to.equal(3);
-        expect(newTx.operations[1].source).to.equal(source.accountId());
-        expect(parseInt(newTx.operations[1].amount)).to.equal(2);
+        }),
+        1
+      );
```
Shaptic, line 987: Reviewers' note: this is just whitespace changes from a stray `yarn fmt` (not sure how the unformatted versions snuck in)

sreuland (approved):
detailed pr description was very helpful for doing review, thanks!

willemneal (approved):
Good catch!

PR #764, opened 2024-07-18
Handle cases in `humanizeEvents` where `contractId` is not set entirely

Shaptic:


src/events.js
```diff
@@ -29,9 +29,10 @@ export function humanizeEvents(events) {
 
 function extractEvent(event) {
   return {
-    ...(event.contractId() != null && {
-      contractId: StrKey.encodeContract(event.contractId())
-    }),
+    ...(event.contractId != null &&
```
sreuland, line 32: ah, pure js, interface checking by presence of method, may be able to consolidate this to a single conditional, seen a slight variation to look for strictly function presence:
```
typeof event.contractId === 'function'
```

sreuland (approved):
left one comment for consideration.

src/events.js
Shaptic: Ooh nice call!! I wasn't happy with the double-check but this is a great solution

Shaptic: Ah wait no, we still need to check whether or not the contract ID, if the function is present, is actually valid (because `contractId()` might return `null` :facepalm:), since in that case we don't want to include it in the response

PR #765, opened 2024-07-26
Fix the return type of `walkInvocationTree`'s callback to allow void returns as intended

Shaptic:


sreuland:
changes look good, can approve, but for consideration, wondering if it's time to introduce `ts` test to assert these will work from client before releasing, same observation noted on https://github.com/stellar/js-stellar-base/pull/766/files#r1729557085

Shaptic:
@sreuland good suggestion :pray: generally the type tests don't help much but you can see in [580b2b8](https://github.com/stellar/js-stellar-base/pull/765/commits/580b2b8914d56dca0e9a580ba939ef699e3c1ac4) that this actually uncovered the fact that there is in fact a `void` return value (rather than `undefined`) which is great :clap: please review again!

sreuland (approved):
lgtm, nice work!

PR #766, opened 2024-08-08
Add missing methods to TypeScript definitions

Shaptic:
Self-explanatory, see diff.


types/index.d.ts
```diff
@@ -1058,7 +1059,9 @@ export class TransactionBuilder {
     options?: TransactionBuilder.TransactionBuilderOptions
   );
   addOperation(operation: xdr.Operation): this;
+  addOperationAt(op: xdr.Operation, i: number): this;
```
sreuland, line 1062: is it getting to point where it's worthwhile to add `ts` tests, `index.d.test.ts`? how do we know `addOperation` will work when used by a client, or is there some base validation done in build pipeline?

Shaptic: There's JS unit tests for it but you're right that we should generally add more to the `types/test.ts` file. Unfortunately said tests only fail when you, well, add to them, and if you're adding to them then you're adding to the `index.d.ts` file anyway, so it wouldn't have prevented this slip-up (which came from an external contributor anyway).

Shaptic: Added the typing tests in c5c5f9b!

sreuland (approved):
lgtm, and thanks for insight there on ts testing mechanics.

PR #772, opened 2024-09-06
Fix bug with signing auth entries in multisig scenarios

kalepail:
Not every auth entry will be signing for itself, in the case of a multisig account where the master is GAA and the sub is GBB the auth entry will be GAA but we need to sign and check for GBB. We need to check the payload against the key that actually did the signing.

Context: https://discord.com/channels/897514728459468821/1281628550008406078

### DIFF
--- a/src/auth.js
+++ b/src/auth.js
@@ -136,7 +136,7 @@ export async function authorizeEntry(
   } else {
     signature = Buffer.from(signer.sign(payload));
   }
-  const publicKey = Address.fromScAddress(addrAuth.address()).toString();
+  const publicKey = signer.publicKey();
 
   if (!Keypair.fromPublicKey(publicKey).verify(payload, signature)) {
     throw new Error(`signature doesn't match payload`);


src/auth.js
```diff
@@ -136,7 +136,7 @@ export async function authorizeEntry(
   } else {
     signature = Buffer.from(signer.sign(payload));
   }
-  const publicKey = Address.fromScAddress(addrAuth.address()).toString();
+  const publicKey = signer.publicKey();
```
Shaptic, line 139: `signer` can be a `function` (see previous `if` statement) if you're passing a callback; this will only work in the `Keypair` case (hence test failures)

Shaptic (approved):
Approving a PR I modified? Checkmate auth :chess_pawn:

src/auth.js
```diff
@@ -131,12 +135,14 @@ export async function authorizeEntry(
   const payload = hash(preimage.toXDR());
 
   let signature;
+  let publicKey;
   if (typeof signer === 'function') {
     signature = Buffer.from(await signer(preimage));
+    publicKey = Address.fromScAddress(addrAuth.address()).toString();
```
overcat, line 141: Hi @kalepail, I think we can make the `SigningCallback` return both the `signature` and the `public_key` at the same time, which will give users more flexibility. What do you think?

Shaptic: This would be a breaking change to the API but it seems like a good idea to solve this issue for both cases! We can include this in the protocol 22 release.

PR #779, opened 2024-09-17
Prepare v13.0.0-beta.1 (Protocol 22) for release.

Shaptic:
In order to keep `master` clean for bugfixes during the transition period, we should release from the Protocol 22 branch which has a well-defined set of already-approved and merged PRs (see lucky https://github.com/stellar/js-stellar-base/pull/777).

This PR tags off the version, publishes under the `beta` tag on NPM, and updates the changelog.

Note that this is tagged as a `beta` instead of an `rc` so that we can properly handle any anticipated changes to core that occur throughout the release process (e.g. patch fixes after `v22.0.0` that require XDR structure or behavior changes), as was the case with Protocol 20.

### DIFF
--- a/.github/workflows/npm_publish.yml
+++ b/.github/workflows/npm_publish.yml
@@ -24,9 +24,9 @@ jobs:
 
       - name: Publish npm package to both places
         run: |
-          yarn publish --access public --tag beta
+          yarn publish --access public --tag protocol-22-beta
           sed -i -e 's#"@stellar/stellar-base"#"stellar-base"#' package.json
-          yarn publish --tag beta
+          yarn publish --tag protocol-22-beta
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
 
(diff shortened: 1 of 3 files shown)


.github/workflows/npm_publish.yml
```diff
@@ -24,9 +24,9 @@ jobs:
 
       - name: Publish npm package to both places
         run: |
-          yarn publish --access public
+          yarn publish --access public --tag beta
```
sreuland, line 27: since the tag has no ties to versions, it's hard to know what beta tag resolves to, looks like others [have asked about the ambiguity](https://github.com/yarnpkg/berry/issues/2444), do you think it's worth fully qualifying the tag with version, i.e. something like`13.0.0-beta`?

sreuland (approved):

.github/workflows/npm_publish.yml
Shaptic: Great point. We did `beta` in the past (and of course `beta` just means "latest but not ready") but we have a chance to redefine how we use tags now since it's a new package name. I do think `beta` is fine because it just means "latest but not ready for prod" but we can be clearer. 

What about `protocol-22-beta`? Even less ambiguous.

PR #787, opened 2025-01-16
Add an `xdr.scvMapSorted` constructor to automagically sort map `ScVal`s.

Shaptic:
Injects an `scvMapSorted` constructor into the `xdr` module so that people can created maps with sorted entries from raw `ScMapEntry` instances.

Closes stellar/js-stellar-sdk#1131.

### DIFF
--- a/src/scval.js
+++ b/src/scval.js
@@ -379,14 +379,14 @@ export function scValToNative(scv) {
 
 /// Inject a sortable map builder into the xdr module.
 xdr.scvSortedMap = (items) => {
-  let sorted = Array.from(items).sort((a, b) => {
+  const sorted = Array.from(items).sort((a, b) => {
     // Both a and b are `ScMapEntry`s, so we need to sort by underlying key.
     //
     // We couldn't possibly handle every combination of keys since Soroban
     // maps don't enforce consistent types, so we do a best-effort and try
     // sorting by "number-like" or "string-like."
-    let nativeA = scValToNative(a.key()),
-      nativeB = scValToNative(b.key());
+    const nativeA = scValToNative(a.key());
+    const nativeB = scValToNative(b.key());
 
     switch (typeof nativeA) {
       case 'number':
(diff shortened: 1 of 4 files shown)


src/scval.js
```diff
@@ -383,3 +376,27 @@ export function scValToNative(scv) {
...
+    const nativeA = scValToNative(a.key());
+    const nativeB = scValToNative(b.key());
+
+    switch (typeof nativeA) {
+      case 'number':
+      case 'bigint':
+        return nativeA < nativeB ? -1 : 1;
+
+      default:
+        return nativeA.toString().localeCompare(nativeB.toString());
```
leighmcculloch, line 397: It is intriguing how... imperfect this is. Why does this use a different approach to `nativeToScVal`? `nativeToScVal` sorting logic appears to look like this:

```js
      return xdr.ScVal.scvMap(
        Object.entries(val)
          // The Soroban runtime expects maps to have their keys in sorted
          // order, so let's do that here as part of the conversion to prevent
          // confusing error messages on execution.
          .sort(([key1], [key2]) => key1.localeCompare(key2))
          .map(([k, v]) => {
            // the type can be specified with an entry for the key and the value,
            // e.g. val = { 'hello': 1 } and opts.type = { hello: [ 'symbol',
            // 'u128' ]} or you can use `null` for the default interpretation
            const [keyType, valType] = (opts?.type ?? {})[k] ?? [null, null];
            const keyOpts = keyType ? { type: keyType } : {};
            const valOpts = valType ? { type: valType } : {};
            return new xdr.ScMapEntry({
              key: nativeToScVal(k, keyOpts),
              val: nativeToScVal(v, valOpts)
            });
          })
      );

```

Shaptic: I think this actually reveals an issue with the way `nativeToScVal` sorts, instead. It made the assumption that map keys are strings (because in JS all `{}`s use strings for keys), but if those are being encoded differently (via `type`), then this might :boom:. 

Of course this convenience method isn't supposed to be a catch-all for all possible `ScVal`s, but it could be improved there for the easy cases (number-like). Both will work reliably for the common case, which is `struct`s, and that's the primary goal here.

leighmcculloch (approved):

PR #794, opened 2025-03-26
Improve encoding and decoding support for addresses

Shaptic:
This makes 
 * `Keypair`s compatible with `nativeToScVal`
 * `ScAddress`es decoded more reliably with `scValToNative`

### DIFF
--- a/src/scval.js
+++ b/src/scval.js
@@ -1,5 +1,6 @@
 import xdr from './xdr';
 
+import { Keypair } from './keypair';
 import { Address } from './address';
 import { Contract } from './contract';
 import { ScInt, XdrLargeInt, scValToBigInt } from './numbers/index';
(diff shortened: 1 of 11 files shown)


src/scval.js
```diff
@@ -151,6 +152,10 @@ export function nativeToScVal(val, opts = {}) {
         return val.toScVal();
       }
 
+      if (val instanceof Keypair) {
```
sreuland, line 155: may be worthwhile to see a unit test assert this happens, and what is expected if `publicKey` were null/undefined?

sreuland (approved):
lgtm, minor suggestion on test coverage for consideration.

src/scval.js
Shaptic: nice call, added! `publicKey` can't be undefined, though, only the `secretKey` can if you create it `fromPublicKey` instead of `fromSecret`.

PR #795, opened 2025-03-27
Remove custom `Buffer.subarray` polyfill

overcat:
In the latest version of React Native, the current patch implementation seems to not work smoothly, so let's remove it and instead use [`@exodus/patch-broken-hermes-typed-arrays`](https://github.com/ExodusMovement/patch-broken-hermes-typed-arrays) to fix this issue. The `@exodus/patch-broken-hermes-typed-arrays` is used by well-known applications like Rabby Wallet and Exodus, so I believe it should be trustworthy.

P.S. Without removing this patch, using @exodus/patch-broken-hermes-typed-arrays can also make everything work smoothly, so there is no rush to release a new version.

https://github.com/stellar/js-xdr/pull/128


Shaptic (approved):
Nice, a robust package you can just install is way better than a package-specific monkeypatch. Since this is breaking I'll have to merge it as part of Protocol 23 but I'll approve it. I'll also have to remember to add this caveat to the [RN instructions](https://github.com/stellar/js-stellar-sdk?tab=readme-ov-file#usage-with-react-native).

overcat:
Hi @Shaptic, I have recently been working on Ledger Live and Trezor Suite, and I feel that asking them to globally introduce [patch-broken-hermes-typed-arrays](https://github.com/ExodusMovement/patch-broken-hermes-typed-arrays) is not particularly realistic. The Ledger Live team and I tried a global introduction, but it broke the Polkadot integration (I haven't investigated the reason yet; it's quite chaotic here). So I'm wondering if we should keep the Buffer Patch in the SDK.

Additionally, [this Buffer patch](https://github.com/stellar/js-stellar-base/blob/master/config/buffer.js) in the SDK is not particularly complete and does not cover all scenarios. I wonder if we can directly introduce [patch-broken-hermes-typed-arrays](https://github.com/ExodusMovement/patch-broken-hermes-typed-arrays/blob/main/index.js) into this SDK to fix this issue?

overcat:
Hi @Shaptic, I'm a bit lost. If we remove this polyfill and patch it directly where it's used, like in [Trezor Suite](https://github.com/trezor/trezor-suite/blob/develop/suite-native/app/globalPolyfills.js#L19-L46), will it affect the Stellar SDK internally? It seems not? Is this related to our webpack configuration?

overcat:
I created a repository to reproduce the current SDK's strange behavior. https://github.com/overcat/StellarSDKBufferDebugApp/blob/main/app/(tabs)/index.tsx#L7-L60

If you want to run it, you need to configure the environment according to RN's instructions: https://reactnative.dev/docs/set-up-your-environment

This is the output on my end:
```
 (NOBRIDGE) LOG  Memo XDR: AAAAAQAAAAtIZWxsbyB3b3JsZAA=
 (NOBRIDGE) LOG  ----------------------------------------------------
 (NOBRIDGE) LOG  Transaction XDR: AAAAAgAAAAA0M9clt3S3+wc6seyCYDyVic6vmnjjhZjFZ96tcMQFVwAAAGQAAAAAAAAAAgAAAAEAAAAAAAAAAAAAAABoEMDNAAAAAQAAAAtIZWxsbyB3b3JsZAAAAAABAAAAAAAAAAEAAAAACz8P0xh5CP5PLpwPKq/AqrJVsDeqstmO9euSCApoB94AAAABREVNTwAAAADhCnWmdZbYIpNgKp76E8nE2eOlj0Y3PYWRiGNiCY68eQAAAAJUHqiwAAAAAAAAAAA=
 (NOBRIDGE) LOG  ----------------------------------------------------
 (NOBRIDGE) LOG  tx2 parsing failed
 (NOBRIDGE) LOG  [Error: Asset code is invalid (maximum alphanumeric, 12 characters at max)]
 (NOBRIDGE) LOG  tx3 parsed successfully
```

overcat:
It seems that polyfill in `js-xdr` is effective, but not in `stellar-base`. Maybe it's a configuration issue? Exploring...

overcat:
🤦‍♂️ The stellar-sdk does not use the compiled stellar-base, so the Buffer used by stellar-base is the unpatched version provided by stellar-sdk, which leads to this strange behavior. Why doesn't stellar-sdk use the compiled stellar-base? I don't know.

overcat:
https://github.com/stellar/js-stellar-base/pull/798 should be able to solve the above problem

Shaptic (commented):
@overcat do we still want this in? It's hard to say based on your later comments in the PR.

Shaptic (approved):

PR #647, opened 2023-07-17
Prepare `soroban.4` tag for release

Shaptic:
This includes #645 and upgrades applicable `devDependencies` to minor versions.

PR #798, opened 2025-05-01
Fix browser compatibility with proper module resolution and UMD configuration.

overcat:


package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Low-level support library for the Stellar network.",
   "main": "./lib/index.js",
   "browser": {
-    "main": "./dist/stellar-base.min.js",
+    "./lib/index.js": "./dist/stellar-base.min.js",
```
overcat, line 7: See https://github.com/stellar/js-stellar-base/pull/647#discussion_r2068086482

package.json
```diff
@@ -85,6 +85,7 @@
     "@typescript-eslint/parser": "^6.20.0",
     "babel-loader": "^9.2.1",
     "babel-plugin-istanbul": "^6.1.1",
+    "buffer": "^6.0.3",
```
overcat, line 88: I think only build tools will use it, so I moved it to devDependencies.

overcat: I'm not 100% sure, so let's move it back for now.

Shaptic (approved):

PR #800, opened 2025-05-19
Regenerate XDR with latest Protocol 23 version

Shaptic:
This corresponds to https://github.com/stellar/stellar-xdr/commit/4b7a2ef7931ab2ca2499be68d849f38190b443ca.

This is a great example of the step-by-step process to update XDR with each commit message describing the detailed separate steps of the process.


sreuland:
the separate commits for each step are great. Do you think it's worth including link to this pr as example of those steps in the existing [README#xdr-update](https://github.com/stellar/js-stellar-base/blob/bump-xdr/README.md#to-develop-and-test-js-stellar-base-itself), it doesn't mention this type of guidance.

sreuland (approved):
lgtm, my approval is probably half strengh, I'm not active in js layer as much.

Shaptic:
Yep good call @sreuland, I jotted that down as a TODO but might as well throw it in. I think your :heavy_check_mark: is good enough - there's no major code changes here and CI passes.

PR #801, opened 2025-05-20
Add support for encoding `Address` to/from new `StrKey`s

Shaptic:
* LPs, CBs, and muxed accounts now can be converted:
  - from strkey -> Address
  - Address -> ScVal
  - Address -> ScAddress
  - ScVal -> Address

### DIFF
--- a/src/address.js
+++ b/src/address.js
@@ -86,11 +86,11 @@ export class Address {
   }
 
   /**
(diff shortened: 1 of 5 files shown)


sreuland (approved):
lgtm, nice work, thanks for addressing feedback questions.

PR #803, opened 2025-06-18
Allow specifying mixed types for vectors in `nativeToScVal`

Shaptic:
Previously, you could only specify a single type or use defaults:

```js
nativeToScVal([1, 2, 3], { type: 'i128' }); // Vec<I128>
nativeToScVal([1, "2", 3], { type: 'i128' }); // error
nativeToScVal([1, "2", 3]); // Vec<u32, string, u32>
```

but now you can specify types for each element in the array

```js
nativeToScVal([1, "2", 3], { type: ['i128', 'string', 'i128'] }); // Vec<I128, String, I128>
```

This is super helpful because contracts are just `Vec<Val>` so they can be anything and unnamed enums also require this format.

### DIFF
--- a/src/scval.js
+++ b/src/scval.js
@@ -91,6 +91,7 @@ import { ScInt, XdrLargeInt, scValToBigInt } from './numbers/index';
  * nativeToScVal(true); // scvBool
  * nativeToScVal([1, 2, 3]);                    // gives scvVec with each element as scvU64
  * nativeToScVal([1, 2, 3], { type: 'i128' });  // scvVec<scvI128>
+ * nativeToScVal([1, '2'], { type: ['i128', 'symbol'] });  // scvVec with diff types
  * nativeToScVal({ 'hello': 1, 'world': [ true, false ] }, {
  *   type: {
  *     'hello': [ 'symbol', 'i128' ],
@@ -126,6 +127,7 @@ import { ScInt, XdrLargeInt, scValToBigInt } from './numbers/index';
  *     etc: false
  *   },
  *   vec: ['same', 'type', 'list'],
+ *   vec: ['diff', 1, 'type', 2, 'list'],
  * };
  *
  * // then, simply:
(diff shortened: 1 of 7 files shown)


src/scval.js
```diff
@@ -91,6 +91,7 @@ import { ScInt, XdrLargeInt, scValToBigInt } from './numbers/index';
  * nativeToScVal(true); // scvBool
  * nativeToScVal([1, 2, 3]);                    // gives scvVec with each element as scvU64
  * nativeToScVal([1, 2, 3], { type: 'i128' });  // scvVec<scvI128>
+ * nativeToScVal([1, '2'], { type: ['i128', 'symbol'] });  // scvVec with diff types
```
sreuland, line 94: might be worth highlighting one more example where the val array has more elements than type array, and how auto typing is applied on the additional val's?

test/unit/scval_test.js
```diff
@@ -209,6 +209,18 @@ describe('parsing and building ScVals', function () {
   it('doesnt throw on arrays with mixed types', function () {
     expect(nativeToScVal([1, 'a', false]).switch().name).to.equal('scvVec');
   });
+  it('allows type specifications across an array', function () {
+    const scv = nativeToScVal([1, 'a', false, 'b'], {
+      type: ['i128', 'symbol']
+    });
+    expect(scv.switch().name).to.equal('scvVec');
+    expect(scv.value().length).to.equal(4);
+    ['scvI128', 'scvSymbol', 'scvBool', 'scvString'].forEach(
```
sreuland, line 218: wow, that is very elastic, many ways for the caller to leverage!

sreuland (approved):
lgtm, I'm light on experience in js base though, so, can take this approval at whatever grade you choose.

Shaptic:
@sreuland you're my go to on all of these :pray: so I'll take what I can get :shipit:

Issue #722, opened 2024-01-01
Simplify how to convert from ScVal types into BigInts/Numbers


leighmcculloch:
### Is your feature request related to a problem? Please describe.
The APIs for working with XDR ScVal number types are reasonably confusing and non-obvious. There are also multiple ways to do the same thing.

For example there are three ways to get at / work with number values:

1. Use the `scValToBigInt` function:
   ```js
   // ScVal -> BigInt
   const bigint = scValToBigInt(scval);
   ```

2. Use the `ScInt` type:
   ```js
   // ScVal -> BigInt
   const bigint = ScInt.fromScVal(scval);
   ```
   ```js
   // BigInt -> ScVal
   const scval = new ScInt(bigint).toI128();
   ```

   _Another problem with `ScInt` is that it looks like an XDR type, but it isn't._

3. Use the `XdrLargeInt` type. This type should probably not be exported as it seems like an implementation detail?

   ```js
   // ScVal -> BigInt
   const bigint = new XdrLargeInt(scIntType, [scval.value().lo(), scval.value().hi()]).toBigInt();
   ```

   _Another problem with `XdrLargeInt` is that there's another type `LargeInt` which serves another purpose._

### Describe the solution you'd like
Functions on `ScVal`. Close proximity to `ScVal` so that the functionality is discoverable, and less new concepts (`XdrLargeInt`, `ScInt`) are introduced).

Expand the existing ScVal functions to accept Number/BigInt:
```diff
-xdr.ScVal.scvI128(value: xdr.Int128Parts): xdr.ScVal
+xdr.ScVal.scvI128(value: xdr.Int128Parts | BigInt): xdr.ScVal
```

Add functions for converting from ScVal to Number/BigInt:
```js
const bigint = scval.toBigInt();
const number = scval.toNumber(); // with check
```

### Describe alternatives you've considered
Another option would be to focus on using just one of these methods, such as combining `ScInt` and `XdrLargeInt` so all conversions (simple and advanced) are in one type, and removing `scValToBigInt` since it's superfluous.

### Additional context
Thinking about this after seeing the following question in Discord:
- https://discord.com/channels/897514728459468821/1189109010754977802

[...]

PR #809, opened 2025-06-30
Unify `ScInt` & `scValToBigInt` into `XdrLargeInt`

Shaptic:
### What
* `XdrLargeInt` has been renamed to `Int`
* `ScInt` and `scValToBigInt` have been removed to simplify the API surface:
```typescript
// before:
new ScInt(value);
new ScInt(value, { type: 'i128' });
// after:
Int.fromValue(value);
new Int('i128', value);

// before:
scValToBigInt(scv);
// after:
Int.fromScVal(scv).toBigInt();
```
* `Int` also has new methods and features:
  - `fromValue(x: number|string|BigInt)` will create an `Int` of an appopriate size for the given value
  - `fromScVal(x: xdr.ScVal)` will create an `Int` from an `ScVal`
  - `.toU32()` will return an `ScVal` of the type `scvU32`
  - `.toI32()` will return an `ScVal` of the type `scvI32`
  - the constructor now accepts `i32` and `u32` as the first `type` parameter

### Why
Addresses #722.


orbitlens:
LGTM. The only question, does this implementation really need static methods `fromValue` and `fromScVal`? I'd prefer to just use a constructor with whatever input people going to put there. You can dynamically determine the type of the constructor argument and parse it. In this case the type argument can be optional.

E.g.
```ts
constructor(value: number|string|BigInt|xdr.ScVal, type: string?) {
  switch(typeof value) {
    case 'number': ... break
    case 'string': ... break
    case 'bigint': ... break
    default: //check for ScVal with duck-typing
    break
  }
}
``` 
```js
new Int(12, 'i128'); 
new Int('12', 'u64'); 
new Int(12n, 'u128'); 
new Int(scv); 
```

The only potential problem might be with ScVal if people use multiple StellarBase libs in a project (for example importing StellarBase and StellarSDK, I faced this problem multiple times with XDR and Transactions). So would be nice to use some basic duck-typing check for ScVal.

leighmcculloch:
Because a lot of the types in the SDK offer static functions, to aid with discoverability we could keep the `from` function, but make it so there is only one, and that the `from` function mirrors the constructor, so there's really only one function and one concept, it's just available either as a constructor or as a static function.

leighmcculloch:
> ```js
> new Int(12, 'i128'); 
> new Int('12', 'u64'); 
> new Int(12n, 'u128'); 
> new Int(scv); 
> ```

Would the second argument, the type, be required for non-ScVal inputs?

Shaptic:
@orbitlens that's the way that `ScInt` was designed initially, but `XdrLargeInt` required the `type` so it came first (bad call at the time in retrospect), so making `XdrLargeInt` -> `Int` and also switch the parameter order `(values, type?)` instead of `(type, values)` would be a bigger breaking change to any `XdrLargeInt` users that I wanted to avoid, hence the `fromValue`. Ideally yes, `fromValue` could be absorbed into the constructor, but it'd be annoying to migrate, and letting it be `(undefined, [values])` is ugly :/

If people don't mind migrating we can do it this way, then also allow `values` to be an `ScVal`. That'd be the ideal case. Worth it?

kalepail (approved):
I've reviewed and don't have anything substantive to add. I haven't used the existing method much and the proposed change seems clean 👍

PR #814, opened 2025-08-09
Fixes muxed account handling in `Address` class

overcat:
The encoding of muxed accounts in XDR is different from that in StrKey; XDR uses an 8-byte ID + 32-byte key, while StrKey uses a 32-byte key + 8-byte ID.

### DIFF
--- a/src/address.js
+++ b/src/address.js
@@ -118,8 +118,13 @@ export class Address {
         return Address.account(scAddress.accountId().ed25519());
       case xdr.ScAddressType.scAddressTypeContract().value:
         return Address.contract(scAddress.contractId());
-      case xdr.ScAddressType.scAddressTypeMuxedAccount().value:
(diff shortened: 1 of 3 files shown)


src/address.js
```diff
@@ -118,8 +118,13 @@ export class Address {
         return Address.account(scAddress.accountId().ed25519());
       case xdr.ScAddressType.scAddressTypeContract().value:
         return Address.contract(scAddress.contractId());
-      case xdr.ScAddressType.scAddressTypeMuxedAccount().value:
-        return Address.muxedAccount(scAddress.muxedAccount());
```
overcat, line 122: There is also a type error here; `Address.muxedAccount` accepts a `Buffer` parameter, but `scAddress.muxedAccount()` is not a `Buffer`.

PR #834, opened 2026-01-22
Update readme

quietbits:
- Update readme to reflect the current Node v20 version we use.
- I'm not sure if the formatting update is coming from something I have locally, or if this file hasn't been touched in a while to get the current formatting.


Ryang-21 (approved):
lgtm

PR #847, opened 2026-02-09
Simple utils ts migration

Ryang-21:
Converts a single file in the util folder to check migration process


Ryang-21:
I believe the workflow checks are failing because I initially had this PR merge into master and it still thinks that after fixing and rerunning

src/operations/set_options.js
```diff
@@ -1,5 +1,3 @@
-/* eslint-disable no-param-reassign */
```
quietbits: Why are we removing these?

PR #872, opened 2026-02-23
 Migrate `keypair.js` file to TS

quietbits:


src/keypair.ts
```diff
@@ -130,12 +134,10 @@ export class Keypair {
...
+   * @param [id] - stringified integer indicating the underlying muxed
    *     ID of the new account object
-   *
-   * @return {xdr.MuxedAccount}
    */
-  xdrMuxedAccount(id) {
+  xdrMuxedAccount(id?: string): xdr.MuxedAccount {
     if (typeof id !== "undefined") {
       if (typeof id !== "string") {
         throw new TypeError(`expected string for ID, got ${typeof id}`);
```
Ryang-21: We can make `id` a required argument as the manual declarations specify this.

src/keypair.ts
```diff
@@ -49,45 +53,46 @@ export class Keypair {
...
+    } else if (keys.publicKey) {
       this._publicKey = Buffer.from(keys.publicKey);
 
       if (this._publicKey.length !== 32) {
         throw new Error("publicKey length is invalid");
       }
+    } else {
+      throw new Error(
+        "At least one of publicKey or secretKey must be provided",
+      );
```
Ryang-21: The generated constructor does not match the manual type declarations

src/keypair.ts
```diff
@@ -96,32 +101,31 @@ export class Keypair {
...
-    return this.fromRawEd25519Seed(secret);
+  static random(): Keypair {
+    const { secretKey } = ed25519.keygen();
+    return this.fromRawEd25519Seed(Buffer.from(secretKey));
   }
 
-  xdrAccountId() {
-    return new xdr.AccountId.publicKeyTypeEd25519(this._publicKey);
+  xdrAccountId(): xdr.AccountId {
+    return xdr.PublicKey.publicKeyTypeEd25519(this._publicKey);
```
Ryang-21: Looks like this should be `AccountId`

Ryang-21 (changes requested):

PR #873, opened 2026-02-23
Migrate memo.js to TS

Ryang-21:


src/memo.ts
```diff
@@ -23,6 +23,36 @@ export const MemoHash = "hash";
...
+  ? null
+  : T extends MemoType.ID
+    ? string
+    : T extends MemoType.Text
+      ? Buffer | string
+      : T extends MemoType.Hash
+        ? Buffer
+        : T extends MemoType.Return
+          ? Buffer
+          : MemoValue;
```
quietbits: Nesting this deep is really hard to read. Maybe we could do something like this instead?

```typescript
type MemoValueMap = {
  [MemoNone]: null;
  [MemoID]: string;
  [MemoText]: Buffer | string;
  [MemoHash]: Buffer;
  [MemoReturn]: Buffer;
};

type MemoTypeToValue<T extends MemoType> = MemoValueMap[T];
```

PR #891, opened 2026-03-03
Some fixes

quietbits:
- `src/soroban.js` - `parseTokenAmount()`
- `src/auth.js` - `bytesToInt64()`
- `src/numbers/sc_int.js` - `nearestBigIntSize()`
- `src/transaction_base.js` - `set networkPassphrase()`
- updated `.gitignore`

### DIFF
--- a/src/soroban.js
+++ b/src/soroban.js
@@ -18,28 +18,28 @@ export class Soroban {
    * formatTokenAmount("123", 3) === "0.123";
    */
   static formatTokenAmount(amount, decimals) {
-    if (amount.includes(".")) {
-      throw new TypeError("No decimals are allowed");
+    if (amount.includes('.')) {
+      throw new TypeError('No decimals are allowed');
     }
 
     let formatted = amount;
     if (decimals > 0) {
       if (decimals > formatted.length) {
-        formatted = ["0", formatted.toString().padStart(decimals, "0")].join(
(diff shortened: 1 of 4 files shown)


src/soroban.js
```diff
@@ -69,6 +69,12 @@ export class Soroban {
       throw new Error(`Invalid decimal value: ${value}`);
     }
 
+    if (fraction?.length > decimals) {
+      throw new Error(
+        `Too many decimal places in "${value}": expected at most ${decimals}, got ${fraction.length}`
+      );
+    }
```
Ryang-21: Should we allow this and just truncate the return value to to the decimals they specified?

src/transaction_base.js
```diff
@@ -60,7 +60,7 @@ export class TransactionBase {
   }
 
   set networkPassphrase(networkPassphrase) {
-    this._networkPassphrase = networkPassphrase;
+    throw new Error('Transaction is immutable');
   }
```
Ryang-21: This will have to be noted as a breaking change

src/soroban.js
Ryang-21: Yeah not opposed to throwing an error but wanted to see if there was a reason we did not go that

Ryang-21 (approved):

PR #915, opened 2026-03-06
Migrated `liquidity_pool_asset.js` file to TS

quietbits:


src/liquidity_pool_asset.ts
```diff
@@ -92,29 +101,28 @@ export class LiquidityPoolAsset {
 
   /**
    * @see [Assets concept](https://developers.stellar.org/docs/glossary/assets/)
-   * @returns {AssetType.liquidityPoolShares} asset type. Can only be `liquidity_pool_shares`.
    */
-  getAssetType() {
+  getAssetType(): AssetType.liquidityPoolShares {
     return "liquidity_pool_shares";
   }
```
Ryang-21: `asset.ts` has an exported `AssetType` that we can use

PR #939, opened 2026-03-17
Fix mutable tx handle and memo ID validation gap

sagpatil:
Return a defensive XDR copy from TransactionBase#tx getter to prevent mutation of the internal XDR object after construction, closing a split-brain where displayed fields could diverge from signed bytes. Internal methods now use this._tx directly for signing/serialization.

Add regex validation in Memo._validateIdValue to reject non-plain-decimal strings (e.g. "1e18", "1.0") that pass BigNumber checks but crash BigInt() at XDR serialization time.

Fixes (https://github.com/stellar/internal-agents/issues/160)

### DIFF
--- a/src/fee_bump_transaction.js
+++ b/src/fee_bump_transaction.js
@@ -46,7 +46,7 @@ export class FeeBumpTransaction extends TransactionBase {
     const innerTxEnvelope = xdr.TransactionEnvelope.envelopeTypeTx(
       tx.innerTx().v1()
     );
-    this._feeSource = encodeMuxedAccountToAddress(this.tx.feeSource());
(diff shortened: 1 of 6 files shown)


quietbits (approved):

Ryang-21:
I have a concern with this approach in that it's a silent change to the previous functionality. We would need to note this as a breaking change for users who are mutating the underlying `xdr.Transaction`. I think a better approach is to allow for TransactionBase._tx to be mutable but modify the subclass's getter methods to retrieve relevant data from the underlying instead of relying on static data from initialization.

sagpatil:
@Ryang-21 do we need this or with js base refactor this is n longer valid.

quietbits:
@sagpatil this issue was fixed in `typescript-migration` branch.

PR #948, opened 2026-03-24
Account: allow negative sequence number

quietbits:
There is a use case for a negative sequence number. For example, we use it in `WebAuth.buildChallengeTx` ([code](https://github.com/stellar/js-stellar-sdk/blob/master/src/webauth/challenge_transaction.ts#L77)). 

> [!NOTE]
> This was introduced during migration. This check does not exist in the current version.

### DIFF
--- a/src/account.ts
+++ b/src/account.ts
@@ -45,10 +45,6 @@ export class Account {
       throw new Error("sequence is not a valid number");
     }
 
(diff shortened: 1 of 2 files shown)


Ryang-21 (approved):
Nice catch

PR #963, opened 2026-04-02
Various fixes, part 5

quietbits:
Medium items 6-10 from [the doc](https://docs.google.com/document/d/1x9AcaCsz4MNKNcGx0y_4FTwBno6Jcirl09yax1KRDB0/edit?tab=t.0).

### DIFF
--- a/test/unit/operations/change_trust.test.ts
+++ b/test/unit/operations/change_trust.test.ts
@@ -105,26 +105,30 @@ describe("Operation.changeTrust()", () => {
 
   it("round-trips an Asset changeTrust through fromXDRObject and back", () => {
     const op = Operation.changeTrust({ asset: usd, limit: "50.0000000" });
+    const xdrHex = op.toXDR("hex");
     const parsed = expectOperationType(
-      Operation.fromXDRObject(xdr.Operation.fromXDR(op.toXDR("hex"), "hex")),
+      Operation.fromXDRObject(xdr.Operation.fromXDR(xdrHex, "hex")),
       "changeTrust",
     );
 
     // parsed has `line` (not `asset`); changeTrust accepts both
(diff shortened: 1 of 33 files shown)


test/unit/operations/change_trust.test.ts
```diff
@@ -103,6 +103,30 @@ describe("Operation.changeTrust()", () => {
...
+  it("round-trips a LiquidityPoolAsset changeTrust through fromXDRObject and back", () => {
+    const op = Operation.changeTrust({ asset: lpAsset });
+    const parsed = expectOperationType(
+      Operation.fromXDRObject(xdr.Operation.fromXDR(op.toXDR("hex"), "hex")),
+      "changeTrust",
+    );
+
+    // parsed has `line` (not `asset`); changeTrust accepts both
+    const rebuilt = Operation.changeTrust(parsed);
+    expect(rebuilt).toBeInstanceOf(xdr.Operation);
```
Ryang-21, line 130: we should check that the rebuilt and the original op are the same. An easy way to do this would be string comparing the xdr strings of the two

test/unit/operations/manage_data.test.ts
```diff
@@ -46,6 +46,23 @@ describe("Operation.manageData()", () => {
...
+      Operation.fromXDRObject(operation),
+      "manageData",
+    );
+
+    // Rebuilding from the parsed result must not throw
+    const rebuilt = Operation.manageData({
+      name: parsed.name,
+      value: parsed.value,
+    });
+    expect(rebuilt).toBeInstanceOf(xdr.Operation);
```
Ryang-21, line 63: ditto here we should check that they are the same operation

Ryang-21 (approved):

PR #971, opened 2026-05-27
Allow bitwise setOptions flag types

### LINKED ISSUE #347: TypeScript definition for clearFlags  and setFlags don't allow bitwise operations.
You can use bitwise operation to turn on multiple flags in the same transaction, however, TypeScript complains that the value isn't valid, since it's expecting the value to be either 1,2, or 4.

```
Operation.setOptions({
    setFlags: AuthRequiredFlag | AuthRevocableFlag
})
```
It's very likely other fields with bitwise support have the same issue.

chrisx9z:
## Summary
- Allows `setOptions` `setFlags` and `clearFlags` inputs to accept bitwise bitmap values without TypeScript casts.
- Adds `AuthFlagInput` for user-provided numeric/string flag values while keeping operation output flags numeric.
- Updates the dtslint coverage to use bitwise flag expressions directly and cover string flag inputs.

Fixes #347.

## Verification
- `corepack yarn dtslint --localTs node_modules/typescript/lib types/`
- `node --check types/test.ts`
- `git diff --check`

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -263,7 +263,9 @@ export type AuthFlag =
   | AuthFlag.required
   | AuthFlag.immutable
   | AuthFlag.revocable
-  | AuthFlag.clawbackEnabled;
+  | AuthFlag.clawbackEnabled
+  | number;
+export type AuthFlagInput = AuthFlag | string;
 
 export namespace TrustLineFlag {
   type deauthorize = 0;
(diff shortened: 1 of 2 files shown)


Ryang-21:
Hey I appreciate the contribution! This was implemented in https://github.com/stellar/js-stellar-sdk/releases/tag/v16.0.0. Closing this but feel free to open a pr in https://github.com/stellar/js-stellar-sdk if there are other changes you would like to see.

PR #972, opened 2026-05-28
SPIKE: Protocol 27 — CAP-71 + CAP-83 XDR regen

sisuresh:
> ⚠️ **This is an experiment, not a real protocol-bump PR.** It was opened from a [pod-fsr#33](https://github.com/stellar/pod-fsr/issues/33) spike to render live CAP-71 / CAP-83 (Protocol 27) transactions in stellar-laboratory against a local Quickstart Protocol 27 build. It is not intended to be merged as-is — it bundles several workarounds that should be replaced by proper upstream fixes before merging. Filed as a reference point for the workflow.

## What this changes

- **Makefile:** bump `XDR_BASE_URL_{CURR,NEXT}` to `5187e69` (the Protocol 27 .x with CAP-71 + CAP-83). Add a \`stellar-xdr xfile preprocess --features CAP_0071,CAP_0083\` step before the Ruby xdrgen invocation, because Ruby xdrgen does not parse \`#ifdef CAP_<N>\` directives — [stellar/rs-stellar-xdr#503](https://github.com/stellar/rs-stellar-xdr/pull/503) is the canonical preprocessor. Drop \`docker run -it\` (breaks non-TTY). Switch the dts-xdr container from \`node:alpine\` to \`node:lts-alpine\` and \`apk add yarn\` (newer alpine no longer bundles yarn).

- **Source post-process (the experimental workaround):** inline each \`xdr.const("<NAME>", <value>)\` value at every bare-identifier usage site (\`xdr.string(SCSYMBOL_LIMIT)\` → \`xdr.string(32)\`). xdrgen master no longer emits the \`var <NAME> = <value>;\` declarations the released v15.0.0 had, and \`@stellar/js-xdr\`'s \`TypeBuilder.const()\` doesn't inject the identifier into the calling scope. The obvious patch (re-inject the \`var\` decls) works in unminified dev builds but **terser tree-shakes them out of the production browser dist**, leaving bare identifiers at runtime. Inlining the literal is the only fix that survives minification.

- **Regenerated** \`src/generated/{curr,next}_generated.js\` + \`types/curr.d.ts\` with CAP-71-02 (\`SOROBAN_CREDENTIALS_ADDRESS_V2\`, \`HashIdPreimageSorobanAuthorizationWithAddress\`), CAP-71-01 (\`SorobanAddressCredentialsWithDelegates\`, \`SorobanDelegateSignature\`), and CAP-83 (\`STELLAR_VALUE_EMPTY_TX_SET\`).

## Verification

Decoded a live CAP-71-01 envelope all the way through \`xdr.TransactionEnvelope.fromXDR\` → \`sorobanCredentialsAddressWithDelegates\` → root contract address → 1 \`SorobanDelegateSignature\` → \`scvVec\` delegate sig, in both Node and the browser-bundled (terser-minified) \`dist/stellar-base.min.js\`. Then end-to-end through [js-stellar-sdk](https://github.com/stellar/js-stellar-sdk/pull/1423) → [laboratory](https://github.com/stellar/laboratory) to render the transaction view in lab.

## Why this should not be merged as-is

- The inline-const post-process is a downstream workaround for an upstream issue in xdrgen master + \`@stellar/js-xdr\`. The real fix is in those tools: either xdrgen should re-emit the explicit \`var\` declarations the v15.0.0 release had, or \`TypeBuilder.const()\` should attach the identifier to a scope terser can see (e.g., \`globalThis\`). This PR's regen pipeline is a stopgap.
- The Makefile container swap (\`node:alpine\` → \`node:lts-alpine\`) is incidental; a proper fix would pin a specific tag.
- The XDR commit (\`5187e69\`) is a pre-release stellar-xdr SHA; the eventual real bump would land after Protocol 27 freeze.

## Related experiments
- [stellar/js-stellar-sdk#1423](https://github.com/stellar/js-stellar-sdk/pull/1423) — re-points stellar-base + rebuilds browser dist.
- [stellar/js-stellar-xdr-json#52](https://github.com/stellar/js-stellar-xdr-json/pull/52) — wasm decoder at \`rs-stellar-xdr@a749b69b\` features=[\"cap_0071\"].
- [stellar/rs-soroban-sdk#1878](https://github.com/stellar/rs-soroban-sdk/pull/1878) — \`Address::delegate_account_auth\` wrapper.

Source: [pod-fsr#33](https://github.com/stellar/pod-fsr/issues/33) — full-stack CAP spike workflow exploration.


xdr/curr/Stellar-transaction.x
```diff
@@ -581,6 +596,10 @@ case SOROBAN_CREDENTIALS_SOURCE_ACCOUNT:
     void;
 case SOROBAN_CREDENTIALS_ADDRESS:
     SorobanAddressCredentials address;
+case SOROBAN_CREDENTIALS_ADDRESS_V2:
+    SorobanAddressCredentials addressV2;
```
Shaptic, line 600: I'm confused about what makes V2 actually a V2 if it's using the exact same struct type for that branch. Is this a generator bug?

PR #983, opened 2026-08-31
SPIKE: Protocol 30 (CAP-0088)

sisuresh:
## Changes
* Bump XDR pin to stellar-xdr `03cbf40c` and regen; enables `MS_CLOSE_TIME` on the `next` channel only (`StellarValueSignedMsValue`/`StellarValueProposedMsValue` + `TimePointMilliseconds`). `curr` also picks up the pin's already-live CAP-0083 `STELLAR_VALUE_EMPTY_TX_SET` arm (no CAP-0088 content), since both channels share one `.x` snapshot.
* Bring in the preprocess/post-process regen infra (`stellar-xdr xfile preprocess`, `scripts/post-process-generated.py` const-inlining, dropped `-it`, `node:lts-alpine`) — none of this existed on `master` yet.
* Pin `XDRGEN_COMMIT` to `d54959f` (just before stellar/xdrgen#233 removed the JS generator upstream) so `make generate` still works.
* Hand-patch `types/curr.d.ts`/`types/next.d.ts` for the new arms — dts-xdr's raw transform output no longer matches this repo's committed module wrapper (missing header/`export`/`xdrHidden` hack), so a full regen isn't usable; patched surgically instead, mirroring the CAP-0085 precedent.

## Out-of-scope observations
* ⚠️ stellar/xdrgen#233 (merged 2026-08-18) removed the JavaScript code generator from `xdrgen` `master` entirely — breaks `XDR_BASE`-style regen for every consumer of this Makefile, not just this CAP. Worked around here via the `XDRGEN_COMMIT` pin; flagging upstream separately.
* ⚠️ dts-xdr's `src/transform.js` output no longer matches this repo's committed `types/*.d.ts` wrapper (header comment, `import { Operation } from './index'`, `export {}`, `xdrHidden` namespace) — those aren't produced by the automated transform and must be hand-maintained. Not fixed here (out of scope); flagging for future regens.

Downstream: none yet in this stack.


sisuresh:
⚠️ OUT-OF-SCOPE: stellar/xdrgen#233 (merged 2026-08-18, https://github.com/stellar/xdrgen/pull/233) removed the JavaScript code generator from `xdrgen` `master` entirely (only `go.rb` remains under `lib/xdrgen/generators/`). This breaks `make generate` for every consumer of this repo's Makefile, not just CAP-0088 — unrelated to this CAP. I did NOT fix xdrgen itself; I worked around it in this repo by pinning `XDRGEN_COMMIT=d54959f8949b8f354541bf0cc2af6a9130f0f3a9` (the commit immediately before #233) in the Makefile. Suggest filing an issue against stellar/xdrgen to either restore JS support or officially deprecate it so downstream consumers can migrate.

sisuresh:
⚠️ OUT-OF-SCOPE: dts-xdr's `src/transform.js` (https://github.com/stellar/dts-xdr) produces a bare `declare namespace xdr { ... }` block, but this repo's committed `types/curr.d.ts`/`types/next.d.ts` wrap that in a header (generation-date comment, `import { Operation } from './index'`, `export {}`, and an `xdrHidden` namespace hack for the `Operation`/`Operation2` name collision) that the automated transform does not emit. Running `make generate` for the `.d.ts` targets today overwrites that wrapper and produces a non-module file that fails dtslint (`File '.../types/curr.d.ts' is not a module`). I did NOT fix dts-xdr or restore an automated wrapper-generation step; for this PR I hand-patched the existing wrapped files with the new type declarations (verified against a scratch full regen for correctness) instead of running `make generate` for `types/*.d.ts`. Future regens should do the same until dts-xdr's transform (or a wrapper script in this repo) is updated to re-emit that header.

